PR #13, opened 2026-01-06
refactor: extend structural hash/equal to support all IRNode types and add base Stmt class

lyfne123:
- Rename type_name() to TypeName() and change return type from const char* to std::string
- Generalize reflection system from Expr-specific to IRNode-agnostic:
  * Rename IsExprField -> IsIRNodeField
  * Rename IsExprVectorField -> IsIRNodeVectorField
  * Rename VisitExprField -> VisitIRNodeField
  * Rename VisitExprVectorField -> VisitIRNodeVectorField
- Extend structural_hash() and structural_equal() to work with IRNodePtr instead of ExprPtr
- Add base Stmt class and support for Stmt nodes in structural hash and equality checks
- Remove backward compatibility code for ExprPtr overloads

### DIFF
--- a/src/ir/transform/structural_hash.cpp
+++ b/src/ir/transform/structural_hash.cpp
@@ -17,8 +17,10 @@
 #include <vector>
 
 #include "pypto/core/logging.h"
+#include "pypto/ir/core.h"
 #include "pypto/ir/reflection/field_visitor.h"
 #include "pypto/ir/scalar_expr.h"
+#include "pypto/ir/stmt.h"
 #include "pypto/ir/tensor_expr.h"
 #include "pypto/ir/transform/base/functor.h"
 #include "pypto/ir/transform/transformers.h"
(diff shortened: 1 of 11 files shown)


src/ir/transform/structural_hash.cpp
```diff
@@ -149,46 +151,59 @@ int64_t StructuralHasher::HashVar(const VarPtr& op) {
...
+    HASH_DISPATCH(TensorVar)
+
+    // Unknown expression type
+    throw pypto::TypeError("Unknown expression type in StructuralHasher::HashNode");
   }
 
-  // Tensor expressions
-  HASH_DISPATCH(TensorVar)
+  // Check if it's a Stmt
+  if (auto stmt = std::dynamic_pointer_cast<const Stmt>(node)) {
```
Hzfengsy, line 189: I wonder if we need to separate the Stmt node and the Expr node. Or we can flatten all nodes

lyfne123: updated，use EQUAL_DISPATCH handle different types

PR #14, opened 2026-01-06
refactor(ir): simplify structural equal/hash type dispatch and add TensorVar auto-mapping

lyfne123:
- Simplify type dispatch logic in structural_hash.cpp
- Add TensorVar auto-mapping support to structural_hash.cpp

### DIFF
--- a/src/ir/transform/structural_hash.cpp
+++ b/src/ir/transform/structural_hash.cpp
@@ -51,6 +51,7 @@ class StructuralHasher {
  private:
   int64_t HashNode(const IRNodePtr& node);
   int64_t HashVar(const VarPtr& op);
+  int64_t HashTensorVar(const TensorVarPtr& op);
 
   template <typename NodePtr>
   int64_t HashNodeImpl(const NodePtr& node);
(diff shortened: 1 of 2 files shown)


src/ir/transform/structural_hash.cpp
```diff
@@ -51,13 +51,16 @@ class StructuralHasher {
...
+  int64_t HashTensorVar(const TensorVarPtr& op);
 
   template <typename NodePtr>
   int64_t HashNodeImpl(const NodePtr& node);
 
   bool enable_auto_mapping_;
   std::unordered_map<const Var*, int64_t> var_map_;
   int64_t free_var_counter_ = 0;
+  std::unordered_map<const TensorVar*, int64_t> tensor_var_map_;
+  int64_t free_tensor_var_counter_ = 0;
```
Hzfengsy, line 63: No need to classify the free tensor var and free var. They can use a common counter, as it's only used for mapping

src/ir/transform/structural_hash.cpp
```diff
@@ -135,20 +138,42 @@ int64_t StructuralHasher::HashVar(const VarPtr& op) {
...
   } else {
     // Without auto-mapping: hash the VarPtr itself (pointer-based)
     h = hash_combine(h, static_cast<int64_t>(std::hash<VarPtr>{}(op)));
   }
   return h;
 }
 
+int64_t StructuralHasher::HashTensorVar(const TensorVarPtr& op) {
+  int64_t h = static_cast<int64_t>(std::hash<std::string>{}("TensorVar"));
+  if (enable_auto_mapping_) {
```
Hzfengsy, line 155: I wonder the logic is wrong

PR #70, opened 2026-01-22
feat(pass): Add InitMemRef Pass.

zhangqi-chen:
Add InitMemRef pass. This pass initializes MemRef for all Vars in a function, in which size & type are initialized, addr is set to 0.

### DIFF
--- a/src/ir/transform/passes/init_memref.cpp
+++ b/src/ir/transform/passes/init_memref.cpp
@@ -1,0 +1,174 @@
+/*
+ * Copyright (c) PyPTO Contributors.
+ * This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+ * CANN Open Software License Agreement Version 2.0 (the "License").
+ * Please refer to the License for details. You may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


src/ir/transform/passes/init_memref.cpp
```diff
@@ -0,0 +1,170 @@
...
+    }
+    // Continue visiting arguments
+    IRVisitor::VisitExpr_(op);
+  }
+};
+
+// Mutator to initialize MemRef for variables
+class InitMemRefMutator : public IRMutator {
+ public:
+  const std::set<std::string>& ddr_vars_;
```
Hzfengsy: keep members private

src/ir/transform/passes/init_memref.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+namespace pypto {
+namespace ir {
+
+namespace {
+
+// Visitor to identify variables that should be in DDR memory space
+class MemRefUsageVisitor : public IRVisitor {
+ public:
+  std::set<std::string> ddr_vars_;  // Store variable names that should be DDR
```
Hzfengsy: keep members private

PR #91, opened 2026-01-26
feat(pass): Add OutOfOrderSchedulerPass for cross-pipe dependency opt…

luohuan19:
…imization

This pass performs dependency-preserving reordering of statements within reorderable segments to reduce peak live cross-pipe dependencies (max 8 events), intended to run before InsertSyncPass to avoid hardware event_id exhaustion.

Key features:
- Kahn-based topological scheduling with resource constraints
- Multiple heuristic strategies to avoid greedy dead-ends
- Graceful fallback when strict limits cannot be satisfied
- Comprehensive Python bindings and test suite
- Integration into PassManager's XPlatform optimization strategy

### DIFF
--- a/include/pypto/ir/scalar_expr.h
+++ b/include/pypto/ir/scalar_expr.h
@@ -76,7 +76,7 @@ using ScalarExprPtr = std::shared_ptr<const ScalarExpr>;
  */
 class ConstInt : public Expr {
  public:
-  const int value_;  // Numeric constant value (immutable)
+  const int64_t value_;  // Numeric constant value (immutable)
 
   /**
    * @brief Create a constant expression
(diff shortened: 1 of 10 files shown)


docs/dev/passes/OutOfOrderSchedulerPass.md
Hzfengsy, line 1: 1. rename file to `00-out-of-order-schedule.md`
2. Make it shorter, each docs should be around 200-300 lines of markdown

src/ir/transform/out_of_order_scheduler_pass.cpp
```diff
@@ -0,0 +1,574 @@
...
+      case PickStrategy::kMinSumThenMaxThenIndex:
+        return "min_sum_then_max_then_index";
+      case PickStrategy::kMinMaxThenIndex:
+        return "min_max_then_index";
+      default:
+        return "unknown";
+    }
+  };
+
+  auto Better = [](const CandidateScore& a, const CandidateScore& b, PickStrategy strategy) -> bool {
```
Hzfengsy: ask AI to reorganize code. The current version contains one large function, which is hard to read

docs/dev/passes/OutOfOrderSchedulerPass.md
luohuan19: Alright, I'll make the changes.

src/ir/transform/out_of_order_scheduler_pass.cpp
luohuan19: Alright, I'll refactor that function.

PR #115, opened 2026-01-31
feat(ir): Change dtype FP8 to FP8e4m3fn and FP8e5m2

tsung-li:
FP8 has several standards, single DataType could not present all fp8 format, replace FP8 with two formats most frequently used


tsung-li:
> python/pypto/language/__init__.py

issue fixed, please check the latest code

PR #267, opened 2026-02-26
feat(codegen): Add PTOAS-aware type annotations and update PTO naming

zhangqi-chen:
Emit accurate tile_buf/tensor_view type strings derived from TileType metadata instead of hardcoded defaults. Rename pto.subview to pto.partition_view and update memory space names (ub->vec, ddr->gm, l1->mat, etc.) to match current PTO-ISA specification. Add ins/outs type annotation support to operator codegen callbacks.


zhangqi-chen:
@Hzfengsy

Hzfengsy (approved):

PR #272, opened 2026-02-26
feat(codegen): Add kernel wrapper generation for PTO backend

zhangqi-chen:
Extract PTO codegen logic into dedicated pto_codegen module with kernel wrapper generation, config file generation, and PTOAS output preprocessing. Expose OrchestrationResult, GenerateOrchestration, and InferFunctionCoreType through C++ bindings and type stubs. Add PTOAS strategy to pass manager with ConvertToSSA, FlattenCallExpr, and RunVerifier passes.


python/pypto/ir/pto_codegen.py
```diff
@@ -0,0 +1,301 @@
...
+                "ptoas binary not found. Set PTOAS_ROOT to the extracted release directory, "
+                f"or add ptoas to your PATH.\nDownload from: {_PTOAS_RELEASE_URL}"
+            )
+
+    cmd = [ptoas_bin, pto_path, "-o", output_path]
+    if ptoas_flags:
+        cmd.extend(ptoas_flags)
+
+    try:
+        result = subprocess.run(
```
zhangqi-chen, line 70: @Hzfengsy
PTO backend uses PTOAS to compile kernel.pto to kernel.cpp, which is implemented in python in this PR. This results in additional python bindings for orchestration compiling. Should we move this to C++?

zhangqi-chen:
@Hzfengsy

Hzfengsy (approved):

python/pypto/ir/pto_codegen.py
Hzfengsy: Keeping it in python is good now. We can move it to C++ later if needed

PR #278, opened 2026-02-26
feat(harness): Add PTO backend support and PTOAS on-board tests

lwDavid:
feat(harness): Add PTO backend support and PTOAS on-board tests

Enable test harness to use PTO backend (via ptoas compiler) in addition
to the default CCE backend.

Harness changes:
- Add get_backend_type() to PTOTestCase (default CCE, backward compatible)
- Accept backend_type in ProgramCodeGenerator, replacing hardcoded CCE
- Propagate backend_type from test case through TestRunner

Runtime tests:
- Add PTO+PTOAS variants for DAG, elementwise (add/mul), and matmul
- Fix TestTileAddWithPTOAS: was missing backend_type override, causing
  PTOAS strategy to run with CCE backend (sync not inserted → wrong results)
- Remove skip marker on elementwise PTOAS test now that root cause is fixed

CI:
- chmod +x ptoas-bin/bin/ptoas (actual binary, not just wrapper script)


lwDavid:
@Hzfengsy Request review.

PR #284, opened 2026-02-27
feat(op): add tmp_tile argument to block.sum for CCE workspace support

doraemonmj:
Background
The CCE hardware intrinsics TCOLSUM and TROWSUM both require a workspace buffer (tmp_tile). Previously, block.sum only accepted one argument (tile), making it impossible to pass the required workspace at codegen time.

Changes
C++ IR (src/ir/op/block_ops)

block.sum now takes a second argument tmp_tile
Refactored type deduction into three functions:
ComputeReductionOutputType: shared helper that computes output shape from axis/keepdim kwargs
DeduceBlockSumType: used by block.sum (2 args, validates tmp_tile is TileType)
DeduceBlockReductionType: used by block.max/block.min (1 arg, unchanged behavior)
C++ Backend (src/backend/910B_CCE/backend_910b_cce_ops.cpp)

block.sum axis=0 → TCOLSUM(result, tile, tmp_tile, false) (2 args)
block.sum axis=1 → TROWSUM(result, tile, tmp_tile) (2 args)
block.max/block.min axis=1 → new MakeBlockRowReduction1ArgCodegenCCE emitting TROWMAX/TROWMIN(result, tile) (1 arg, no tmp_tile needed)
Python (python/pypto/ir/op/block_ops.py, python/pypto/language/op/block_ops.py)

Added tmp_tile parameter to sum()
Tests & Examples

tests/ut/ir/operators/test_block_ops.py: updated block.sum cases to pass tmp_tile
examples/ir_builder/block_ops_example.py: updated all block.sum call sites

### DIFF
--- a/python/pypto/language/op/block_ops.py
+++ b/python/pypto/language/op/block_ops.py
@@ -871,18 +871,19 @@ def cmps(lhs: Tile, rhs: int | float | Expr | Scalar, cmp_type: int = 0) -> Tile
     return Tile(expr=call_expr)
 
 
-def sum(tile: Tile, axis: int, keepdim: bool = False) -> Tile:
+def sum(tile: Tile, tmp_tile: Tile, axis: int, keepdim: bool = False) -> Tile:
     """Sum reduction along specified axis.
 
     Args:
         tile: Input tile
+        tmp_tile: Temporary tile for reduction workspace
         axis: Reduction axis (0 for rows, 1 for columns, -1 for last)
         keepdim: Whether to keep the reduced dimension as 1
 
     Returns:
         Tile wrapping the sum operation
     """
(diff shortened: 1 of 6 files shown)


python/pypto/language/op/block_ops.py
```diff
@@ -871,18 +871,19 @@ def cmps(lhs: Tile, rhs: int | float | Expr | Scalar, cmp_type: int = 0) -> Tile
     return Tile(expr=call_expr)
 
 
-def sum(tile: Tile, axis: int, keepdim: bool = False) -> Tile:
+def sum(tile: Tile, tmp_tile: Tile, axis: int, keepdim: bool = False) -> Tile:
```
YunjiQin, line 874: tmp_tile不应该让用户输入，应该在这个op内根据需要自动 create tmp_tile

PR #368, opened 2026-03-05
feat(codegen): fix pl.transpose to work correctly with TTRANS instruction

Little-oil:
## Summary

- Fix `block.transpose` codegen which was incorrectly calling `MakeTileReshapeCodegenCCE` instead of `MakeTileTransposeCodegenCCE`
- Fix `TTRANS` instruction generation to include required temporary tile buffer
- Add matmul transpose examples and tests demonstrating `A @ B^T`, `A^T @ B`, and `A^T @ B^T` patterns

### Bug Fix 1: Wrong codegen function
`block.transpose` was registered to call `MakeTileReshapeCodegenCCE`, which is incorrect. Changed to call `MakeTileTransposeCodegenCCE`.

### Bug Fix 2: TTRANS missing temporary buffer
The `TTRANS` instruction requires a temporary tile buffer as the third argument. The fix generates:
```cpp
using target_ttrans_tmpType = LocalTensor<float, ...>;
target_ttrans_tmpType target_ttrans_tmp(rows, cols);
TASSIGN(target_ttrans_tmp, 0x8000);
TTRANS(target, input, target_ttrans_tmp);
```

## Test plan

- [x] `test_matmul_trans_b` - C = A @ B^T
- [x] `test_matmul_trans_a` - C = A^T @ B  
- [x] `test_matmul_trans_ab` - C = A^T @ B^T

### DIFF
--- a/examples/language/beginner/matmul_transpose.py
+++ b/examples/language/beginner/matmul_transpose.py
@@ -1,0 +1,191 @@
+# Copyright (c) PyPTO Contributors.
+# This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+# CANN Open Software License Agreement Version 2.0 (the "License").
+# Please refer to the License for details. You may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


lyfne123:
需要在Matmul里做随路转置，而不是在vector里做转置

PR #605, opened 2026-03-19
feat(ir): add operator==/!= to TileView and centralize ExprPtr comparison

### LINKED ISSUE #551: [Code Health] Add operator== to TileView and TileType for structural comparison
### Category

Robustness (potential edge-case failure)

### Component

IR Core

### Description

`BasicMemoryReuse` pass uses `AreTileTypesCompatible()` to check whether two TileType variables can share the same buffer. This function manually compares every field of `TileType` and `TileView` one by one:

```cpp
bool AreTileTypesCompatible(const VarPtr& var1, const VarPtr& var2) {
  // ...
  if (t1->dtype_ != t2->dtype_) return false;
  if (!AreExprVectorsEqual(t1->shape_, t2->shape_)) return false;
  bool has_view1 = t1->tile_view_.has_value();
  bool has_view2 = t2->tile_view_.has_value();
  if (has_view1 != has_view2) return false;
  if (has_view1) {
    const auto& v1 = t1->tile_view_.value();
    const auto& v2 = t2->tile_view_.value();
    if (!AreExprVectorsEqual(v1.valid_shape, v2.valid_shape)) return false;
    if (v1.pad != v2.pad) return false;
    if (v1.blayout != v2.blayout) return false;
    if (v1.slayout != v2.slayout) return false;
    if (v1.fractal != v2.fractal) return false;
  }
  return true;
}
```

**Problem:** This is fragile — if a new field is added to `TileView` (e.g., a new layout attribute), the comparison will silently miss it, potentially allowing incorrect buffer reuse that leads to hardware misbehavior.

### Location

- `src/ir/transforms/basic_memory_reuse_pass.cpp:271-293` (`AreTileTypesCompatible`)
- `include/pypto/ir/type.h` (`TileView` struct, `TileType` class)

### Proposed Fix

1. Add `operator==` to `TileView` struct in `include/pypto/ir/type.h`, comparing all fields.
2. Add a shape/dtype comparison method or `operator==` to `TileType` (comparing shape element-wise via `AreExprVectorsEqual`).
3. Refactor `AreTileTypesCompatible` to use these operators, reducing it to ~3 lines.
4. This also enables other passes to safely compare tile types without manual field enumeration.

### Priority

Medium (minor risk, should fix in next few releases)

Hzfengsy:
## Summary

- Add `operator==`/`operator!=` to `TileView`, comparing all 7 fields (`valid_shape`, `stride`, `start_offset`, `blayout`, `slayout`, `fractal`, `pad`)
- Extract `AreExprsEqual`/`AreExprVectorsEqual` into `expr.h`/`expr.cpp` as reusable helpers (ConstInt by value, non-ConstInt by pointer identity)
- Refactor `AreTileTypesCompatible` in `BasicMemoryReuse` pass to delegate TileView comparison to the new `operator==`, fixing a bug where `stride` and `start_offset` fields were silently skipped
- Add Python bindings (`__eq__`/`__ne__`) and type stubs for `TileView`
- Update EN and ZH-CN documentation for the memory reuse pass

## Testing

- [x] 12 new tests in `test_tile_view_equality.py` covering all fields, default views, symbolic exprs
- [x] All 26 existing `test_basic_memory_reuse.py` tests pass
- [x] Full suite: 2491 passed
- [x] clang-tidy clean on new lines
- [x] All pre-commit hooks pass

## Related Issues

Fixes #551

### DIFF
--- a/tests/ut/ir/core/test_tile_view_equality.py
+++ b/tests/ut/ir/core/test_tile_view_equality.py
@@ -1,0 +1,133 @@
+# Copyright (c) PyPTO Contributors.
+# This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+# CANN Open Software License Agreement Version 2.0 (the "License").
+# Please refer to the License for details. You may not use this file except in compliance with the License.
+# THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+# INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+# See LICENSE in the root of the software repository for the full text of the License.
+# -----------------------------------------------------------------------------------------------------------
+
+"""Tests for TileView equality operators."""
+
+import pytest
+from pypto import DataType, ir
+
+
+def _make_span():
+    return ir.Span.unknown()
+
+
+def _make_const(value, span=None):
+    if span is None:
+        span = _make_span()
+    return ir.ConstInt(value, DataType.INT64, span)
+
+
+def _make_view(
+    valid_shape=None,
+    stride=None,
+    start_offset=None,
+    blayout=ir.TileLayout.row_major,
+    slayout=ir.TileLayout.none_box,
+    fractal=512,
+    pad=ir.PadValue.null,
+):
+    """Create a TileView with given parameters, using sensible defaults."""
+    span = _make_span()
+    if valid_shape is None:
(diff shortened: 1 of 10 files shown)


tests/ut/ir/core/test_tile_view_equality.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def _make_view(
+    valid_shape=None,
+    stride=None,
+    start_offset=None,
+    blayout=ir.TileLayout.row_major,
+    slayout=ir.TileLayout.none_box,
+    fractal=512,
+    pad=ir.TilePad.null,
```
lyfne123: TilePad is renamed to PadValue

PR #838, opened 2026-04-02
fix(ir): add K-dimension validation to tensor.matmul type deduction

### LINKED ISSUE #837: fix(ir): tensor.matmul type deduction missing K-dimension validation
### Description

`DeduceTensorMatMulType` in `src/ir/op/tensor_ops/matmul.cpp` computes output shapes without validating that the inner (K) dimensions of lhs and rhs actually match. This allows mathematically invalid matmul configurations to pass type deduction silently.

For example, `test_tensor_matmul_with_transpose` in `tests/ut/debug/test_torch_codegen.py` uses tensor shapes that are invalid under the specified transpose flags, but the test passes because K is never checked:

```python
def test_tensor_matmul_with_transpose():
    """tensor.matmul with a_trans/b_trans should emit .mT."""
    a = _tensor_var("a", [64, 128])
    b = _tensor_var("b", [128, 64])
    out = _tensor_var("out", [64, 64])
    call = _op_call("tensor.matmul", [a, b], {"a_trans": True, "b_trans": False, "c_matrix_nz": False})
    assign = ir.AssignStmt(out, call, _span())
    func = _simple_function("f", [a, b], assign)
    code = torch_codegen(func)
    assert "a.mT" in code
    assert "torch.matmul" in code
```

The `a_trans=True` kwarg is passed to `tensor.matmul` via `_op_call`, which calls `ir.create_op_call("tensor.matmul", ...)` and ends up in `DeduceTensorMatMulType`. In the type deduction logic, `a_trans=True` means:

- `lhs_k = a_trans ? lhs_shape[0] : lhs_shape[1]` → `lhs_shape[0]`
- `rhs_k = b_trans ? rhs_shape[1] : rhs_shape[0]` → `rhs_shape[0]`

With the original shapes `a=[64, 128]` and `b=[128, 64]`:
- `lhs_k = 64`, `rhs_k = 128` → **K mismatch (64 ≠ 128)**

### Root Cause

In the 2D×2D branch of `DeduceTensorMatMulType` (lines 107-111), only M and N dimensions are extracted — K is never computed or validated:

```cpp
} else if (lhs_shape.size() == 2 && rhs_shape.size() == 2) {
    ExprPtr m_dim = a_trans ? lhs_shape[1] : lhs_shape[0];
    ExprPtr n_dim = b_trans ? rhs_shape[0] : rhs_shape[1];
    output_shape = {m_dim, n_dim};
}
```

The same missing validation exists in all other branches: 1D×1D dot product, 2D×1D matrix-vector, 1D×2D vector-matrix, and batched matmul.

By contrast, `DeduceTensorMatMulAccType` (lines 206-217) correctly validates K dimensions, as do all `tile.matmul` variants.

### Fix

Add K-dimension matching checks (when statically known) to all branches of `DeduceTensorMatMulType`, following the existing `matmul_acc` pattern. Fix the test to use valid shapes: `a=[128, 64]`, `b=[128, 64]` → `K=128` on both sides.

zhaozhaozz:
## Summary

`DeduceTensorMatMulType` in `src/ir/op/tensor_ops/matmul.cpp` computed output shapes without validating that the inner (K) dimensions of lhs and rhs actually match. This allowed mathematically invalid matmul configurations (e.g., mismatched K when using `a_trans`/`b_trans`) to pass type deduction silently.

- Add K-dimension matching checks (when statically known) to **all branches** of `DeduceTensorMatMulType`: 1D×1D dot product, 2D×1D matrix-vector, 1D×2D vector-matrix, 2D×2D matrix multiply, and batched matmul
- This follows the same validation pattern already used by `tensor.matmul_acc` and all `tile.matmul` variants
- Fix `test_tensor_matmul_with_transpose` tensor shapes from `a=[64,128]` to `a=[128,64]` so that `K=128` matches `b=[128,64]` when `a_trans=True`

## Test plan

- [x] Full UT suite: 3306 passed, 16 skipped
- [x] Verified K-mismatch is correctly caught with old invalid shapes
- [x] Verified valid shapes still pass type deduction

Fixes #837

### DIFF
--- a/src/ir/op/tensor_ops/matmul.cpp
+++ b/src/ir/op/tensor_ops/matmul.cpp
@@ -97,17 +97,50 @@ TypePtr DeduceTensorMatMulType(const std::vector<ExprPtr>& args,
 
   if (lhs_shape.size() == 1 && rhs_shape.size() == 1) {
     // Vector x vector (dot product): [K] x [K] -> scalar (0D tensor)
+    auto k_lhs_const = As<ConstInt>(lhs_shape[0]);
+    auto k_rhs_const = As<ConstInt>(rhs_shape[0]);
+    if (k_lhs_const && k_rhs_const) {
+      CHECK(k_lhs_const->value_ == k_rhs_const->value_)
+          << "tensor.matmul dot product requires matching dimensions, but got lhs K=" << k_lhs_const->value_
+          << " and rhs K=" << k_rhs_const->value_;
+    }
     output_shape = {};
   } else if (lhs_shape.size() == 2 && rhs_shape.size() == 1) {
     // Matrix x vector: [M, K] x [K] -> [M]
+    auto k_lhs_const = As<ConstInt>(lhs_shape[1]);
+    auto k_rhs_const = As<ConstInt>(rhs_shape[0]);
+    if (k_lhs_const && k_rhs_const) {
+      CHECK(k_lhs_const->value_ == k_rhs_const->value_)
+          << "tensor.matmul requires matching inner dimensions, but got lhs K=" << k_lhs_const->value_
+          << " and rhs K=" << k_rhs_const->value_;
+    }
     output_shape = {lhs_shape[0]};
   } else if (lhs_shape.size() == 1 && rhs_shape.size() == 2) {
     // Vector x matrix: [K] x [K, N] -> [N]
+    auto k_lhs_const = As<ConstInt>(lhs_shape[0]);
+    auto k_rhs_const = As<ConstInt>(rhs_shape[0]);
+    if (k_lhs_const && k_rhs_const) {
+      CHECK(k_lhs_const->value_ == k_rhs_const->value_)
+          << "tensor.matmul requires matching inner dimensions, but got lhs K=" << k_lhs_const->value_
+          << " and rhs K=" << k_rhs_const->value_;
+    }
     output_shape = {rhs_shape[1]};
   } else if (lhs_shape.size() == 2 && rhs_shape.size() == 2) {
     // 2D x 2D matrix multiplication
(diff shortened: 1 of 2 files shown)


zhaozhaozz:
@Hzfengsy Good catch. You're right — the test passing indicates a bug in the shape deduction itself.

**Root cause**: `DeduceTensorMatMulType` computes output shape (M, N) based on transpose flags, but never extracts or validates the K dimension. The 2D×2D branch was just:

```cpp
ExprPtr m_dim = a_trans ? lhs_shape[1] : lhs_shape[0];
ExprPtr n_dim = b_trans ? rhs_shape[0] : rhs_shape[1];
output_shape = {m_dim, n_dim};
// K is never checked
```

The same issue exists in all other branches (1D×1D, 2D×1D, 1D×2D, and batched matmul). By contrast, `DeduceTensorMatMulAccType` and all `tile.matmul` variants already validate K correctly.

**Fix**: Added K-dimension matching checks (when statically known) to all 5 branches of `DeduceTensorMatMulType`, following the existing `matmul_acc` pattern:

```cpp
auto k_lhs_const = As<ConstInt>(k_lhs);
auto k_rhs_const = As<ConstInt>(k_rhs);
if (k_lhs_const && k_rhs_const) {
  CHECK(k_lhs_const->value_ == k_rhs_const->value_)
      << "tensor.matmul requires matching inner dimensions, but got lhs K="
      << k_lhs_const->value_ << " and rhs K=" << k_rhs_const->value_;
}
```

Also fixed the test shapes to be mathematically valid: `a=[128, 64]` with `a_trans=True` gives K=128, matching `b=[128, 64]`'s K=128.

Full UT suite passes (3306 passed, 16 skipped).

Hzfengsy (approved):

PR #994, opened 2026-04-13
ci(daily): add a5 system tests job to daily ci

luohuan19:
- Upgrade daily ci ptoas v0.24 → v0.25 (update SHA256 checksum)
- Extract PTOAS_VERSION/PTOAS_SHA256 as job-level env vars for clarity
- Cache ptoas binary via actions/cache to avoid redundant downloads
- Add new a5-system-tests job targeting [self-hosted, a5] runners:
  runs pytest tests/st/ with --forked --platform=a5 via task-submit


.github/workflows/daily_ci.yml
```diff
@@ -3,8 +3,16 @@ name: Daily CI
 on:
   schedule:
     - cron: '0 22 * * *'  # UTC 22:00 = Shanghai 06:00
+  push:
```
lyfne123: do not execute in every ci task

lyfne123 (approved):

PR #1205, opened 2026-04-28
feat(parser): Support assemble write syntax dst[slices] = src

### LINKED ISSUE #1201: [Feature] Support assemble write syntax a[:] = b (symmetric to slice read a = b[:])
### Summary

Add support for an assemble write syntax `a[offsets] = b` in the pypto frontend, symmetric to the existing slice read syntax `a = b[offsets]`.

### Motivation / Use Case

Currently the frontend supports slice reads via subscript notation:

```python
tile = tensor[offset_0:offset_0+size_0, offset_1:offset_1+size_1]
```

There is no symmetric write syntax for assemble. Users must fall back to the `pl.assemble(tile, offsets=[...], output_tensor=out)` call form. Adding a write subscript form makes the API consistent and improves readability — reads and writes are expressed with the same notation:

```python
# read
tile = src[i:i+64, j:j+64]
# write (proposed)
dst[i:i+64, j:j+64] = tile
```

### Proposed API / Behavior

```python
# Current (still valid)
out = pl.assemble(tile, offsets=[i, j], output_tensor=out)

# Proposed sugar (desugars to pl.assemble)
out[i:i+64, j:j+64] = tile
```

The subscript write form should lower to the same IR as `pl.assemble`.

### Alternatives Considered

Using `pl.assemble(...)` explicitly — works but is asymmetric with the slice read notation.

Hzfengsy:
## Summary

- Adds the symmetric subscript-write sugar `dst[i:i+H, j:j+W] = src` as a parse-time desugaring to `pl.assemble(dst, src, [i, j])`, mirroring the existing slice-read sugar `tile = src[i:i+H, j:j+W]`.
- Sugar is **pre-SSA only** — the desugaring rebinds `dst`, which strict SSA forbids. `@pl.function(strict_ssa=True)` rejects the syntax up front with a clear `UnsupportedFeatureError`.
- Parser performs a static shape-mismatch check using the same arithmetic simplifier as the slice-read path: `out[i:i+16, j:j+32] = src` folds `(i+16)-i → 16` and verifies it against `src.shape[0]`. Truly-symbolic extents like `out[:, :k] = src` (runtime `k`) are trusted.
- Element-write `dst[i, j] = scalar` is rejected for now since no element-write op is wired to the parser yet (separate concern).
- Adds `__setitem__` stubs on `Tensor` / `Tile` so static type checkers (pyright) accept the new syntax.
- Updates EN + zh-CN language guide and operation reference docs to document the sugar and its SSA constraint.

## Testing

- [x] 14 new tests in `tests/ut/language/parser/test_subscript_syntax.py` covering happy paths, every error case, and arith-simplifier matching/mismatching behavior
- [x] Full language test suite (746 tests) passes
- [x] Full IR + language sweep (3581 tests) passes
- [x] Parser sweep (630 tests) passes
- [x] Pyright passes
- [x] Ruff format / check passes

## Related Issues

Fixes #1201

### DIFF
--- a/docs/en/user/01-language_guide.md
+++ b/docs/en/user/01-language_guide.md
@@ -156,6 +156,14 @@ elem = A[i, j]         # equivalent to pl.tensor.read(A, [i, j]) / pl.tile.read(
 block = A[0:16, 0:32]  # equivalent to pl.slice(A, [16, 32], [0, 0])
 ```
 
+The symmetric write form `dst[<slices...>] = src` is sugar for `pl.assemble`:
+
+```python
+out[i:i+16, j:j+32] = src   # equivalent to out = pl.assemble(out, src, [i, j])
+```
+
+This sugar is only available before SSA conversion — it rebinds `dst`, which is incompatible with strict SSA. Under `@pl.function(strict_ssa=True)` (or any post-SSA context), use the explicit `pl.assemble(...)` call instead.
+
 Use `pl.tile.*` for tile-specific operations (memory transfers, broadcast, bitwise, etc.).
 
 ## Variable Assignment and SSA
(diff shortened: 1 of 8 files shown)


Hzfengsy:
@codex review

lyfne123 (approved):

PR #1275, opened 2026-05-06
feat(runtime): Expose worker-resident DeviceTensor (child_memory) to pypto frontend

lyfne123:
## Summary

Lets `CompiledProgram(...)` accept a worker-resident buffer in any
tensor-parameter slot, so users no longer pay H2D + D2H per call for a
weight that lives across many invocations. The frontend exposes the
underlying `ContinuousTensor.child_memory=True` mode that `runtime/`
already supported but had no path to from `pypto.runtime`.

User-facing flow:

```python
with Worker(config=RunConfig(platform="a2a3sim")) as w:
    weight = w.alloc_tensor((1024, 4096), torch.float16, init=host_w)
    for batch in batches:
        compiled(batch, weight, out)        # weight reused on device
    w.free_tensor(weight)
```

## What lands

- `pypto.runtime.DeviceTensor` — frozen `(data_ptr, shape, dtype)` handle, caller-managed lifetime, no Worker reference.
- `Worker.{malloc, free, copy_to, copy_from}` — forwarders to the underlying chip worker, gated on `_initialized`.
- `Worker.{alloc_tensor, free_tensor}` — convenience wrappers (alloc + optional H2D upload, with rollback on shape/dtype mismatch).
- `CallArg` extended with `DeviceTensor`; `execute_compiled` translates each `DeviceTensor` to `ContinuousTensor.make(..., child_memory=True)` so the runtime skips H2D, D2H, and tensor-pair recording for that arg.
- Re-exports of the simpler symbols the dispatch path needs (`ContinuousTensor`, `DataType`, `torch_dtype_to_datatype`).

## Semantics / caveats

- Mixed args (host `torch.Tensor` + `DeviceTensor`) preserve order; the host path is unchanged.
- Return-style auto-allocation remains torch-only — callers wanting a `DeviceTensor` as an output must pass it explicitly (in-place style).
- `DeviceTensor` is never copied back to the host; callers that need to read a written-to buffer call `Worker.copy_from` explicitly.
- Lifetime is caller-managed: pair every `alloc_tensor` with `free_tensor` (or `malloc` with `free`) before the Worker is closed.

## Test plan

Unit (`tests/ut/`):
- [x] `tests/ut/runtime/test_device_tensor.py` — construction validation, nbytes math, frozen + hashable
- [x] `tests/ut/runtime/test_worker_memory.py` — primitive forwarding, `_initialized` guard, `alloc_tensor` upload + mismatch rollback
- [x] `tests/ut/runtime/test_execute_compiled_device_tensor.py` — `execute_compiled` threads `child_memory=True` for DeviceTensor args while host tensors keep the `make_tensor_arg` path; skips cleanly when `simpler` is missing (same pattern as `test_worker_reuse.py::TestExecuteOnDeviceReuse`)
- [x] `tests/ut/ir/test_compiled_program.py::TestCompiledProgramDeviceTensor` — union-arg acceptance + type-rejection

System (`tests/st/`):
- [ ] `tests/st/runtime/test_device_tensor.py` — end-to-end on hardware/simulator: `alloc_tensor(init=host_b)` + two consecutive `compiled(host_a, weight_dev, host_out)` calls produce expected `a + b` (proves the device buffer persisted across calls and was not overwritten by stale H2D); plus a Worker primitive round-trip (`alloc_tensor → copy_from`) that recovers exact uploaded bytes

## Docs

- `docs/en/user/00-getting_started.md` (and zh-cn mirror) — new section "Reusing weights on the worker (DeviceTensor)" documenting the `alloc_tensor → compiled(...) → free_tensor` flow and the no-D2H caveat.

### DIFF
--- a/docs/en/user/00-getting_started.md
+++ b/docs/en/user/00-getting_started.md
@@ -213,6 +213,39 @@ print(VectorAddProgram.as_python())
 print(vector_add.as_python(concise=True))
 ```
 
+## Reusing weights on the worker (DeviceTensor)
+
+When the same large tensor is consumed by many kernel invocations — e.g. a
+weight matrix used across batches of a forward pass — uploading it on every
+call wastes bandwidth. `Worker.alloc_tensor` allocates persistent device
+memory and returns a `DeviceTensor` handle that `CompiledProgram` accepts in
+place of a `torch.Tensor`. The runtime treats the buffer as already resident
+and skips both H2D and D2H copies for that argument.
+
+```python
+import torch
+from pypto import ir
+from pypto.runtime import Worker, RunConfig
+
+compiled = ir.compile(MyKernel)
+
+with Worker(config=RunConfig(platform="a2a3sim")) as w:
+    weight = w.alloc_tensor((1024, 4096), torch.float16, init=host_weight)
+    for batch in batches:
+        out = torch.empty(batch.shape[0], 4096, dtype=torch.float16)
+        compiled(batch, weight, out)
+    w.free_tensor(weight)
+```
+
+### Caveats
+
+- A `DeviceTensor` is never copied back to the host. If a kernel writes to
+  one, call `Worker.copy_from(host_ptr, t.data_ptr, t.nbytes)` to read the
+  result.
+- Free the handle with `Worker.free_tensor` before the Worker is closed,
+  otherwise the memory leaks for the lifetime of the Worker.
+- Only the Worker that allocated the buffer can use it.
+
 ## What's Next
 
 - **[Language Guide](01-language_guide.md)** — complete reference for types, operations, control flow, memory, and compilation
--- a/docs/zh-cn/user/00-getting_started.md
+++ b/docs/zh-cn/user/00-getting_started.md
@@ -212,6 +212,36 @@ print(VectorAddProgram.as_python())
 print(vector_add.as_python(concise=True))
 ```
 
+## 在 worker 上复用权重（DeviceTensor）
+
+当同一个大张量被多次内核调用复用 —— 例如前向计算每个 batch 都要用到的权重矩阵 ——
+每次都重新上传会浪费带宽。`Worker.alloc_tensor` 在 device 上分配一块常驻内存，并返回
+一个 `DeviceTensor` 句柄；`CompiledProgram` 接受它替代 `torch.Tensor` 入参。runtime
+把这块 buffer 视为已经驻留在 device 上，对该入参跳过 H2D 与 D2H 拷贝。
+
+```python
+import torch
+from pypto import ir
+from pypto.runtime import Worker, RunConfig
+
+compiled = ir.compile(MyKernel)
+
+with Worker(config=RunConfig(platform="a2a3sim")) as w:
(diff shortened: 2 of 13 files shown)


lyfne123:
Thanks for the careful review! Pushed a fixup amending the squashed commit (now \`e6ef4fb0\`) addressing all the actionable items:

**Fixed**

- **`worker.py:236` — `alloc_tensor` shape validation.** Reject negative dims before `malloc()` so an invalid shape never gets a buffer + handle. Test: `test_alloc_negative_dim_rejected_before_malloc`.

- **`worker.py:246` — leak on staging/upload failure.** Wrapped staging + `copy_to` in `try/except` so any post-`malloc` exception (`init.contiguous().cpu()`, `copy_to`, …) frees the pointer before propagating. Same path catches the existing shape/dtype mismatch. Test: `test_alloc_copy_to_failure_frees_pointer`.

- **`worker.py:251` — non-zero `worker_id` guard.** `alloc_tensor` and `free_tensor` now reject `worker_id != 0` since `DeviceTensor` doesn't yet encode worker scope, so a `free_tensor` with the default would silently free the wrong worker's pointer. Low-risk lift later when the handle gets a `worker_id` field. Tests: `test_alloc_rejects_non_zero_worker_id`, `test_free_tensor_rejects_non_zero_worker_id`.

- **`compiled_program.py:370` — DeviceTensor metadata validation.** Added `_validate_device_tensor` that checks shape (when IR has a static shape — skip dynamic dims) and dtype against \`info.shape\`/\`info.dtype\` before dispatch, so mismatches surface as a clear \`TypeError\` at the API boundary instead of deep in the runtime. Tests: `test_device_tensor_shape_mismatch_rejected_early`, `test_device_tensor_dtype_mismatch_rejected_early`.

- **`device_tensor.py:55` — input hardening.** Reject \`bool\` for \`data_ptr\` and any dim (since \`bool\` is an \`int\` subclass), and reject non-\`int\` dims instead of silently truncating via \`int(d)\`. Tests: `test_bool_ptr_rejected`, `test_bool_dim_rejected`, `test_non_int_dim_rejected_no_silent_truncation`.

- **`test_device_tensor.py:61` — regex.** Switched to \`r\"torch\\.dtype\"\` so the dot is matched literally (also clears the Ruff RUF043 hint).

**Not changed (rationale)**

- **0-D \`DeviceTensor\` (gemini comment on `device_tensor.py:55`).** PyPTO IR routes scalar parameters through the \`pl.Scalar[T]\` / \`info.shape is None\` path with a separate \`CallArg\` branch (see `compiled_program.py:355-362`). Allowing a 0-D \`DeviceTensor\` would land it on the *tensor* branch, creating a parameter ambiguity with no consumer (the weight-reuse use case driving this PR doesn't need it). I'd rather lift that restriction in a follow-up if a real use case appears, when we can also align the scalar/tensor coercion paths.

All UT pass locally (43 added/touched cases green, 4 simpler-gated cases skip cleanly without simpler installed). The end-to-end ST test in `tests/st/runtime/test_device_tensor.py` runs on hardware/sim only and is unchanged.

Hzfengsy (approved):

PR #1371, opened 2026-05-14
fix(passes): drop unsound disjoint-store exception in DeriveCallDirections

lyfne123:
## Summary

- `DeriveCallDirections` R-seq rule had an exception that kept a callee `Out` as `OutputExisting` (instead of `InOut`) inside a sequential loop when the callee's `tile.store` offset depended on a parameter — on the assumption that a parameter-keyed offset implies disjoint per-iteration writes.
- That check was purely **syntactic**: it never verified offset stride vs. tile extent, offset injectivity, or other write paths to the same buffer. So it could **silently drop real WAW edges** — e.g. `stride < tile extent` (partial stripe overlap), or a non-injective offset like `step // 2` (multiple iterations hit the same slot).
- This PR removes the exception: a callee `Out` under any sequential ancestor is now promoted to `InOut` **unconditionally**. A genuinely-disjoint optimization can be reintroduced later behind a sound dependence analysis (affine offset extraction, stride-vs-extent, injectivity, cross-procedural composition).
- Deletes the now-dead helpers (`CalleeHasOnlyVariableOffsetStores`, `DisjointStoreVisitor`, `ScalarDefCollector`, `ExprReferencesAnyOf`/`ExprReferencesAnyOfExpanded`, `IsScalarOrTupleOfScalarsType`), the corresponding `CallDirectionMutator` members/ctor params, and the now-unused `scalar_expr.h` include.
- Updates the pass docs (en + zh-cn): direction table + R-seq prose.
- UT: converts `test_out_param_variable_offset_store_in_seq_loop_*` to expect `InOut`, drops three tests that only exercised the removed exception path.

Net: +53 / -450 lines.

## Testing

- [x] `tests/ut/ir/transforms/test_derive_call_directions.py` — 28 passed
- [x] `tests/ut/ir/transforms/` + `tests/ut/codegen/` — 1535 passed, 25 skipped (1 pre-existing unrelated failure in `test_split_vector_kernel.py` deselected — confirmed failing on `main` independent of this change)
- [x] clang-tidy on the diff — clean (the `bugprone-unchecked-optional-access` flags are on pre-existing unmodified `PriorWriterCollector` lines surfaced only by hunk shift)
- [x] Code review completed
- [x] Documentation updated (en + zh-cn)

### DIFF
--- a/docs/en/dev/passes/33-derive_call_directions.md
+++ b/docs/en/dev/passes/33-derive_call_directions.md
@@ -65,19 +65,19 @@ The pass is a `ProgramPass`. For each `Function` body it runs three sub-passes.
 
 `CallDirectionMutator` walks every non-builtin `Call`. For Group/Spmd callees the effective per-position directions are recovered via `ComputeGroupEffectiveDirections` (`orchestration_analysis.h`); other callees use their declared `param_directions_`. A `sequential_depth_` counter is incremented on non-`Parallel` `For` and on `While`, driving the *R-seq* promotion below.
 
-For each positional argument the mutator picks a direction by this table:
+For each positional argument the mutator picks a direction by this table. A callee `Out` is resolved by trying three promotion rules in order — R-seq → R-prior → R-enclosing; if none fires it stays `OutputExisting`:
 
-| Callee `ParamDirection` | Argument origin | `sequential_depth > 0`? | Prior writer in scope? | Result |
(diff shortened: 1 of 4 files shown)


Hzfengsy (approved):
LGTM — removes unsound syntactic check that could silently drop WAW dependencies.

PR #1409, opened 2026-05-19
test(st): loosen AutoTileMatmulL0 golden tolerance to 1e-4

lyfne123:
## Summary

- `matmul_autol0_64x128x256` (and sibling K=128 shapes) has been flaking in CI with single-element diffs ~1.08e-5 — barely above the strict `rtol=atol=1e-5` default.
- The error is not a bug: `AutoTileMatmulL0` splits K=128 into two k=64 chunks, so HW accumulates in a different order than `torch.matmul`'s BLAS reference. The worst-case FP32 rounding drift is bounded by `K * eps_fp32` (~1.5e-5 at K=128), which sits right at the 1e-5 boundary.
- Loosen golden tolerance to `rtol=1e-4 / atol=1e-5` for both `test_matmul_autol0` and `test_matmul_autol0_bf16` by passing a `RunConfig` through to `PTOTestCase`, and update the explanatory comment to reflect the new bound.

## Repro of the original flake

```
AssertionError: Output 'c' does not match golden.
Mismatched elements: 1/16384
rtol=1e-05, atol=1e-05
First 1 mismatches:
    [2943] actual=-0.006658013444393873, expected=-0.006668867543339729
```

## Test plan

- [ ] CI green on `test_matmul_autol0[*]` and `test_matmul_autol0_bf16[*]` parametrizations
- [ ] Verify the 1e-4 bound still catches a real correctness regression (e.g. a known-bad K-split would diff by orders of magnitude more than 1e-4 — sanity check during review)

### DIFF
--- a/tests/st/runtime/test_matmul.py
+++ b/tests/st/runtime/test_matmul.py
@@ -23,6 +23,7 @@ import pytest
 import torch
 from examples.kernels.matmul import matmul_acc_64
 from harness.core.harness import PLATFORMS, DataType, PTOTestCase, TensorSpec
+from pypto.runtime.runner import RunConfig
 
 
 class TestMatmul(PTOTestCase):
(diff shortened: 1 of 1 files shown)


Hzfengsy (approved):
Verified the high-priority bot concern: RunConfig tolerances DO reach golden.py via TestRunner._write_golden_for_test_case / _schedule_exec_after_golden, which rewrites golden.py with the real RunConfig before validate_golden reads RTOL/ATOL. The change is effective. The missing platform parametrize on test_matmul_autol0 is a pre-existing minor inconsistency, not blocking.

PR #1436, opened 2026-05-21
refactor(codegen): unify shape-Var collectors via a single C++ binding

### LINKED ISSUE #1419: [Code Health] Unify CollectVarsFromExpr (C++) and _collect_shape_dim_vars (Python) to prevent silent drift in PTO kernel wrapper dynamic-dim recovery
### Category

Technical Debt (cleanup, refactor)

### Component

Codegen

### Description

PTO kernel wrapper relies on two independently-implemented Var collectors that **must** walk the exact same shape-expression node set in the exact same order. Any divergence silently breaks the contract between the ptoas function signature and the wrapper's forward call.

- **C++ side** — `CollectVarsFromExpr` / `CollectTensorShapeDynVars` in `src/codegen/pto/pto_codegen.cpp` walks each tensor parameter's shape expressions and emits trailing `%argN: index` params on the `func.func` signature in first-seen order.
- **Python side** — `_collect_shape_dim_vars` in `python/pypto/backend/pto_backend.py` does the same walk inside `_generate_arg_unpacking` to decide which dynamic dims the wrapper recovers from `tensor->shapes[]` and in what order to pass them into the inner call.

If the two collectors diverge (e.g. someone adds a new `Expr` subclass that can appear in shape — say, a `SelectExpr` or a new arithmetic node — and only updates one side), the wrapper will silently pass arguments in the wrong positions or omit them entirely. The runtime then reads wrong `dim` values and can OOB without any explicit error from codegen.

Today the only safeguard is the cross-reference comments inside each function pointing at the mirror. That is a manual coupling: code review must catch any drift, and there is no CI signal.

Context: introduced in PR #1380 (recursive shape-expression Var collection + wrapper-side affine inversion for `pl.dynamic`). The PR itself is correct; this issue is the follow-up to address the structural duplication it leaves behind.

### Location

- `src/codegen/pto/pto_codegen.cpp:271-320` — `CollectVarsFromExpr` and `CollectTensorShapeDynVars`
- `python/pypto/backend/pto_backend.py:258-285` — `_collect_shape_dim_vars`
- `python/pypto/backend/pto_backend.py:131-179` — `_append_dynamic_dim_unpacking` (consumer that depends on order matching the C++ emission)

Commit: `72f77f350b863459b99a063a49c09d3407ae7e44`

### Proposed Fix

Open to discussion — listing the candidate directions for the team to choose from:

1. **Expose the C++ collector to Python via nanobind.** Eliminate `_collect_shape_dim_vars` entirely; have the wrapper codegen call the bound C++ function and iterate the result. This is the only option that *structurally* prevents drift, but it does add a binding surface and couples the wrapper to the C++ build.

2. **Add a checklist reminder.** Update `docs/en/dev/ir/` (and/or the `add-op` skill) to require simultaneous updates to both collectors when introducing a new `Expr` subclass that may appear in shape. Lightest-touch option; relies on humans following the checklist.

3. **Add a parity regression test.** Construct a sample IR containing every shape-expressible `Expr` node kind (Var / BinaryExpr / UnaryExpr / Call / TupleGetItemExpr / ConstInt / ConstFloat / ConstBool / etc.), run both collectors, and assert that the returned Var sequences match by `unique_id`. CI then fails the moment a new kind is added to one side without the other. Can be combined with option 2.

Option 1 is the most robust; option 3 is the lowest-friction safety net. Worth discussing whether to do option 3 immediately as a stopgap and queue option 1 as a follow-up.

### Priority

Medium (minor risk, should fix in next few releases)

lyfne123:
## Summary

PTO kernel wrapper codegen depended on two independently-implemented `Var` collectors that had to walk the exact same shape-expression node set in the exact same order:

- **C++** — `CollectVarsFromExpr` / `CollectTensorShapeDynVars` in `src/codegen/pto/pto_codegen.cpp`, driving the trailing `%argN: index` params on the emitted `func.func` signature.
- **Python** — `_collect_shape_dim_vars` in `python/pypto/backend/pto_backend.py`, driving the wrapper's recovery from `tensor->shapes[]` and forward-call argument order.

Any divergence (e.g. someone adds a new `Expr` subclass that can appear in a tensor shape and only updates one side) silently broke the contract: the wrapper passed args in the wrong positions or omitted them, and the runtime read wrong `dim` values. The only existing safeguard was cross-reference comments — a manual coupling with no CI signal.

This PR implements Option 1 from the issue: make the C++ walker the single source of truth.

## Changes

- Add public `CollectVarsFromShapeExpr(expr)` in `include/pypto/codegen/pto/pto_codegen.h`. Internal recursive helper renamed to `CollectVarsFromShapeExprImpl` (still reused by `CollectTensorShapeDynVars` for cross-param dedup).
- Expose it through the codegen nanobind module as `codegen.collect_vars_from_shape_expr(expr)`.
- Update `python/pypto/pypto_core/codegen.pyi` with the matching signature.
- Delete `_collect_shape_dim_vars` in `pto_backend.py`; the lone caller in `_append_dynamic_dim_unpacking` now calls the C++ binding. Cross-dim/param dedup-by-`unique_id` is preserved — it sits above the walker, not inside it.
- Update the surrounding C++ comment to drop the "two walkers must agree" warning.

Behavior is unchanged.

## Test plan

- [x] `cmake --build build --parallel` clean.
- [x] `pytest tests/ut/` — 4983 passed, 26 skipped, 0 failed.
- [x] Targeted: `tests/ut/codegen/test_dynamic_shape.py`, `tests/ut/codegen/test_pto_codegen.py::TestGenerateArgUnpacking`, `tests/ut/jit/test_decorator.py` — all green.

## Related issues

Fixes #1419

### DIFF
--- a/include/pypto/codegen/pto/pto_codegen.h
+++ b/include/pypto/codegen/pto/pto_codegen.h
@@ -44,6 +44,31 @@ struct TpopResultInfo {
 };
 
 /**
+ * @brief Collect Vars referenced by a tensor-shape expression, in first-seen DFS order.
+ *
+ * Used by:
+ *   - PTOCodegen, to emit trailing `%argN: index` params on `func.func` signatures
+ *     (see `CollectTensorShapeDynVars` in pto_codegen.cpp).
+ *   - The Python kernel-wrapper codegen, to recover dynamic dims from
+ *     `tensor->shapes[]` and forward them to the inner call in matching positional
+ *     order. This is the single source of truth shared by both paths: the wrapper
+ *     and the compiled function signature stay in lockstep by construction.
+ *
+ * Supported node kinds: Var / BinaryExpr / UnaryExpr / Call / TupleGetItemExpr /
+ * ConstInt / ConstFloat / ConstBool. Any other expression kind triggers an
+ * INTERNAL_CHECK failure. Adding a new shape-expressible `Expr` subclass requires
+ * updating only this function.
+ *
+ * Dedup key: raw `Var*` (sound because the IR holds the canonical shared_ptr
+ * graph, so each Var has exactly one address). The dedup scope is this single
+ * call; cross-expression dedup is the caller's responsibility.
+ *
+ * @param expr Tensor-shape expression (a dim from `TensorType::shape_`).
+ * @return Vars in first-seen DFS order, deduped within this single call.
+ */
+std::vector<ir::VarPtr> CollectVarsFromShapeExpr(const ir::ExprPtr& expr);
+
+/**
  * @brief PTO MLIR code generator
  *
  * Generates PTO-ISA MLIR format code from PyPTO IR Program.
--- a/python/bindings/modules/codegen.cpp
+++ b/python/bindings/modules/codegen.cpp
@@ -83,6 +83,17 @@ void BindCodegen(nb::module_& m) {
                      "    func: The function to infer core type for\n\n"
                      "Returns:\n"
                      "    CoreType.CUBE or CoreType.VECTOR");
+
+  codegen_module.def("collect_vars_from_shape_expr", &CollectVarsFromShapeExpr, nb::arg("expr"),
+                     "Collect Vars from a tensor-shape expression in first-seen DFS order.\n\n"
+                     "Used by the Python kernel-wrapper codegen to recover dynamic dims from "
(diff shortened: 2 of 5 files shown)


Hzfengsy (approved):
LGTM — unifies shape-Var collection on a single C++ source of truth.

Issue #1483, opened 2026-05-22
[Bug] (ST Flake) Swimlane DAG observability is not stable enough for exact fanout and branch-chain assertions


Leaf-Salix:
### Component

Other (please specify in description)

## Summary

> I found this while cleaning up the `phase-fence-dep-compression-passVer` ST coverage. My current plan is to fix the flaky swimlane assertions in this project at the same time, because the changes are test-side only and clarify the runtime swimlane observability contract. If this should be split out, I can keep the issue and patch separate.

Some runtime ST swimlane checks currently assume that the exported swimlane DAG is a complete and stable semantic dependency graph. In practice, a few fields and edge patterns are observability artifacts of the runtime trace, and they can vary between otherwise successful runs.

This shows up in two related ways:

1. Aggregate `fanout_count` checks can fluctuate for the same manual-scope witness.
2. `pl.at` branch-chain ordering checks are only reliable when the swimlane DAG exposes enough edges to reconstruct each per-branch chain.

The underlying programs can still execute correctly. The issue is that these ST assertions are stronger than the current swimlane observability contract.

## Affected tests

```text
tests/st/runtime/test_manual_scope_pipeline.py::TestManualScopeSwimlane::test_intra_iteration_dep_present
tests/st/runtime/test_pl_at_deps_pipeline.py::TestBranchChainPlAtSwimlane::test_intra_branch_linear_chain
```

The tests are usually run with:

```bash
pytest tests/st/runtime/test_manual_scope_pipeline.py::TestManualScopeSwimlane --enable-l2-swimlane -q
pytest tests/st/runtime/test_pl_at_deps_pipeline.py::TestBranchChainPlAtSwimlane --enable-l2-swimlane -q
```

## Suggested test-side handling

For manual-scope fanout checks:

```python
# Do not assert the exact aggregate fanout total here. The swimlane export is
# runtime-dependent and may expose 14/15/16 fanout edges for the same witness.
# Keep only the stable dependency-shape checks in this ST.
```

For `pl.at` branch-chain checks:

```python
chains = reconstruct_branch_chains(swimlane_dag)
if not chains_are_complete(chains):
    pytest.skip("swimlane DAG did not expose enough edges to reconstruct branch chains")

for prev, next_ in adjacent_pairs(chains):
    assert next_.start >= prev.end
```

## Why this should be tracked separately

This is not a semantic dependency bug in the tested programs. It is an observability-boundary issue in the swimlane trace used by STs.

[...]

Issue #1462, opened 2026-05-21
[Pass Bug] ScopeOutliner leaves stale seeded Var uses after freshened Var reconciliation


Leaf-Salix:
### Pass Name

OutlineIncoreScopes / ScopeOutliner

## Summary

> I found this while implementing the `phase-fence-dep-compression-passVer` change. With that branch, the existing test `tests/ut/jit/test_decorator.py::TestInlineFuncIntegration::test_nested_inline_diamond` fails with `Exception: Verification failed after 'ExpandManualPhaseFence' for properties {SSAForm}`. My current plan is to fix this ScopeOutliner issue in the same `phase-fence-dep-compression-passVer` PR and add the regression test below. If this should not be handled in the same PR, please leave a comment and I can split it out.

This is an IR transform bug in `ScopeOutliner`. It can produce an outlined AIV function whose signature uses a freshened `Var`, while the outlined body still references the old seeded `Var`. The resulting post-pass IR violates `SSAForm`: a variable is used inside the outlined function without being a parameter or a local definition.

This is exposed by a nested `@jit.inline` diamond where both branches call the same inline helper, and that helper contains a `with pl.at(level=pl.Level.CORE_GROUP)` scope that gets outlined.

## Minimal repro

```python
import pytest

import pypto.language as pl
from pypto.jit.decorator import jit
from pypto.pypto_core import passes


def test_inline_diamond_outlined_body_uses_freshened_signature_vars():
    torch = pytest.importorskip("torch")

    @jit.inline
    def shared(a: pl.Tensor, out: pl.Out[pl.Tensor]):
        with pl.at(level=pl.Level.CORE_GROUP):
            tile = pl.load(a, [0, 0], [32, 32])
            pl.store(tile, [0, 0], out)
        return out

    @jit.inline
    def a_helper(a: pl.Tensor, out: pl.Out[pl.Tensor]):
        out = shared(a, out)
        return out

    @jit.inline
    def b_helper(a: pl.Tensor, out: pl.Out[pl.Tensor]):
        out = shared(a, out)
        return out

    @jit
    def entry_diamond(a: pl.Tensor, out: pl.Out[pl.Tensor]):
        out = a_helper(a, out)
        out = b_helper(a, out)
        return out

    post_pass = entry_diamond.compile_for_test(torch.randn(32, 32), torch.empty(32, 32))

    props = passes.IRPropertySet()
    props.insert(passes.IRProperty.SSAForm)
[...]

PR #1486, opened 2026-05-22
feat(pass): compress manual_scope phase-fence deps with ExpandManualPhaseFence

Leaf-Salix:
## Summary

- Adds a dedicated `ExpandManualPhaseFence` IR pass that compresses profitable
  explicit `manual_scope` full-array `TaskId` dependencies before backend
  lowering.
- Rewrites profitable phase-fence fanout from
  `N producers -> M consumers`
  into
  `N producers -> dummy barrier -> M consumers`
  by inserting a single dependency-only `system.task_dummy`.
- Fixes the `range -> parallel` snapshot point so same-phase branches do not
  accidentally observe `TaskId` slots already updated by earlier branches in the
  same phase.
- Preserves direct fallback behavior for low-benefit, mixed, partial-slot,
  unresolved, pure range-only, and update-hazard dependency shapes.
- Includes a separate upstream `ScopeOutliner` SSA closure fix and regression
  test that were required to keep the validated pipeline correct.

This PR has two clearly separated parts:

1. the main feature: pass-led manual phase-fence compression
2. an additional upstream correctness fix: `ScopeOutliner` SSA closure

The second item is included because it was exposed during validation, but it is
not part of the phase-fence design itself.

## ExpandManualPhaseFence

This PR moves phase-fence dependency compression into a dedicated pass:

- it recognizes explicit `manual_scope` consumer sites with exactly one full
  `Array[TASK_ID]` manual dep;
- it inserts a `system.task_dummy` barrier IR node carrying the original
  dependency array;
- it rewrites the consumer site to depend on the barrier `TaskId`;
- it limits the transformation to profitable local fanout scopes in
  orchestration code.

The main semantic cleanup is the loop-entry snapshot for loop-carried phase
arrays. Instead of letting each branch expand `deps=[tids]` after earlier
branches have already written back into `tids`, the pass builds one dependency
point before the current phase fanout:

```text
old tids[N] -> dummy barrier -> current phase consumers[M] -> next tids
```

That keeps the dependency source aligned with the intended previous phase rather
than accidentally introducing same-phase serialization.

One caveat worth calling out: a shared outer `tids` carrier is not implicitly
lifted into per-lane ownership. In other words,
`tids -> para { range { para { deps=[tids] } } }`
does not mean
`para { tids -> range { para { deps=[tids] } } }`;
independent outer parallel lanes would require the carrier itself to be scoped
or indexed per lane.

## Profitability And Fallback Boundaries

This pass is intentionally narrow rather than a general DAG optimizer.

Current behavior:

- profitable full-array fanout compresses;
- minimum-profitable `3 -> 3` fanout compresses;
- low-benefit `N -> 1` and `2 -> 2` stay direct;
- pure `range -> range` repeated consumers stay direct;
- mixed deps, multiple arrays, unresolved arrays, auto-scope deps, and
  same-body update hazards stay direct;
- partial-slot deps such as `prev = tids[i]; deps=[prev]` remain guarded scalar
  direct deps rather than barrier-compressed deps;
- non-orchestration functions are left untouched.

This keeps the recognition range explainable and avoids inventing barriers in
shapes that are not clearly profitable or safe.

In practice, when the pass cannot recognize one clear, profitable full-array
manual phase-fence pattern, it keeps the existing direct dependency path.

### Additional Upstream Fix: ScopeOutliner SSA Closure

This branch also includes an upstream correctness fix in
`scope_outline_utils.h`, and this part should be called out separately in the
PR.

While validating the pass, a pre-existing `ScopeOutliner` bug showed up in a
diamond-inline repro:

- the outlined function signature used a freshened `InOut` var such as
  `out__ssa_v1`;
- the outlined body still kept a stale store target such as `out__ssa_v0`;
- later SSA verification then failed because the body was no longer closed over
  the final signature vars.
- the upstream issue is tracked separately as `hw-native-sys/pypto#1462`:
  https://github.com/hw-native-sys/pypto/issues/1462

The fix canonicalizes external scope vars through store-target renames during
input/output analysis and bridges old body aliases back to the final freshened
input params so outlined-body substitution can fully close over the final var
set.

This PR adds a focused regression test:

- `tests/ut/jit/test_scope_outliner_ssa_closure_repro.py`

That repro is intentionally independent from the phase-fence pass so the
upstream bug and its fix stay separately explainable, and it serves as the
branch-local repro/coverage point for `hw-native-sys/pypto#1462`.

### Runtime Test Sync Notes

The branch also syncs a few older `manual_scope` / `pl.at` runtime swimlane
tests away from stale pre-compression assumptions: phase-fence witnesses no
longer require direct all-to-all fanout observability, and the outlined
`pl.at` witness now stays within the properties the swimlane artifact can
actually expose. The underlying swimlane observability issue is tracked
separately as `hw-native-sys/pypto#1483`:
https://github.com/hw-native-sys/pypto/issues/1483

## Tests

Validated on this branch with the following emphasis:

- UT covers pass behavior, emitted codegen shape, and fallback boundaries
- runtime ST covers execution correctness and strict phase ordering
- swimlane checks only assert properties the artifact can expose reliably

Representative commands that passed on macOS:

- `pytest tests/ut -n auto --maxprocesses 8 -v`
- `pytest tests/ut/ir/transforms/test_expand_manual_phase_fence.py`
- `pytest tests/ut/codegen/test_phase_fence_dep_compression.py`
- `pytest tests/ut/codegen/test_orchestration_codegen.py -k "phase_fence or manual_scope"`
- `pytest tests/ut/jit/test_scope_outliner_ssa_closure_repro.py`
- `pytest tests/st/runtime/test_manual_scope_pipeline.py -q --platform=a2a3`
  - `TestManualScopeSwimlane::test_total_task_count`
  - `TestManualScopeSwimlane::test_intra_iteration_dep_present`
  - `TestManualScopeSwimlane::test_no_blocking_serialization_chain`
- `pytest tests/st/runtime/test_phase_fence_dep_compression.py -q --platform=a2a3`
  - `TestPhaseFenceDepCompressionSwimlane::test_multiloop_chain_default`
  - `TestPhaseFenceDepCompressionSwimlane::test_submit_three_level_strict`
- `pytest tests/st/runtime/test_pl_at_deps_pipeline.py -q --platform=a2a3`
  - `TestPlAtDepsSwimlane::test_total_task_count`
  - `TestPlAtDepsSwimlane::test_no_blocking_serialization_chain`
  - `TestPhaseFencePlAtSwimlane::test_total_task_count`
  - `TestPhaseFencePlAtSwimlane::test_phase_fence_strict`
  - `TestBranchChainPlAtSwimlane::test_total_task_count`
  - `TestBranchChainPlAtSwimlane::test_no_cross_branch_fanout`
- `PYPTO_PHASE_FENCE_EXTRA_SWIMLANE=1 pytest tests/st/runtime/test_phase_fence_dep_compression.py::TestPhaseFenceDepCompressionSwimlane::test_multiloop_chain_default --enable-l2-swimlane -q --platform=a2a3`
- `PYPTO_PHASE_FENCE_EXTRA_SWIMLANE=1 pytest tests/st/runtime/test_phase_fence_dep_compression.py::TestPhaseFenceDepCompressionSwimlane::test_submit_three_level_strict --enable-l2-swimlane -q --platform=a2a3`
- `PYPTO_PHASE_FENCE_EXTRA_SWIMLANE=1 pytest tests/st/runtime/test_phase_fence_dep_compression.py::TestPhaseFenceDepCompressionSwimlane::test_pl_at_three_level_strict --enable-l2-swimlane -q --platform=a2a3`

These cover the key shapes: profitable full-array phase-fence compression,
loop-entry snapshot correctness, low-benefit and mixed-dep fallback,
partial-slot scalar direct deps, strict runtime phase ordering after dummy
barrier insertion, and the independent upstream ScopeOutliner SSA closure
regression.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,6 +149,7 @@ set(PYPTO_SOURCES
     src/ir/transforms/collect_comm_groups_pass.cpp
     src/ir/transforms/optimize_orch_tensors_pass.cpp
     src/ir/transforms/derive_call_directions_pass.cpp
+    src/ir/transforms/expand_manual_phase_fence_pass.cpp
     src/ir/transforms/convert_to_ssa_pass.cpp
     src/ir/transforms/ctrl_flow_transform_pass.cpp
     src/ir/transforms/flatten_call_expr_pass.cpp
--- a/include/pypto/ir/expr.h
+++ b/include/pypto/ir/expr.h
@@ -693,6 +693,16 @@ inline std::vector<std::pair<std::string, std::any>> WithArgDirectionOverridesAt
 inline constexpr const char* kAttrManualDepEdges = "manual_dep_edges";
 
 /**
+ * @brief Reserved attr key marking an internal dependency-only TaskId call.
+ *
+ * Value type: ``bool``. Written by the manual phase-fence expansion pass on
+ * ``system.task_dummy`` calls. Orchestration codegen lowers the marked call to
+ * ``rt_submit_dummy_task(...)`` and uses ``manual_dep_edges`` on that same call
+ * as the dummy task fanin.
+ */
+inline constexpr const char* kAttrDummyTask = "dummy_task";
+
+/**
  * @brief Reserved attr key for the producer-TaskId Var captured by a
  * ``with pl.at(...) as tid:`` block.
  *
--- a/include/pypto/ir/transforms/pass_properties.h
+++ b/include/pypto/ir/transforms/pass_properties.h
@@ -242,6 +242,12 @@ inline const PassProperties kCanonicalizeIOOrderProperties{
 inline const PassProperties kDeriveCallDirectionsProperties{.required = {IRProperty::SplitIncoreOrch},
                                                             .produced = {IRProperty::CallDirectionsResolved}};
 
+inline const PassProperties kExpandManualPhaseFenceProperties{
+    .required = {IRProperty::SSAForm, IRProperty::NoNestedCalls, IRProperty::NormalizedStmtStructure,
+                 IRProperty::CallDirectionsResolved},
+    .produced = {IRProperty::SSAForm, IRProperty::NoNestedCalls, IRProperty::NormalizedStmtStructure,
+                 IRProperty::CallDirectionsResolved}};
+
 // -- No-op tile.reshape folding pass -----------------------------------------
 
 inline const PassProperties kFoldNoOpReshapeProperties{
--- a/include/pypto/ir/transforms/passes.h
+++ b/include/pypto/ir/transforms/passes.h
@@ -663,6 +663,16 @@ Pass FuseCreateAssembleToSlice();
 Pass DeriveCallDirections();
 
 /**
+ * @brief Expand profitable manual_scope Array[TASK_ID] fanout deps into
+ *        explicit dependency-only dummy barrier calls.
+ *
+ * Rewrites selected consumer ``manual_dep_edges=[source_array]`` to
+ * ``manual_dep_edges=[barrier_tid]`` after inserting a marked
+ * ``system.task_dummy`` assignment at the chosen phase-fence placement point.
+ */
+Pass ExpandManualPhaseFence();
+
+/**
  * @brief Fold no-op tile.reshape assignments into Var-to-Var assignments
  *
  * After LegalizePTOBufferReuse, two TileType variables can share the same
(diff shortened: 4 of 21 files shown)


Leaf-Salix:
Closing this PR for now because the fix for [#1462](https://github.com/hw-native-sys/pypto/issues/1462) has already landed in [#1492](https://github.com/hw-native-sys/pypto/pull/1492). I need to remove the overlapping changes from my commits, rebase or merge the latest version, and then re-test the remaining changes before reopening.

PR #1518, opened 2026-05-26
bisect: revert fc0f927b (pl.cmp wrappers) + AICPU fix

Little-oil:
## Purpose

This branch reverts **fc0f927b** (`feat(language): unified pl.cmp and pl.set_validshape dispatch wrappers (#1372)`) on top of origin/main. The AICPU plumbing fix is intentionally NOT included — earlier bisect showed all 4 branches failed with 507018 because the fix itself masks the original error. Reverting cleanly exposes whether this commit is the root cause.

## Commits
- revert of fc0f927b

## Read the signal
- daily-CI **passes** → fc0f927b is the cause
- daily-CI **fails with the original error** → fc0f927b is innocent


Little-oil:

PR #1519, opened 2026-05-26
bisect: revert 382293cb (NormalizeStmtStructure merge) + AICPU fix

Little-oil:
## Purpose

This branch reverts **382293cb** (`refactor(passes): merge NormalizeStmtStructure and manual-scope lowering (#1365)`) on top of origin/main. The AICPU plumbing fix is intentionally NOT included — earlier bisect showed all 4 branches failed with 507018 because the fix itself masks the original error. Reverting cleanly exposes whether this commit is the root cause.

## Commits
- revert of 382293cb (resolved conflicts with `-X theirs` + restored 4 files deleted by the merge: `derive_manual_scope_deps_pass.cpp`, doc + test counterparts)

## Read the signal
- daily-CI **passes** → 382293cb is the cause
- daily-CI **fails with the original error** → 382293cb is innocent


Little-oil:

PR #1520, opened 2026-05-26
bisect: revert 91782e66 (SPMD block_idx/num refactor) + AICPU fix

Little-oil:
## Purpose

This branch reverts **91782e66** (`refactor(backend): thread SPMD block_idx/num through wrapper-passed params (#1363)`) on top of origin/main. The AICPU plumbing fix is intentionally NOT included — earlier bisect showed all 4 branches failed with 507018 because the fix itself masks the original error. Reverting cleanly exposes whether this commit is the root cause.

## Commits
- revert of 91782e66 (resolved conflicts with `-X theirs`)

## Read the signal
- daily-CI **passes** → 91782e66 is the cause
- daily-CI **fails with the original error** → 91782e66 is innocent


Little-oil:

Issue #1338, opened 2026-05-12
[Feature] RFC: numpy/torch-style rank-reducing index/slice for Tensor and Tile in the DSL


Hzfengsy:
### Summary

Adopt numpy/torch-style subscript semantics for `Tensor` (DDR) and `Tile` (on-chip) in the PyPTO DSL:

- **Rank-reducing scalar index** — a scalar index removes its dimension instead of erroring or producing a size-1 dim. `C[i, j]` on a `64×64×64×64` tensor yields a `64×64` view (today: error "requires 4 indices").
- **Partial / chained indexing** — fewer indices than `rank` implies trailing `:`. `C[i]` → 3D view; `C[i][j]` → 2D view; `C[i:i+8, j]` → `8×64×64`.
- **Unified Tensor/Tile semantics** — same rules for both, with one tile-specific adjustment (see below).

Mechanism: extend the existing `tensor.slice` / `tile.slice` IR ops with an optional `drop_dims: list[int]` parameter rather than adding a new op.

### Motivation / Use Case

Today (`python/pypto/language/parser/ast_parser.py`, `_parse_tensor_subscript` / `_parse_tile_subscript`):

- Subscript requires **exactly `rank` indices** — `C[i]` on a 4D tensor is a hard error.
- A scalar index inside a mixed subscript becomes a **size-1 dim**, not a dropped dim — `C[i, j, :, :]` gives `1×1×64×64`, not `64×64`.
- All-scalar full-rank index → scalar via `tensor.read` / `tile.read`.
- `C[i][j]` fails because the inner `C[i]` is itself an invalid 1-index subscript.

So extracting a `64×64` plane from a 4D tensor requires writing all four indices *and* then a manual reshape/squeeze to drop the leading `1, 1`. This is verbose, surprising to anyone with numpy/torch background, and inconsistent between the "all scalar" path (rank-reduces to a scalar) and the "mixed" path (keeps size-1 dims). Unifying on numpy semantics makes tensor/tile views composable (`C[i][j]`, partial indexing) and removes the boilerplate.

### Proposed API / Behavior

**User-facing semantics** (4D `C` of shape `64⁴`):

| Expression | Today | Proposed |
| --- | --- | --- |
| `C[i, j, k, l]` (all scalar, full rank) | scalar via `tensor.read` | unchanged — scalar |
| `C[i, j]` (partial, all scalar) | error | `64×64` view |
| `C[i]` | error | `64×64×64` view |
[...]

PR #1521, opened 2026-05-26
bisect: revert ec6482c6 (rank-reducing index/slice) + AICPU fix

Little-oil:
## Purpose

This branch reverts **ec6482c6** (`feat: numpy-style rank-reducing index/slice for Tensor & Tile (RFC #1338) (#1343)`) on top of origin/main. The AICPU plumbing fix is intentionally NOT included — earlier bisect showed all 4 branches failed with 507018 because the fix itself masks the original error. Reverting cleanly exposes whether this commit is the root cause.

## Commits
- revert of ec6482c6 (resolved conflicts with `-X theirs`)

## Read the signal
- daily-CI **passes** → ec6482c6 is the cause
- daily-CI **fails with the original error** → ec6482c6 is innocent


Little-oil:

Issue #1569, opened 2026-05-28
[Bug] scf.if else branch yields raw ptr instead of tensor_view inside pl.spmd


zhangqi-chen:
### Component

Codegen

### Description

When `pl.spmd` body contains an `if` (with no `else` user code) that performs a conditional slice-assign or `pl.assemble` write to an outer-scope tensor passed into the spmd-outlined function, pypto codegen emits an `scf.if` whose `if` branch yields a `tensor_view` (via `pto.partition_view`) while the empty `else` branch yields the raw `!pto.ptr<f32>` function argument. The `scf.if` itself is typed `!pto.tensor_view<?x?xf32>`, so ptoas rejects the MLIR.

### Steps to Reproduce

Save the minimal repro (~30 lines) and run:

```bash
python repro_spmd_if_yield.py -p a2a3sim
```

Repro file:

```python
import pypto.language as pl

N_BLOCKS = 32
N_COLS = 16


@pl.jit
def repro_spmd_if_write(
    flag: pl.Tensor[[N_BLOCKS], pl.INT32],
    out: pl.Out[pl.Tensor[[N_BLOCKS, N_COLS], pl.FP32]],
):
    for idx in pl.spmd(N_BLOCKS, name_hint="spmd_if_write"):
        cond = pl.read(flag, [idx])
        if cond > 0:
            tile = pl.full([1, N_COLS], dtype=pl.FP32, value=1.0)
            out[idx : idx + 1, :] = tile
```

Full standalone harness (with golden + `run_jit` boilerplate) lives at `pypto-lib models/deepseek/v4/repro_spmd_if_yield_draft.py` on branch `perf-cmp128` (commit `9f6641ff`). The same pattern is also currently in the dsv4 decoder `models/deepseek/v4/decode_compressor_ratio128.py` `softmax_pool` scope (same branch).

### Expected Behavior

Compilation succeeds. Both branches of `scf.if` should yield a value of the declared `!pto.tensor_view<?x?xf32>` type (e.g. the unmodified `%out__ssa_v0_view` produced earlier by `pto.make_tensor_view %arg1`).

### Actual Behavior

```
Failed to compile function 'spmd_if_write': ptoas compilation failed:
loc("..../spmd_if_write.pto":22:15): error: use of value '%arg1'
expects different type than prior uses:
'!pto.tensor_view<?x?xf32>' vs '!pto.ptr<f32>'
```

[...]

Issue #1533, opened 2026-05-26
[Pass Bug] pto_backend emits pto.partition_view against SCF-phi tensor_view (loses static dims; ptoas rejects)


zhangqi-chen:
### Pass Name

pto_backend (pypto → ptoas IR lowering, partition_view emission via SCF)

### Description

After locally landing the fix for pypto#1532, fusing QK+softmax+PV in the dsv4 sparse-attention kernel into a single `pl.at(name_hint=\"qk_pv\")` scope still fails at ptoas with:

```
ptoas compilation failed: ...qk_pv.pto:115:48: error: 'pto.partition_view' op
operand #0 must be , but got 'memref<?x?xf32, strided<[?, ?], offset: ?>>'
```

The failing op is a `pto.partition_view` whose input is the SCF-yielded phi version of an output tensor that is `pl.assemble`'d inside an `if` branch within a `pl.range` loop. The lowering emits `!pto.tensor_view<?x?xf32>` for the phi — fully dynamic dims/strides/offset — but the tensor's static shape is known at the python level (`pl.create_tensor([n_blocks * H_TILE, HEAD_DIM], dtype=pl.FP32)`). ptoas's `pto.partition_view` op only accepts a concrete typed input.

In the same scope, `partition_view` ops on function-parameter inputs (e.g. `sparse_kv`) succeed even though their `tensor_view` type is also `<?x?xbf16>` — see lines 94 and 98 in the IR excerpt below. The breakage is specific to **phi-typed scratch tensors after SCF-if lowering**.

### Git Commit ID

b0057a391207049a13ff2318aee9487f78c3ff6a (plus user's local fix for pypto#1532)

### Before IR (Input)

Production source — the merged `qk_pv` scope in `models/deepseek/v4/decode_sparse_attn.py`:

```python
# Inside `for attn_t0 in pl.parallel(T): for h0 in pl.parallel(0, H, H_TILE):`
sparse_blk_mi = pl.create_tensor([((TOPK + SPARSE_ATTN_TILE - 1) // SPARSE_ATTN_TILE) * H_TILE, 1], dtype=pl.FP32)
sparse_blk_li = pl.create_tensor([((TOPK + SPARSE_ATTN_TILE - 1) // SPARSE_ATTN_TILE) * H_TILE, 1], dtype=pl.FP32)
[...]

PR #1574, opened 2026-05-28
test(codegen): add spmd regression guard for IfStmt tensor phi (#1569)

lyfne123:
## Summary

Issue #1569 (`scf.if` else branch yields a raw `!pto.ptr` instead of a
`tensor_view` inside `pl.spmd`) is the same root cause as #1533 and is
**already resolved** by PR #1555, which keeps `TensorType` `IfStmt`
return_vars out of `scf.if` results. `IfStmt` codegen is shared across
control structures, so that fix also covers the `pl.spmd`-outlined
per-block kernel.

This PR adds a **spmd-specific regression guard** — #1555's regression
test only covered the `pl.range` / cross-`if` InCore form.

## Change

`SpmdIfTensorReturnScratch` in `tests/ut/codegen/test_if_tensor_return_partition_view.py`:
a conditional slice-assign to an outer tensor inside a `pl.spmd` body.
The `pl.spmd` body is outlined into a per-block kernel whose `if`
produces a `TensorType` `IfStmt` phi. The three new tests assert codegen:

- compiles through the pipeline + PTO codegen,
- emits **no** `scf.if -> (tensor_view)` result, and
- sources `pto.partition_view` from the concrete `make_tensor_view`, not a
  fully-dynamic `scf.if` phi.

## Verification

- The exact `for idx in pl.spmd(N)` repro from #1569 now compiles cleanly
  through ptoas on current main.
- New tests pass; existing #1533 tests in the same file unaffected
  (6 passed).

Refs #1569, #1533, #1555.

### DIFF
--- a/tests/ut/codegen/test_if_tensor_return_partition_view.py
+++ b/tests/ut/codegen/test_if_tensor_return_partition_view.py
@@ -224,9 +224,7 @@ def test_spmd_no_tensor_view_scf_if_result(spmd_scratch_mlir: str):
 def test_spmd_partition_view_not_on_scf_phi(spmd_scratch_mlir: str):
     """partition_view inside the spmd kernel must source the concrete base view, not a phi."""
     sources = [
-        _partition_view_source(line)
-        for line in spmd_scratch_mlir.split("\n")
-        if "pto.partition_view" in line
+        _partition_view_source(line) for line in spmd_scratch_mlir.split("\n") if "pto.partition_view" in line
     ]
     assert sources, f"Expected at least one pto.partition_view:\n{spmd_scratch_mlir}"
     bad = [src for src in sources if "__phi" in src]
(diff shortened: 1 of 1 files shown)


lyfne123:
Re CodeRabbit's suggestion to change the `spmd_scratch_mlir` fixture target to `"orchestrator"`: the fixture intentionally targets `"spmd_kernel"`. That is the InCore function containing the `scf.if` and `pto.partition_view` ops this test guards. The bug (#1569) manifests there — control-flow lowering turns the conditionally-written outer tensor into a TensorType IfStmt phi, and a later `partition_view` on that phi is what ptoas rejects. The `orchestrator` function only dispatches to `spmd_kernel` via `pl.spmd`; it produces no `partition_view` ops, so testing it would not catch a regression in the InCore kernel codegen.

Hzfengsy (approved):

PR #1597, opened 2026-05-29
Add opt-in out-window externalization switch

sunkaixuan2018:
## Summary
- Add a default-off `enable_out_window_externalization` compile/pass switch.
- Thread the switch through `PassContext`, Python `ir.compile`, `RunConfig`, JIT forwarding, and the JIT cache key.
- Keep the conservative dependency guard by default; only enable the windowed externalization path when explicitly requested.

## Background
The out-window behavior from d30df60 helped avoid serialized Qwen3 decode execution, but it was reverted in f8731b2 because cross-scope cases could lose dependencies, as tracked by issue 1463. This PR restores the behavior behind an explicit opt-in switch so existing users keep the safer default.

## Testing
- `python -m py_compile python/pypto/ir/compile.py python/pypto/ir/pass_manager.py python/pypto/jit/cache.py python/pypto/jit/decorator.py python/pypto/runtime/runner.py tests/ut/ir/transforms/test_optimize_orch_tensors.py tests/ut/ir/transforms/test_pass_pipeline.py tests/ut/jit/test_cache.py tests/ut/jit/test_decorator.py`
- Remote focused pytest on `/data/sunkaixuan/all_pyptos/out-window-enable`: 4 passed.
- Remote Qwen3 `models/qwen3/14b/decode_fwd.py -p a2a3 --num-layers 2 --enable-l2-swimlane` validation:
  - baseline and modified default-off: no `__windowed` kernels emitted.
  - modified opt-in: `q_proj__windowed`, `k_proj__windowed`, `v_proj__windowed`, `qk_norm__windowed`, etc. emitted.
  - full swimlane runs completed for baseline, modified default-off, and modified opt-in.

### DIFF
--- a/docs/en/dev/passes/13-optimize_orch_tensors.md
+++ b/docs/en/dev/passes/13-optimize_orch_tensors.md
@@ -95,6 +95,7 @@ Safety rules:
 - multi-`Out` rewrite is all-or-nothing
 - sequential-loop siblings are rewritten only when every rewritten `Out` can be proven disjoint across sibling iterations
 - call-site externalization is skipped when a rewritten window output's buffer root is later read as a full-parent tensor in the enclosing orchestration scope; until the runtime has root-aware view/parent dependency tracking, this preserves auto dependency tracking on the same parent `Tensor`
+- callers can explicitly opt into those guarded call-site rewrites with `enable_out_window_externalization=True` on `PassContext`, `ir.compile()`, or runtime `RunConfig`; the default remains conservative
 - `DeriveCallDirections` keeps its existing sound sequential `Out -> InOut` rule; Pattern 5 only makes disjoint windows explicit before that pass runs
 
 ## Example (Pattern 1)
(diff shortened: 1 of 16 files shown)


sunkaixuan2018:
CI note: the only failing check on this PR is `pypto-lib-model`, and the failure happens during the Qwen3 `decode_layer.py` runtime phase with `halMemCtl failed with rc=13` followed by `aclrtSynchronizeStreamWithTimeout (AICPU) failed: 507018`.

This appears to match the current `main` CI failure rather than a regression from this PR:
- PR run: https://github.com/hw-native-sys/pypto/actions/runs/26641243600/job/78514594499
- main run at e3a5c96d: https://github.com/hw-native-sys/pypto/actions/runs/26639731030/job/78509223346

Both runs compile the model successfully and fail in the same device runtime path. Other PR checks are green, and the new out-window behavior remains default-off.

PR #1607, opened 2026-05-30
fix(verifier): reject unwritten Out/InOut params, not just tensor-creating reassigns (#1525)

### LINKED ISSUE #1525: [DX] pypto silently emits all-zero when a pl.Out is reassigned to a detached value in a mix/split CORE_GROUP scope — should be a compile-time diagnostic
### Diagnosis

**pypto** — compiles clean, runs on a2a3, output is all zeros. Distinct from the matmul_acc crash #1523 (single matmul, no assert; silent miscompile). Shares a root with #1507 (cube→vec valid_shape loss) but no slice involved and adds acc-reuse, so neither contains the other.

### Description

Two chained matmuls in one `CORE_GROUP` scope — cube → vec cast → cube — compile but produce all-zero output on device (4094/4096 mismatch, `actual=0.0`). Each weight is a distinct param (transpose pass#198 clears). Final IR (`37_after_Simplify`):

- Stage1 `t1 = a@w` and stage2 `y = t1b@w2` both write the **same** `mem_acc_9` (offset 0) → 2nd matmul overwrites 1st before consume.
- AIV side: stage2 `tpop_from_aic` carries `valid_shape=[0,0]` → `cast` emits empty tile → `t1b` is zero → `y` zero. No `pl.slice(valid_shape=...)` anywhere.

~40-line repro:

\`\`\`python
import pypto.language as pl
M = K = N = 64
@pl.program
class Repro:
    @pl.function(type=pl.FunctionType.Opaque)
    def fused(self, a: pl.Tensor[[M,K],pl.BF16], w: pl.Tensor[[K,N],pl.BF16],
              w2: pl.Tensor[[K,N],pl.BF16], y: pl.Out[pl.Tensor[[M,N],pl.FP32]]) -> pl.Tensor[[M,N],pl.FP32]:
        with pl.at(level=pl.Level.CORE_GROUP, optimizations=[pl.split(pl.SplitMode.NONE)]):
            t1 = pl.matmul(a, w, out_dtype=pl.FP32)
            t1b = pl.cast(t1, target_type=pl.BF16, mode="rint")
            y = pl.matmul(t1b, w2, out_dtype=pl.FP32, b_trans=True)
        return y
\`\`\`

Real kernel: deepseek/v4 decode_indexer rope_slice+rope_assemble cannot mix-fuse for this reason. Related: #1507, #1523, #1352.

### Environment

| Component | Version |
|---|---|
| pypto-lib | \`9b95926\` |
| pypto | \`78e44c6\` (branch: main) |
| pypto runtime (submodule) | \`e03faff\` |
| pto-isa | \`0a1ce52\` |
| ptoas | 0.41 |
| CANN | 9.0.0 |

### Host Platform

Linux (aarch64)

lyfne123:
## Summary

Fixes the real root cause of #1525: a bare reassignment of a `pl.Out`/`pl.InOut` parameter to a **detached** value (e.g. `y = pl.matmul(...); return y`) rebinds the local name and leaves the external output buffer unwritten, so the kernel silently produces its init value (all-zero). This is a front-end output-binding footgun — **not** a device/pto-isa bug (the earlier device-side diagnosis was walked back; pto-isa#149 was closed as not-reproduced).

## What changed

**Generalize the `OutParamNotShadowed` verifier** (`src/ir/verifier/verify_out_param_pass.cpp`) from a tensor-creating-op blacklist (`tensor.create`/`tensor.full` only) to the precise invariant:

> A call-valued reassignment of an Out/InOut param is invalid **unless the call threads the param through its arguments**.

- ✅ Threading reassigns stay valid: `out = pl.assemble(out, tile, off)`, `out = pl.store(value, off, out)`, indexed write `out[...] = value`.
- ❌ Detached reassigns now fail compilation with a clear diagnostic: `out = pl.matmul(...)`, `out = tensor.create(...)`.

**Make the check actually fire**: add `OutParamNotShadowed` to `GetVerifiedProperties()` (`src/ir/transforms/ir_property.cpp`). It was already in the *structural* set but absent from the *verified* set, so the inline pipeline-input check skipped it and the footgun compiled silently. It now runs **once at pipeline input (pre-SSA)** — where the bare reassignment's `AssignStmt.var_` still targets the param Var (post-SSA it detaches into a dead value).

### Why not a broader "no `=` to parameters" rule
Refuted empirically: an AST scan of `examples/`+`tests/` found **889** bare-name `=` assignments to parameters, **741** of which are the blessed accumulator idiom `param = pl.store(t, off, param)` / `pl.assemble`. The distinguishing factor is whether the param is threaded through the rhs, not whether `=` is used.

## Validation

- Bare #1525 form now raises a `VerificationError` at compile time with an actionable message; the idiomatic form compiles clean (no false positive).
- Full `tests/ut` suite green (5346 passed), no false positives.
- New regression tests in `tests/ut/ir/transforms/test_verifier.py` (matmul + InOut cases).
- `examples/utils/error_handling.py` and the #1525 device test updated to the new behavior / supported idiom.
- Verifier docs updated (en + zh).

Closes #1525.

### DIFF
--- a/docs/en/dev/passes/99-verifier.md
+++ b/docs/en/dev/passes/99-verifier.md
@@ -72,7 +72,7 @@ The `run_verifier()` utility creates a standalone `Pass` for ad-hoc use in custo
 | **IncoreTileOps** | IncoreTileOps | InCore functions use tile ops (no tensor-level ops remain) |
 | **HasMemRefs** | HasMemRefs | All TileType variables have MemRef initialized |
 | **AllocatedMemoryAddr** | AllocatedMemoryAddr | All MemRefs have valid addresses within buffer limits |
-| **OutParamNotShadowed** | OutParamNotShadowed | Out/InOut params not reassigned to a detached value — a call that does not thread the param (e.g. `out = pl.matmul(...)`, `out = tensor.create(...)`) silently leaves the external buffer unwritten (issue #1525). Threading reassignments like `out = pl.assemble(out, ...)` are allowed |
+| **OutParamNotShadowed** | OutParamNotShadowed | Out/InOut params not reassigned to a detached value — a call whose result is a fresh value (`out = pl.matmul(...)`, `out = tensor.create(...)`, or even `out = pl.add(out, x)` which only reads the param) silently leaves the external buffer unwritten (issue #1525). Allowed: output-side ops threading the param through their target arg (`out = pl.assemble(out, ...)`, `out = pl.store(v, idx, out)`) and user-function calls writing it via an Out/InOut arg (`out = self.kernel(a, out)`) |
 | **NoNestedInCore** | NoNestedInCore | No nested InCore scopes (`InCoreScopeStmt` inside `InCoreScopeStmt`) |
 | **InOutUseValid** | InOutUseValid | Variables passed as InOut/Out to user-function calls are not read after the call (RFC #1026). Group-typed function bodies are skipped pending follow-up. |
 | **PipelineLoopValid** | PipelineLoopValid | Bidirectional invariant on every `ForStmt`: `kind_ == ForKind::Pipeline` ⇔ `pipeline_stages` attr present. Either direction failing means the pipeline loop is malformed. |
--- a/docs/en/dev/passes/99-verifier.md
+++ b/docs/en/dev/passes/99-verifier.md
@@ -72,7 +72,7 @@ The `run_verifier()` utility creates a standalone `Pass` for ad-hoc use in custo
 | **IncoreTileOps** | IncoreTileOps | InCore functions use tile ops (no tensor-level ops remain) |
 | **HasMemRefs** | HasMemRefs | All TileType variables have MemRef initialized |
 | **AllocatedMemoryAddr** | AllocatedMemoryAddr | All MemRefs have valid addresses within buffer limits |
-| **OutParamNotShadowed** | OutParamNotShadowed | Out/InOut params not reassigned with tensor-creating ops |
+| **OutParamNotShadowed** | OutParamNotShadowed | Out/InOut params not reassigned to a detached value — a call that does not thread the param (e.g. `out = pl.matmul(...)`, `out = tensor.create(...)`) silently leaves the external buffer unwritten (issue #1525). Threading reassignments like `out = pl.assemble(out, ...)` are allowed |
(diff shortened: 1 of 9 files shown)


lyfne123:
Thanks @coderabbitai and @gemini-code-assist — addressed in 18264aa.

**coderabbit (write-through vs any read):** valid and now fixed. I confirmed on a board-equivalent codegen dump that `out = pl.add(out, x)` (out: `pl.Out`) produces `add_input(ext_out)` + a detached `add_output(ret0__out)` — i.e. the external buffer is never written, the same silent-zero footgun as #1525. `CallThreadsParam` now mirrors the orchestration codegen's `TraceReturnedToParam`:
- builtin output-side ops alias **only their target arg** (`tensor.assemble`→args[0], `tile.store`→args[2], `tensor.set_validshape`→args[0]);
- any other builtin op (`tensor.add`/`tensor.matmul`/…) produces a fresh value → flagged even when the param is an input operand;
- user-function calls may write the param through an Out/InOut arg, so the param appearing in the args still counts as threading.

I did **not** apply a pure "write-target position" rule globally, because that would false-positive on the legitimate user-function threading idiom `out = self.kernel(a, out)` (84+ occurrences across examples/tests). Mirroring `TraceReturnedToParam` keeps the verifier consistent with what codegen actually binds. Validated against the full unit suite: **zero false positives** (the 182 `param = pl.store(...)` / 151 `pl.assemble(...)` accumulator reassigns and the user-call threading all still pass), and `out = pl.add(out, x)` is now caught. Added a regression test (`test_out_param_inplace_read_op_detected`).

**gemini (CallPtr → ExprPtr / intermediate-variable loophole):** acknowledged. A non-`Call` rebind such as `tmp = pl.matmul(...); out = tmp` is not yet flagged. Closing it correctly requires aliasing/lineage tracing through `out = tmp` (codegen does this via `VarLineageCollector`), and a naive "flag any `out = <var>`" would false-positive on legit chains like `tmp = pl.assemble(out, …); out = tmp`. I've left this as a follow-up rather than risk false positives here; the high-frequency footgun (`out = <op>(...)`) is covered.

**coderabbit (example except too broad):** hardened — `error_handling.py` now re-raises when the caught exception isn't the expected verifier diagnostic, so unrelated failures don't masquerade as success.

PR #1608, opened 2026-05-30
fix(codegen): carry full-column box for no-split dual-AIV tpush transport

lwDavid:
## Summary

On 910B the **no-split dual-AIV dispatch** runs a mixed root's producer on two AIV subblocks that share one FIFO slot while the single cube consumer pops the full slot. `EmitSplitTpushTransportValidShape` early-returned for `split == 0`, so a producer that narrowed its `valid_shape` (e.g. `set_validshape` on a partial attention block) left the slot **columns past `valid_col` stale**, feeding garbage into the consumer matmul. This surfaced as a fused-attention numeric corruption on Qwen3-14B decode (`fa_fused`).

This extends **#1454**'s full-box transport to the `split == 0` dual-AIV case (detected via the new `PTOCodegen::IsDualAivDispatchFunction`), but widens **columns only while preserving the producer's row `valid_shape`**:

- subblock 0's real push carries the full column box (fixing the stale-col feed into the consumer matmul);
- subblock 1's `valid_shape=(0, 0)` replay stays a **true 0-row no-op** — widening its rows to the full box would race garbage rows into subblock 0's shared slot.

Genuine `split == 1/2` paths keep widening both axes (the row split is real there); plain `split == 0` **without** `dual_aiv_dispatch` is unchanged (no transport emitted), so non-dual-AIV cross-core pipes are byte-identical.

## Validation

- **New UT** `test_no_split_dual_aiv_tpush_widens_cols_preserves_rows` (`tests/ut/codegen/test_pto_codegen_cross_core.py`): asserts the transport `set_validshape` widens cols (`%c16_index`) but **preserves rows** (`%arg1`), and that the logical -> transport -> restore triple is emitted. Without this fix `split == 0` emits no transport (a single `set_validshape` line) and the test fails.
- **Regression**: `test_pto_codegen_cross_core.py` + `test_split_vector_kernel.py` + `test_pto_codegen.py` -- 126 passed locally.
- **Lint**: clang-format, ruff check/format, cpplint (<=110), check-english-only, check-headers, markdownlint-cli2 (project config) all clean.
- **Docs**: the codegen transport contract is documented in `docs/{en,zh-cn}/dev/passes/23-split_vector_kernel.md`.

## Test plan

- [ ] CI green on the current toolchain
- [ ] (follow-up) on-device fused-attention parity for the pipelined dual-AIV case

## Notes

This is the **transport-correctness half** of the no-split dual-AIV bidirectional-pipe fix: it stops subblock 1's replay from clobbering subblock 0's slot and carries the producer's fillpad'd columns. The remaining pipelined-iteration FIFO-slot isolation (two `pl.pipeline` iterations sharing one bidirectional ring) is a separate follow-up.

## Related

Extends #1454 (full-box transport for `split == 1/2`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/en/dev/passes/23-split_vector_kernel.md
+++ b/docs/en/dev/passes/23-split_vector_kernel.md
@@ -214,6 +214,26 @@ both paths).
    stays — downstream lowering reads it).
 ```
 
+### Codegen transport: full-column box, preserved rows
+
+The lane-1 replay zeroes its tile `valid_shape` so it produces no visible
+writes, but the AIC↔AIV `tpush` it keeps still moves data through the shared
+GM FIFO slot the single cube consumer pops in full. On the codegen side
+(`EmitSplitTpushTransportValidShape`, `pto_ops_common.cpp`) a no-split
+dual-AIV producer that narrowed its `valid_shape` with `set_validshape` must
+therefore transport the **full column box**, or the consumer reads stale slot
+columns past `valid_col`. Unlike the genuine `UpDown` / `LeftRight` splits —
+which widen *both* axes — the no-split path widens **columns only and
+preserves the row `valid_shape`**: subblock 0's real push carries the full
+column box, while subblock 1's `valid_shape=[0, 0]` replay gets **no transport
+at all** (a statically 0-row push moves no data, and emitting a col-widening
+`set_validshape` for it perturbs the shared-slot dual-AIV merge — it regressed
+the `cross_core_v2c_nosplit` golden), so it stays a true 0-row no-op rather than
+racing garbage rows into subblock 0's slot. (Plain `split=0` without
+`dual_aiv_dispatch` emits no transport at all either.) The detection lever is
+`PTOCodegen::IsDualAivDispatchFunction()`, which reads this pass's
+`dual_aiv_dispatch` attribute.
+
 ## Constraints
 
 | Constraint | Why |
--- a/docs/zh-cn/dev/passes/23-split_vector_kernel.md
+++ b/docs/zh-cn/dev/passes/23-split_vector_kernel.md
@@ -203,6 +203,24 @@ Ascend910B（或任意
    保留 —— 后续 lowering 会读它）。
 ```
 
+### Codegen 传输：整列 box、保留行
+
+lane 1 的 replay 会把 tile `valid_shape` 清零，从而不产生任何可见写；但它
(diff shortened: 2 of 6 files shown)


lwDavid:
Addressed the review feedback and triaged the red checks:

**Review comments**
- CodeRabbit nitpick (Doxygen on `IsDualAivDispatchFunction` implying both axes widen): fixed in 67ad8051 — the comment now states the split==0 dual-AIV path widens **columns only** and preserves `valid_shape[0]`.

**CI triage — none of the red checks are caused by this change**
- `unit-tests` (ubuntu + macos), `clang-tidy`, `pre-commit`, `system-tests-a5sim`: ✅ green (incl. the new `test_no_split_dual_aiv_tpush_widens_cols_preserves_rows`).
- `dist-system-tests`: failed once on a transient `507018` (`aclrtSynchronizeStreamWithTimeout`) in `test_l3_notify_wait::test_signal_exchange` — a distributed test that doesn't touch this codegen path; **passed on the re-run**.
- `system-tests`: **cancelled** (24h runner timeout), not a real failure — needs a re-run.
- `pypto-lib-model`: **pre-existing flaky golden checks**, not a regression here. It is red on `main` too (fails at `qwen3/14b/decode_layer.py`). Across the two runs of this PR the failure point *moved* with no behavior change (run 1: `qwen3` PASS → `decode_attention_swa` FAIL; run 2 after a comment-only commit: `qwen3` FAIL) — i.e. the goldens are non-deterministic. Root cause is the no-split dual-AIV bidirectional-pipe FIFO-slot race in the fused attention roots (`fa_fused`, `sparse_attn`), where two `pl.pipeline` iterations share one ring.

  This PR is the **transport-correctness half** and is a functional **no-op for `sparse_attn`/qk_pv**: in the generated `qk_pv` AIV code, subblock 0's V2C push emits **no** transport (its exp tile is full-valid after `fillpad`), and subblock 1's replay stays a **0-row** push (`SetValidShape(0, col)`) — identical behavior to `main`. It does help `fa_fused` (garbage → coherent), but the residual race remains.

  The deterministic fix is the per-pipeline-iteration FIFO-slot isolation follow-up noted in the PR description (give each `pl.pipeline` clone its own `initialize_pipe` id + GM ring), which I'm implementing next.

Re-running `system-tests` and `pypto-lib-model`.

lyfne123 (approved):

Issue #1592, opened 2026-05-29
[Bug] Dynamic-address GM scatter on the vec side of a cube+vec mix CORE_GROUP scope deadlocks (507018); dynamic reads + contiguous writes are fine


wangqin1723-max:
### Summary

A `CORE_GROUP` mix scope (`pl.split(...)`) that fuses a cube op (matmul) with a **per-row dynamic-address GM scatter** on the vec side **compiles clean but deadlocks at runtime with `507018`** (`aclrtSynchronizeStreamWithTimeout (AICPU) failed: 507018`). Reproduces under both `SplitMode.NONE` and `SplitMode.UP_DOWN`.

A **dynamic-address READ** and a **contiguous-address WRITE** in the same mix scope are both fine — only the **dynamic-address WRITE** deadlocks (isolation matrix below).

> Note: I have **not** root-caused the mechanism or the owning layer (pypto codegen vs runtime vs pto-isa FFTS). This report is the **reproducible observed behavior + isolation only**; please treat the cause as open.

### Where it shows up

`models/deepseek/v4/decode_indexer_compressor.py`: trying to fuse the `kv_hadamard` (cube matmul) and `kv_and_cache_write` (per-row block-table scatter into `kv_flat` + paged `idx_kv_cache`) scopes into one mix scope.

Pattern (per POST_CHUNK group, NONE mix):

```python
with pl.at(level=pl.Level.CORE_GROUP, optimizations=[pl.split(pl.SplitMode.NONE)], name_hint="kv_hadamard_cache_write"):
    kv_final = ...                      # cube: hadamard matmul -> [POST_CHUNK, HEAD_DIM]
    for r in pl.range(POST_CHUNK):      # vec: per-row scatter
        c_idx = c0 + r
        if gate(c_idx):
            kv_row = kv_final[r:r+1, :]
            kv_flat = pl.assemble(kv_flat, kv_row, [c_idx*S, 0])                 # contiguous store -> OK
            cache_row = block_table_lookup(c_idx)                                # idx_block_table read -> OK
            idx_kv_cache_flat = pl.assemble(idx_kv_cache_flat,                   # DYNAMIC-ADDRESS store -> 507018
                                            pl.cast(kv_row, pl.BF16, mode="rint"),
                                            [cache_row, 0])
```

### Symptom

- Compiles clean (no codegen error).
- Runtime: `507018` on a2a3. Both NONE and UP_DOWN.

[...]

Issue #1564, opened 2026-05-27
[Bug] Codegen: cube matmul -> per-row scatter to distinct GM rows in one CORE_GROUP scope silently produces all-zero output


wangqin1723-max:
### Diagnosis

**pypto** — compiles clean, runs on a2a3, output is all zeros on every written row. A single `pl.matmul` whose FP32 output is materialized into a scope-local `create_tensor` and then **read back per-row and scattered into distinct (strided) GM rows**, all inside one `CORE_GROUP` scope, silently produces zero data on device.

Distinct from the existing reports:
- **#1525** (chained matmul → cast → matmul) explicitly has *no slice* and adds *acc-reuse* between two matmuls. This case has **one matmul**, **no acc-reuse**, and **does** slice (per-row subview) + scatter. Neither contains the other.
- **#1507** (cube→vec bridge drops matmul `valid_shape`, breaking a downstream vec subview) is the closest mechanism — but it is **CLOSED**, and this still reproduces on current `main`, so this is either an uncovered sibling of that fix or a regression on the matmul→store→subview→scatter path.
- Related: #1523, #1352.

### Description

In one `CORE_GROUP` scope (`pl.split(NONE)`):
1. `kv_final[:, :] = pl.matmul(x, w, out_dtype=pl.FP32)` — cube matmul, whole-tile store into a scope-local `create_tensor`.
2. `for b in pl.range(B): cache[b*CACHE : b*CACHE+1, :] = pl.cast(kv_final[b:b+1, :], BF16)` — per-row subview read + scatter to a distinct strided GM row.

Compiles clean, runs (`compile done`, `runtime done`), but the output `cache` is all-zero on every written row (`actual=0.0`, expected nonzero). The cube→vector RAW dependency from the whole-tile store to the per-row reads is dropped: the first reads see uninitialized (zero) data. The result is **deterministic** (same rows zero across runs), not a race.

#### ~30-line repro (`-p a2a3`)

```python
import pypto.language as pl

B = 16
H = 64
CACHE = 256


def build_program():
    @pl.program
    class Repro:
        @pl.function(type=pl.FunctionType.Opaque)
        def fused(
            self,
            x: pl.Tensor[[B, H], pl.BF16],
            w: pl.Tensor[[H, H], pl.BF16],
            cache: pl.Out[pl.Tensor[[B * CACHE, H], pl.BF16]],
        ) -> pl.Tensor[[B * CACHE, H], pl.BF16]:
[...]

PR #1627, opened 2026-06-01
test(scatter): reproduce #1592 dynamic-address GM scatter deadlock in cube+vec mix scope

Little-oil:
## Summary

Adds a runtime regression repro for #1592 to `tests/st/runtime/test_fused_matmul_scatter.py`: a cube matmul fused with a **per-row, block-table-indirected (dynamic-address) GM scatter** in one `CORE_GROUP` mix scope, in both `SplitMode.NONE` and `SplitMode.UP_DOWN`.

Per the issue's isolation matrix, this dynamic-address WRITE on the vec side of a cube+vec mix scope compiles clean but deadlocks at runtime (`507018`), while the static-address WRITE (#1564, already in this file) runs. This PR adds the repro only — **no fix yet** — to confirm the behavior on a2a3 CI before committing to a fix.

New programs/cases (golden identical to the #1564 static variants — only the addressing mode differs):
- `FusedMatmulDynScatterProgram` / `…NoSplitProgram` — scatter row read from a block table at runtime (`cache_row = pl.read(block_table, [b]) * CACHE`; the `* CACHE` keeps the offset `index`-typed so codegen is clean, matching the issue).
- `TestFusedMatmulDynScatter` / `TestFusedMatmulDynScatterNoSplit`.

## Why a repro-only PR

I grounded the diagnosis offline (generated the post-PTOAS `aiv.cpp` via `--codegen-only --save-kernels`). Findings:
- Both AIV lanes (`subblockid==0` real vs the `else` replay) are **independently pipe-flag balanced** — matching the reporter's note that the outer `set_flag`/`wait_flag` frame is balanced. This **disproves the lane-asymmetry theory**.
- Diffing the generated AIV against the working #1564 contiguous kernel, the only op-level difference is one added block-table scalar load plus the dynamic store base address.

So it is possible this minimal repro does **not** deadlock. CI on a2a3 is the decisive signal:
- If it deadlocks (`507018`) → repro confirmed; proceed with the fix on this branch.
- If it passes → enrich the repro toward the real `models/deepseek/v4/decode_indexer_compressor.py` shape (the `gate(c_idx)` guard + block-table-via-tile-load path).

## Testing

- [x] Compiles clean through PTOAS offline (`--codegen-only`) for both NONE and UP_DOWN
- [ ] a2a3 runtime result (this PR's purpose — awaiting CI)

## Related

#1592

### DIFF
--- a/tests/st/runtime/test_fused_matmul_scatter.py
+++ b/tests/st/runtime/test_fused_matmul_scatter.py
@@ -65,6 +65,18 @@ def _make_kv_final() -> torch.Tensor:
     return rows.expand(B, H).contiguous()
 
 
+def _make_block_table() -> torch.Tensor:
+    """``[B]`` INT32 block table: ``block_table[b] = b`` (block id for row ``b``).
+
+    The kernel reads this block id at runtime and computes the scatter row as
+    ``block_id * CACHE`` (the paged ``block_id * block_size`` idiom), so the
+    store lowers to a *dynamic-address* ``pto.tstore``. The ``* CACHE`` keeps the
+    offset ``index``-typed, so codegen is clean — matching the issue (compiles
+    clean, deadlocks at runtime). The golden output is identical to the
+    static-address variants; only the addressing mode (and codegen path) differ."""
+    return torch.arange(B, dtype=torch.int32).contiguous()
+
+
 @pl.program
 class FusedMatmulScatterProgram:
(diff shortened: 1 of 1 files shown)


Little-oil:
Cleanup: 关闭 ≤2026-06-11 的陈旧 PR（分支一并清理）。如后续仍需，可从分支重开。

PR #1645, opened 2026-06-02
feat(jit): add @pl.jit.host for HOST-level orchestration

### LINKED ISSUE #1637: [Feature] Host-orchestrator JIT signature to author distributed programs without @pl.program
### Summary

Add a JIT decorator/signature variant for **HOST-level orchestration** so distributed programs can be authored entirely with `@pl.jit` functions instead of a `@pl.program` class. Today the host orchestrator — the function that allocates HCCL windows (`pld.alloc_window_buffer` / `pld.window`), loops `for r in pl.range(pld.world_size())`, and dispatches each rank via `chip_orch(..., device=r)` — can only be a `@pl.function(level=pl.Level.HOST, role=pl.Role.Orchestrator)` method inside a `@pl.program` class. There is no jit equivalent.

### Motivation / Use Case

We're aligning the 2-rank EP MoE (`models/deepseek/v4/moe_ep.py`) with the clean single-card jit composition style (`moe.py`: `@pl.jit.inline` kernels composed by a top-level `@pl.jit`). With #1635 (DistributedTensor params in jit), the **chip-level orchestration** (`chip_orch`) can be expressed as a jit function. But the **host layer** still forces `@pl.program`, so the program can't fully drop the class form. A host-orchestrator jit variant would let the entire distributed program — host loop + chip orchestration + InCore kernels — be authored as jit functions, giving one consistent authoring style for single-card and EP.

### Proposed API / Behavior

A distinct decorator (not the default `@pl.jit` / `@pl.jit.inline`) that carries HOST + Orchestrator semantics and permits window allocation + the `world_size` rank loop + per-rank `device=` dispatch, e.g.:

```python
@pl.jit.host          # or @pl.jit(level=pl.Level.HOST, role=pl.Role.Orchestrator)
def host_orch(... per-rank inputs/outputs ..., layer_id):
    pub_counts_buf = pld.alloc_window_buffer(N_RANKS * N_RANKS * N_LOCAL * 4)
    ...
    for r in pl.range(pld.world_size()):
        pub_counts = pld.window(pub_counts_buf, [...], dtype=pl.INT32)
        ...
        chip_orch(..., pub_counts, ..., layer_id, r, device=r)   # chip_orch is a @pl.jit orchestration fn
```

Key requirements the new signature must support that default jit does not: (1) `pld.alloc_window_buffer` / `pld.window`; (2) `pld.world_size()` and the per-rank loop; (3) the `device=` dispatch kwarg targeting a jit chip-orchestration function.

### Alternatives Considered

Keep `host_orch` (and the enclosing `@pl.program` class) as the only non-jit shell while everything below it is jit — i.e. a thin `@pl.program` that just hosts `host_orch` + bare-calls jit `chip_orch`. This is the current fallback and works, but it means the program can never be fully `@pl.program`-free, so two authoring paradigms persist.

### Additional Context

- Depends on / pairs with **#1635** (DistributedTensor params in jit) — that unblocks `chip_orch`; this unblocks `host_orch`.
- Builds on **#1127** (L3 Distributed Programming Interface RFC).
- Current reference shape: `host_orch` in `models/deepseek/v4/moe_ep.py` (`@pl.function(level=pl.Level.HOST, role=pl.Role.Orchestrator)`).
- pypto commit: `e851ba54` (main).

lyfne123:
## Summary

Adds `@pl.jit.host`, a JIT decorator variant for HOST-level orchestrators, so distributed programs can be authored entirely with `@pl.jit` functions instead of falling back to a `@pl.program` class. Pairs with #1638 (DistributedTensor params in jit), which unblocked chip-level jit orchestration — this is the host-side counterpart.

- New `@pl.jit.host` sub-decorator (and `@pl.jit.host()` parens form), reusing the existing `_SubFunctionDecorator` machinery; `level=` is rejected (HOST is implicit).
- Specializer's `_build_decorator` emits `@pl.function(level=pl.Level.HOST, role=pl.Role.Orchestrator)` for `func_type == "host"` — the canonical spelling already accepted by the parser and used by `tests/st/distributed/test_l3_allreduce.py`.
- `_discover_deps` is gated by the caller's `_func_type`: a host entry additionally admits `orchestration` deps (its chip orchestrators), while plain `@pl.jit` entries still ignore other `@pl.jit` entries so two top-level kernels never silently fold into one program.
- The caller's `_func_type` is threaded through `_scan_dep_io` → `_extract_local_tensor_metas` → `_resolve_dep_call_metadata` → `_build_contexts` (via a new `func_type_by_id` map) so meta propagation through `v = chip_orch(args)` honours the same gating.

```python
@pl.jit
def chip_orch(a: pl.Tensor, c: pl.Out[pl.Tensor]):
    return c

@pl.jit.host
def host_orch(inp: pl.Tensor, out: pl.Out[pl.Tensor]):
    for r in pl.range(pld.world_size()):
        chip_orch(inp, out, device=r)
    return out
```

## Scope

Decorator + specializer + dep discovery. Runtime distributed dispatch (plumbing `distributed_config` through `JITFunction._compile` so a host-orchestrator jit kernel can be executed end-to-end) is **out of scope** here and tracked as follow-up. The parser, IR pipeline, and `ir.compile()` already accept the generated source; the JIT runtime side just needs a way for the user to specify `distributed_config` (likely through `RunConfig`).

## Test plan

- [x] `tests/ut/jit/test_decorator.py` — `TestJitHostDecoration` (5 cases: bare/parens form, `level=` rejection, name preservation, `pl.jit.host` access) and `TestHostDiscoversOrchestrationDep` (3 cases: host pulls orchestration dep, plain `@pl.jit` does not, host still discovers incore deps).
- [x] `tests/ut/jit/test_specializer.py` — `test_jit_host_emits_host_orchestrator_decorator`, `test_jit_host_with_orchestration_dep` (decorator emission + multi-function rewrite with `device=r` preserved).
- [x] Full `tests/ut/jit/` suite — 210 passed.
- [x] Full `tests/ut/ir/` regression — 3696 passed, 1 skipped.
- [x] `ruff check` / `ruff format` / `pyright` (pre-commit hooks all green).

## Related Issues

Fixes #1637

### DIFF
--- a/docs/en/user/01-language_guide.md
+++ b/docs/en/user/01-language_guide.md
@@ -397,6 +397,59 @@ class MyProgram:
 - Cross-function calls use `self.method_name(...)`
 - The decorated class becomes an `ir.Program`, not a Python class
 
+### `@pl.jit` family
+
+`@pl.jit` decorators let you author kernels as plain Python functions that
+are specialized into `@pl.program` source on first call (no class boundary
+required). Five variants — one per IR function kind — let a single program
+span host, chip, and core levels:
+
+| Decorator | IR target | Use for |
+| --------- | --------- | ------- |
+| `@pl.jit` | `FunctionType.Orchestration` | Chip-level entry point — top-level kernel that dispatches InCore work |
+| `@pl.jit.host` | `level=HOST, role=Orchestrator` | HOST-level entry — allocates window buffers and dispatches chip orchestrators per rank in distributed (L3+) programs |
+| `@pl.jit.incore` | `FunctionType.InCore` | Separate InCore sub-function (accepts `level=` to target a specific hierarchy level) |
+| `@pl.jit.inline` | `FunctionType.Inline` | Helper spliced at every call site by the `InlineFunctions` pass |
+| `@pl.jit.opaque` | `FunctionType.Opaque` | Separate IR function that may wrap orchestration loops and `pl.at` scopes |
+
+Sub-function deps (`.incore` / `.inline` / `.opaque`) are auto-discovered
+from the entry's body; the user just calls them by name. A `@pl.jit.host`
+entry additionally discovers `@pl.jit` (chip orchestration) deps, so a full
+distributed program can be authored without a single `@pl.program` class:
+
+```python
+import pypto.language as pl
+import pypto.language.distributed as pld
+
+@pl.jit.inline
+def reduce_step(local, peer, out): ...
+
+@pl.jit
+def chip_orch(
+    inp: pl.Tensor, out: pl.Out[pl.Tensor],
+    data: pl.InOut[pld.DistributedTensor], peer: pl.Scalar[pl.INT32],
+):
+    return reduce_step(inp, peer, out)   # auto-discovered sub-function
+
+@pl.jit.host
+def host_orch(
+    inputs: pl.Tensor[[2, 1, 256], pl.FP32],
+    outputs: pl.Out[pl.Tensor[[2, 1, 256], pl.FP32]],
+):
+    data_buf = pld.alloc_window_buffer(256 * 4)
+    for r in pl.range(pld.world_size()):
+        data = pld.window(data_buf, [1, 256], dtype=pl.FP32)
+        chip_orch(inputs[r], outputs[r], data, (r + 1) % pld.world_size(),
+                  device=r)            # device= dispatches per-rank
+    return outputs
+```
+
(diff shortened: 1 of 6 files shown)


Hzfengsy (approved):
LGTM — clean addition of @pl.jit.host with func_type-gated dep discovery, well-tested (decorator + specializer + regression suites green). Runtime dispatch correctly scoped out as follow-up.

PR #1658, opened 2026-06-03
fix(ir): guard ConvertToSSA scope boundaries against bogus escaping (#1351)

### LINKED ISSUE #1351: [Pass Bug] @pl.jit SSA converter incorrectly promotes inner-loop variables to outer-loop init_values
### Pass Name

ConvertToSSA

### Description

The `ConvertToSSA` pass traverses the entire function body including inside `pl.at` blocks to find variable assignments. When it encounters a variable assigned inside `pl.range` or `pl.parallel` body, it unconditionally adds it to the loop's `init_values` tuple. This is incorrect for variables that are computed fresh each iteration (like `idx = loop_var * STRIDE`) — only true accumulators (variables whose previous-iteration value is consumed) should be loop-carried.

### Git Commit ID

92c12c1e

### Before IR (Input)

```python
for ob__idx_v0, (result__iter, k0__iter, oc0__iter) in pl.parallel(N, init_values=(result__ssa, k0, oc0)):
```

```python
Variables computed fresh each iteration (e.g., `idx = loop_var * STRIDE`) should NOT be added to `init_values`. Only true accumulators should be loop-carried. The SSA converter should respect `pl.at` block boundaries and not promote variables defined inside `pl.at` to outer loop init_values.
```

### Actual IR or Error

```shell
Variable 'k0' used outside its defining scope
```

### NPU Kind (if relevant)

None

None

### Coding workarounds

1. **Move `pl.at` inside loops**: `pl.parallel` → `pl.at` → `pl.range` (not `pl.at` → `pl.parallel` → `pl.range`)
2. **Merge `pl.parallel` loops** over the same iteration space into one
3. **Extract iteration-level indices** between `pl.parallel` and `pl.at`: `oc0 = ob * OUT_CHUNK`
4. **Inline index expressions inside `pl.range` body**: use `kb * K_CHUNK` directly, not `k0 = kb * K_CHUNK`
5. **Avoid reusing variable names** across different `pl.range` loops in the same scope — rename or inline
6. **Use `pl.Out`** (not `pl.InOut`) for state tensors in `@pl.jit` functions — `pl.InOut` is not supported

lyfne123:
## Summary

Fixes #1351.

`ConvertScope` inherited `future_needs_` unchanged into the body of every `ScopeStmt` subclass. When a variable was first defined inside `pl.at` (`InCoreScopeStmt` / `HierarchyScopeStmt` / `SpmdScopeStmt` / `ClusterScopeStmt` / `AutoInCoreScopeStmt`) and later referenced after the scope, that downstream reference put the variable into `future_needs_`. Nested loops inside the scope body then saw it in their escaping-variable detection and tried to promote it to `init_values`; `FindInitValue` had no matching pre-loop scalar so it manufactured an unversioned placeholder, producing IR like:

```python
for ob__idx_v0, (k0__iter_v0,) in pl.parallel(N, init_values=(k0__FREE_VAR,)):
    for kb__idx_v0, (k0__iter_v2,) in pl.range(K, init_values=(k0__iter_v0,)):
        k0__ssa_v4 = kb__idx_v0 * K_CHUNK__ssa_v0
        ...
        k0__rv_v3 = pl.yield_(k0__ssa_v4)
    k0__rv_v1 = pl.yield_(k0__rv_v3)
out_scalar__ssa_v1 = k0__rv_v1  # cross-scope reference, SSAVerify rejects it
```

## Fix

- `ConvertScope` now trims `future_needs_` to variables already in `cur_` at scope entry, so the inner-loop escaping path in `ConvertFor` / `ConvertWhile` ignores scope-local newly-defined vars.
- `cur_` itself is left transparent because `InterchangeChunkLoops` and similar later passes intentionally emit `out = pl.assemble(...)` inside `pl.at` and use `out` outside it (covered by `test_host_side_assemble_after_parallel_chunk_not_wrapped` and friends).
- `RuntimeScopeStmt` (`pl.scope()`) stays fully transparent — it is a thin codegen wrapper, not a boundary.

## Test plan

- [x] 3 new regression tests in `TestScopeOutlineBoundary`:
  - `test_scope_blocks_escaping_var_promotion_in_inner_loop` — reproduces #1351 and asserts the inner loops have no `iter_args`.
  - `test_scope_preserves_pre_existing_carried_var` — accumulator pattern (outer parameter reassigned inside `pl.at`) still carried.
  - `test_runtime_scope_remains_transparent_to_escaping` — `pl.scope()` keeps existing transparent behavior.
- [x] All 62 tests in `test_convert_to_ssa_pass.py` pass.
- [x] All 1538 tests in `tests/ut/ir/transforms/` pass (verifies no downstream pass regresses on the same IR).
- [x] Broader `tests/ut/` suite: 5645 passed (3 unrelated runtime test failures exist on `main` too).

## Documentation

- `docs/en/dev/passes/04-convert_to_ssa.md` and `docs/zh-cn/dev/passes/04-convert_to_ssa.md`: added an algorithm bullet describing the scope-boundary escaping guard.
- `include/pypto/ir/stmt.h`: updated the `ScopeStmt` doxygen to spell out the new boundary semantics and the `RuntimeScopeStmt` exception.

### DIFF
--- a/docs/en/dev/passes/04-convert_to_ssa.md
+++ b/docs/en/dev/passes/04-convert_to_ssa.md
@@ -43,7 +43,8 @@ program_ssa = ssa_pass(program)
 3. **Iter_args for Loops**: Convert loop-modified variables to iter_args + return_vars pattern with YieldStmt
 4. **Escaping Variable Promotion**: Variables first defined inside a loop body but used after the loop are promoted to iter_args + return_vars via forward-use analysis
 5. **Scope Tracking**: Track variable definitions across nested scopes
-6. **Preservation**: Keep existing SSA constructs unchanged
+6. **Scope-Boundary Escaping Guard**: For all `ScopeStmt` subclasses *except* `RuntimeScopeStmt` (i.e. `HierarchyScopeStmt` / `InCoreScopeStmt` / `AutoInCoreScopeStmt` / `ClusterScopeStmt` / `SpmdScopeStmt`), `ConvertScope` trims the `future_needs_` set to variables already defined before the scope. This prevents nested loops inside the scope body from promoting *scope-local* variables to bogus `init_values=(foo__FREE_VAR,)` based on a downstream use that lives outside the scope. Variables first-defined inside the scope body still substitute normally outside it (later passes such as `InterchangeChunkLoops` rely on this). `RuntimeScopeStmt` (`pl.scope()`) is a thin codegen wrapper and stays fully transparent.
+7. **Preservation**: Keep existing SSA constructs unchanged
 
 **Key transformations**:
 
--- a/docs/zh-cn/dev/passes/04-convert_to_ssa.md
+++ b/docs/zh-cn/dev/passes/04-convert_to_ssa.md
@@ -43,7 +43,8 @@ program_ssa = ssa_pass(program)
 3. **循环的 Iter_args**：将循环中修改的变量转换为 iter_args + return_vars 模式，带有 YieldStmt
 4. **逃逸变量提升**：通过前向使用分析，将首次在循环体内定义但在循环后使用的变量提升为 iter_args + return_vars
 5. **作用域跟踪**：跨嵌套作用域跟踪变量定义
-6. **保留**：保持现有 SSA 构造不变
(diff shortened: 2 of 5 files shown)


Hzfengsy (approved):
LGTM — guards ConvertToSSA scope boundaries against bogus escaping; tests and docs included.

PR #1682, opened 2026-06-04
fix(ir): preserve Call attrs (dump_vars) through call-site rewrites

luohuan19:
## Summary

Several passes rebuilt `Call` nodes through constructor overloads that
default-initialize `attrs_` to empty, silently dropping `kAttrDumpVars` (set by
`pl.dump_tag` / `dumps=`) during call-site rewrites. As a result, a tagged
tensor could fail to reach the device dump manifest whenever one of these
rewrites fired. This PR makes the affected rewrites copy `attrs_` verbatim,
mirroring the existing `Submit` path.

Fixed sites (all carry `attrs_` through now):

- **`ConvertTensorToTileOps`** — `WrapperForwardMutator` and
  `CallSiteUpdateMutator` (arg-appending rewrite when an InCore callee gains a
  runtime-allocated `Out` param).
- **`ExpandMixedKernel::RewriteGroupCaller`** — the Group→InCore call split into
  AIC/AIV lanes; both lanes now forward `attrs_`, and the AIC lane also forwards
  `kwargs_` (previously built via the bare 3-arg ctor).
- **`OptimizeOrchTensors`** — the iter-arg `Out`→`InOut` merge call-site rewrite.

Reused `args_` are unchanged across each rewrite, so dump/dep `Var` references
stay valid and orchestration codegen still matches them per-arg by identity.

For void-return rewrites, the rebuilt `Call` falls back to `UnknownType` so its
`type_` is identical to the prior constructor path (avoids introducing a null
`type_`).

Safety note: the full `Call` ctor runs `ValidateArgDirectionsAttr()`, which would
reject a verbatim `attrs_` copy if `kAttrArgDirections` were present with a now
mismatched length. This is safe because all three passes run **before**
`DeriveCallDirections` (which seeds that attr), so it is absent at these sites.

## Testing

- [x] `test_dump_vars_attr_preserved_through_call_site_update` (ConvertTensorToTileOps)
- [x] `test_iter_arg_merge_preserves_dump_vars` (OptimizeOrchTensors)
- [x] `test_dump_tag_survives_cluster_group_mixed_split` (ExpandMixedKernel, end-to-end)
- [x] Each regression test verified to fail without its fix and pass with it.
- [x] Full affected suites green (convert / optimize-orch / dump-tag / expand-mixed): 141 passed.
- [x] Clean RelWithDebInfo build; no new clang-tidy findings on changed lines.

### DIFF
--- a/src/ir/transforms/convert_tensor_to_tile_ops_pass.cpp
+++ b/src/ir/transforms/convert_tensor_to_tile_ops_pass.cpp
@@ -1477,9 +1477,14 @@ class WrapperForwardMutator : public TypePropagatingMutator {
       new_return_type = std::make_shared<TupleType>(incore_func->return_types_);
     }
 
-    auto new_call = new_return_type ? std::make_shared<Call>(call->op_, new_args, call->kwargs_,
-                                                             new_return_type, call->span_)
-                                    : std::make_shared<Call>(call->op_, new_args, call->kwargs_, call->span_);
+    // Preserve the original call's attrs_ (e.g. kAttrDumpVars from
+    // pl.dump_tag / dumps=) through the arg-appending rewrite — mirrors the
+    // base IRMutator and the Submit path below. The appended outputs do not
+    // rename existing arg Vars, so a verbatim attr copy keeps any dump/dep Var
+    // references valid. Fall back to UnknownType for a void-return callee so the
+    // rewritten Call's type_ stays identical to the prior 4-arg ctor path.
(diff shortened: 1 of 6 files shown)


luohuan19:
@coderabbitai review

Hzfengsy (approved):
LGTM — preserves Call attrs_ (dump_vars) through call-site rewrites, mirroring the Submit path. Correctly scoped before DeriveCallDirections so the ValidateArgDirectionsAttr concern is moot. Regression tests cover all three fixed sites.

PR #1695, opened 2026-06-08
feat(distributed): runtime-bound SubWorker callbacks via `...` body

lyfne123:
## Summary

Makes a HOST `SubWorker` declared with a `...` body an explicit **runtime-bound callback point**. Its implementation is supplied at `compiled.prepare(callbacks={...})` rather than captured at compile time — for closures that need live runtime state (sampling, host-side inspection) and previously had to be smuggled in as a generated placeholder + silent name-based override.

### Before

```python
@pl.function(level=pl.Level.HOST, role=pl.Role.SubWorker)
def sample(logits: pl.Tensor[[B, V], pl.FP32]) -> pl.Tensor[[B], pl.INT32]:
    return torch.argmax(logits, dim=-1)   # placeholder body, silently replaced at runtime

with compiled.prepare(sub_worker_overrides={"sample": real_sampler}) as rt:  # optional, name-string only
    ...
```

### After

```python
@pl.function(level=pl.Level.HOST, role=pl.Role.SubWorker)
def sample(logits: pl.Tensor[[B, V], pl.FP32]) -> pl.Tensor[[B], pl.INT32]:
    ...   # abstract: implementation bound at runtime

with compiled.prepare(callbacks={"sample": real_sampler}) as rt:  # required; missing → ValueError
    ...
```

## What changed

- **IR**: new `Function.requires_runtime_binding_` field. Reflected, so it round-trips through `.pto` serialization and the Python printer (which re-emits the `...` body). A bare `pass` body stays a concrete no-op SubWorker.
- **Parser**: detects a `...`-only SubWorker body and sets the flag (threaded through the builder → `BeginFunction` → `Function` ctor).
- **Codegen**: emits a guard stub that raises if the callback is dispatched unbound (instead of silently no-op'ing), plus a `sub_workers/__required__.json` manifest listing abstract names.
- **Runtime**: `prepare(callbacks=...)` enforces binding — a missing required callback raises `ValueError` at prepare time, not at dispatch; callback arity is validated. `sub_worker_overrides=` kept as a deprecated alias.
- **Docs**: new "Abstract (runtime-bound) SubWorkers" section in `docs/{en,zh-cn}/dev/ir/01-hierarchy.md`.
- **Example**: `examples/runtime/distributed_callback.py`.

## Testing

- `tests/ut/codegen/test_distributed_codegen.py`: abstract detection, `pass`-is-not-abstract, printer round-trip, `.pto` serialization round-trip, guard-stub emission.
- `tests/ut/runtime/test_distributed_worker.py`: rewrote override/merge coverage onto the `callbacks` API; added missing-required, arity, and deprecated-alias cases.
- Broad `tests/ut` sweep green (pre-existing unrelated `Worker.current` mock failures aside).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/en/dev/ir/01-hierarchy.md
+++ b/docs/en/dev/ir/01-hierarchy.md
@@ -503,6 +503,35 @@ remain `nullopt` unless set by the caller. The Python printer emits the
 `level=` / `role=` keywords on the `@pl.function(...)` decorator whenever they
 are present.
 
+#### Abstract (runtime-bound) SubWorkers
+
+A HOST-level `SubWorker` is a pure-Python callback that runs in the forked
+orchestrator process (its body is captured verbatim as an `InlineStmt`, not
+parsed as DSL). Some callbacks cannot be written at compile time — e.g. a
+sampling closure that needs live model state. Declaring the body as `...`
+marks the SubWorker as an **abstract, runtime-bound callback point**:
+
+```python
+@pl.function(level=pl.Level.HOST, role=pl.Role.SubWorker)
+def sample(logits: pl.Tensor[[B, V], pl.FP32]) -> pl.Tensor[[B], pl.INT32]:
+    ...   # implementation supplied at runtime, NOT here
+```
+
+This sets `Function.requires_runtime_binding_ = true` (a reflected field, so it
+round-trips through `.pto` serialization and the Python printer, which re-emits
+the `...` body). A bare `pass` body is **not** abstract — it is a concrete
+no-op SubWorker.
+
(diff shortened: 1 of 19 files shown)


Hzfengsy (approved):
LGTM. Verified the f-string at pto_backend.py:1198 interpolates func.name correctly (gemini's flag is a false positive). The coderabbitai note on requires_runtime_binding type validation is a minor defensive-hardening nit consistent with sibling boolean fields — non-blocking.

Issue #1693, opened 2026-06-08
[Bug] Orchestration cross-assigns __ssa_v aliases for multiple in-place outputs of an inlined InCore scope


zhangqi-chen:
### Component
Codegen

### Description
When an `@pl.jit.inline` kernel is inlined into a `@pl.jit` caller and writes
several output tensors **in place inside one `pl.at(level=pl.Level.CORE_GROUP)`
(= ScopeStmt InCore) scope via tile `pl.store`**, those tensors are classified
`add_inout`, and the orchestration codegen emits post-write SSA aliases
(`<name>__ssa_v1`) that are **cross-assigned to the wrong source tensors**.
Downstream consumers then read a completely unrelated tensor — a silent
miscompile producing NaN/Inf and a runtime 507018 (AICPU chip fault).

This is the same cross-assignment class as the (closed) #1573, but a distinct
trigger: #1573 is about 3+ outer-loop **carries** (`__phi` nodes) in a
`pl.parallel`/`pl.range` loop; here there is no loop carry — they are plain
in-place outputs of an inlined kernel, and the scrambled aliases are `__ssa_v1`,
not `__phi`. It still reproduces on current `main` (0ac36c0f, which includes
#1573's closure).

### Steps to Reproduce
In pypto-lib, convert `dispatch_ep` in `models/deepseek/v4/dispatch.py` from
`@pl.jit.incore` to `@pl.jit.inline`, writing the recv_*_out outputs in place
inside one `pl.at(CORE_GROUP)` scope, with `recv_scale_out` / `recv_w_out`
written via tile `pl.store` (TROWSUM form). Then:

    python models/deepseek/v4/moe_ep.py -d 0,1   # 2-rank EP, 2 NPUs

`recv_scale_out` and `recv_w_out` are `[N_LOCAL, RECV_MAX]` FP32; both written
via tile `pl.store` of full rows -> classified `add_inout`.

### Expected Behavior
Each in-place output's post-write alias references itself, e.g.:

    recv_scale_out__ssa_v1 = recv_scale_out;
    recv_w_out__ssa_v1     = recv_w_out;

### Actual Behavior
Generated `next_levels/moe_ep/orchestration/moe_ep.cpp`:

    // Task 12: dispatch_ep  (params add_inout recv_scale_out, recv_w_out, ...)
    ...
    const Tensor& recv_w_out__ssa_v1   = recv_count_out;   // WRONG (int32 count)
[...]

Issue #1650, opened 2026-06-02
[Bug] AICPU scheduler deadlock (507018) when one fanned-out spmd scope both writes and reads the same GM scratch (intra-scope self-RAW)


wangqin1723-max:
### Diagnosis

**pypto runtime / simpler scheduler** (intra-scope dependency scheduling). A single fanned-out `pl.spmd` scope that **both writes and reads the same GM scratch tensor** compiles to a legal, acyclic orchestrator program but **deadlocks at runtime** with `507018`. The emitted orch DAG is acyclic and GM dependencies are tracked at **whole-tensor** granularity, so the read-side fan-out units depend on *all* write-side units; under limited core co-residency the units wait on each other and the AICPU scheduler deadlocks. pypto is only an enabler (it permits such a scope); the failure is a runtime scheduling deadlock, not a codegen error, not ptoas/pto-isa, not hardware.

### Description

Within one `pl.spmd` scope, storing to a GM scratch and later reading it back (here via tensor-subscript + `pl.transpose`) deadlocks:

```
aclrtSynchronizeStreamWithTimeout (AICPU) failed: 507018
PTO2 runtime failed: orch_error_code=0 sched_error_code=100 runtime_status=-100
RuntimeError: run_prepared failed with code 507018
```

The deadlocking kernel is `models/deepseek/v4/hc_pre.py` fused into 3 spmd scopes; the offending **scope 2** does (per 16-row tile):

```python
mixes_gm   = pl.create_tensor([T, MIX_PAD], pl.FP32)   # written by scope 1 (matmul)
pre_val_gm = pl.create_tensor([T, HC_PAD],  pl.FP32)   # scratch

for ob in pl.spmd(T // TILE, name_hint="hc_post"):     # scope 2
    t0 = ob * TILE
    # pre: read mixes_gm, compute, WRITE pre_val_gm
    pre_in = pl.load(mixes_gm, [t0, 0], [TILE, HC_PAD], target_memory=pl.MemorySpace.Vec)
    ...
    pl.store(pl.add(pre_sig, HC_EPS), [t0, 0], pre_val_gm)   # write scratch
    # post: read mixes_gm, write `post`
    ...
    # mix_x: READ the just-written pre_val_gm (same task), transpose, write x_mixed
    pre_tile   = pre_val_gm[t0:t0 + TILE, 0:HC_PAD]          # <-- self read-after-write
    pre_tile_t = pl.transpose(pre_tile, axis1=0, axis2=1)
[...]

Issue #1573, opened 2026-05-28
[Bug] Orchestration codegen scrambles phi assignments when one task writes 3+ outer-loop carries


zhangqi-chen:
### Component

Codegen

### Description

When a single InCore `with pl.at(...)` scope inside a `for x in pl.parallel(B):` (or `pl.range`) outer loop writes to **three or more outer-scope tensors that are also carried across loop iterations**, the orchestration cpp codegen emits **scrambled phi-node assignments** after the task call. Each phi gets assigned the wrong source tensor — e.g. `cmp_kv_pool_flat__phi = normed_kv` even though they have different shapes/dtypes.

Downstream behavior:
- If the wrongly-aliased tensor is later passed to a `tensor.reshape`, AICPU hits `Assertion failed: valid_reshape(new_shapes, new_ndims)` at `runtime/tensor.h:435` and the device stalls on never-resolved task deps.
- If the bad alias is only consumed by another InCore task, the scheduler reports `TIMEOUT_EXIT after_idle_iterations=800000` — all cores idle waiting on a dependency that can never be satisfied.

### Steps to Reproduce

Trigger in pypto-lib branch `perf-cmp128`, by merging `models/deepseek/v4/decode_compressor_ratio128.py` `kv_write_and_scatter_next` scope into `rmsnorm_rope`:

```python
for global_c_idx in pl.parallel(B):
    ...
    with pl.at(level=pl.Level.CORE_GROUP, name_hint="rmsnorm_rope"):
        # ... RMSNorm + RoPE writes to normed_kv ...
        # MERGED tail (was its own pl.at scope):
        if pos_b + S >= COMPRESS_RATIO:
            kv_flat[global_c_idx*S:global_c_idx*S+1, :] = ...      # carry #1
            cmp_kv_pool_flat[phys_cmp_row:phys_cmp_row+1, :] = ... # carry #2
            for s in pl.range(pre_tokens_b, S):
                kv_state_pool_flat[state_blk_id:state_blk_id+1, ...] = ...  # carry #3
```

Then run `python models/deepseek/v4/decode_compressor_ratio128.py -p a2a3 --hetero-start-pos`.

### Expected Behavior

After the task call, each phi node receives the corresponding output of the task. With outputs `[kv_flat, cmp_kv_pool_flat, kv_state_pool_flat]`, the phis should be:

```cpp
kv_flat__phi          = kv_flat__rv;
[...]

PR #1718, opened 2026-06-09
fix(codegen): alias multi-output spmd-scope return to the correct output (#1702)

### LINKED ISSUE #1702: [Bug] Orchestration codegen aliases a multi-output spmd scope's return value to the wrong output tensor (out_indices[0] fallback) -> numel-mismatched reshape -> 507018
### Component

Codegen

### Description

When a fanned-out `pl.spmd` scope writes **multiple** GM output tensors and the
function's return value is one of the **non-first** outputs, the orchestration
codegen aliases the scope's returned SSA value to the **wrong** output tensor.
Downstream the wrong tensor is reshaped to the return shape; because the two
tensors have different element counts, the AICPU runtime hits
`debug_assert(valid_reshape(...))` (`runtime/.../runtime/tensor.h:435`,
`valid_reshape() = numel()==product(new_shapes)`). The asserting executor thread
aborts that subtask, it never writes FIN, the producer task never reaches
COMPLETED, its consumers' fan-in never closes, the scheduler latches the
GLOBAL_STUCK timeout (`sched_error_code=100`), and the host surfaces it as
`aclrtSynchronizeStreamWithTimeout (AICPU) failed: 507018`. A kernel-side shape
bug thus masquerades as a scheduler deadlock.

**Root cause (pinned):** `GenerateSingleReturnAlias`
(`src/codegen/orchestration/orchestration_codegen.cpp:2446`):

```cpp
auto returned_idx = FindReturnedParamIndex(callee, program_);
size_t param_idx = returned_idx.value_or(out_indices[0]);  // <-- falls back to FIRST output
EmitTensorAlias(var_name, call, param_idx);
```

For a multi-output `pl.spmd` scope, `FindReturnedParamIndex` returns `nullopt`
(a Spmd wrapper has no traceable top-level `ReturnStmt`), so `param_idx` falls
back to `out_indices[0]` = the **first** Out/InOut arg — but the actually
returned value is a **later** output. The returned SSA value should instead be
aliased to the output arg that shares its SSA base / buffer root.

This is the same **class** as #1573 (closed) and #1693, but a distinct trigger
and code path: #1573 is `__phi` loop carries; #1693 is `__ssa_v1` in-place
outputs of an inlined `pl.at(CORE_GROUP)` InCore scope; **this** is the
`__rv_v` *return-value* alias of a `pl.spmd` scope via the
`GenerateSingleReturnAlias` `out_indices[0]` fallback.

Related: #1693, #1573, #1650, #631

### Steps to Reproduce

In pypto-lib, take the DeepSeek-V4 `hc_pre` kernel fused into **3** `pl.spmd`
scopes where the **second** scope (`hc_post`) writes three GM outputs in one
fanned-out spmd task — `pre_val_gm` (inout), `post` (inout), and `x_mixed`
(output, the function's return) — e.g. `hc_pre_fused_draft.py` with `mix_x`
fused into `hc_post`. The function ends with
`x_mixed = pl.reshape(x_mixed_view, [B, S, D]); return x_mixed`.

```
python models/deepseek/v4/repro_hc_pre_blocker2_selfraw_507018.py -p a2a3
```

(Real `FLASH` config: T=128, D=4096, HC_DIM=16384, MIX_HC=24. Run via
`task-submit` on an a2a3 device; the kernel compiles cleanly and deadlocks at
runtime with 507018.)

The 4-spmd variant (mix_x as its **own** scope, so `x_mixed` is that scope's
sole output) passes — confirming the trigger is the multi-output scope, not the
data flow.

### Expected Behavior

The scope's returned SSA value aliases to the output arg it is actually a new
SSA version of:

```cpp
const Tensor& x_mixed_view_inline52__rv_v2 = x_mixed_view_inline52;   // the x_mixed output
```

### Actual Behavior

Generated `orchestration/hc_pre_test.cpp` (multi-output `hc_post` Submit):

```cpp
// hc_post Submit outputs:
params_t1.add_inout(pre_val_gm_inline83);      // [2]  shape [128, 8]   numel = 1024
params_t1.add_inout(post_2d_inline73);         // [3]
params_t1.add_output(x_mixed_view_inline52);   // [4]  shape [128, 4096]  (== x_mixed)
rt_submit_aiv_task(2, params_t1);

const Tensor& x_mixed_view_inline52__rv_v2 = pre_val_gm_inline83;     // WRONG: aliased to output [2]

uint32_t x_mixed_v1_inline4_shapes[3] = {64, 2, 4096};               // numel = 524288
Tensor x_mixed_v1_inline4 = x_mixed_view_inline52__rv_v2.reshape(x_mixed_v1_inline4_shapes, 3);
```

`pre_val_gm` is `[128, 8]` (numel 1024); reshaping it to `[64, 2, 4096]`
(numel 524288) fails `valid_reshape` → AICPU subtask abort → scheduler
GLOBAL_STUCK → host `507018` (`sched_error_code=100 runtime_status=-100`).

Device log (a2a3, two independent task-submit-locked runs):

```
[ERROR] AICPU ... assert_impl "Assertion failed: valid_reshape(new_shapes, new_ndims)"
[ERROR] AICPU ... "Location: runtime/.../tensormap_and_ringbuffer/runtime/tensor.h:435"
        Stack: AicpuExecutor::run -> aicpu_execute -> simpler_aicpu_exec
```

### Git Commit ID

3f421313 (pypto, branch main)

### NPU Kind

Ascend 910B

### Host Platform

Linux (aarch64)

### Additional Context

Fix direction: in `GenerateSingleReturnAlias` (and the analogous tuple/Submit
paths), map the returned SSA result to the output arg sharing its SSA
base/buffer root, instead of the `out_indices[0]` fallback when
`FindReturnedParamIndex` is empty for a Group/Spmd wrapper. See
`.claude/rules/pass-submit-awareness.md` (Submit tuple-return aliasing).

Two secondary observations surfaced while diagnosing this:

1. **Diagnosability:** an AICPU-side `debug_assert(valid_reshape)` aborts the
   subtask silently and the failure only appears as a scheduler 507018/507046
   stream-sync timeout. Printing `numel()` / `new_shapes` / `new_ndims` /
   `kernel_id` at `tensor.h:435` (and latching a distinct error code instead of
   GLOBAL_STUCK when an executor subtask aborts) would point straight at the
   kernel instead of looking like a scheduler deadlock.

2. **Latent (separate):** `next_block_idx` is a non-atomic `int16_t` mutated by a
   concurrent read-modify-write across scheduler threads
   (`scheduler_dispatch.cpp:311-318`) — a real data race at `aicpu_thread_num>1`,
   though it is NOT this failure (the dispatch trace shows all blocks dispatched).

wangqin1723-max:
## Summary

Fixes #1702. Orchestration codegen aliased a multi-output `pl.spmd` scope's return value to the **wrong** output tensor, producing a numel-mismatched `Tensor::reshape` that aborts the AICPU subtask and surfaces as a 507018 scheduler deadlock.

## Root cause

`GenerateSingleReturnAlias` (`src/codegen/orchestration/orchestration_codegen.cpp`) resolved a scope's single SSA return via:

```cpp
auto returned_idx = FindReturnedParamIndex(callee, program_);
size_t param_idx = returned_idx.value_or(out_indices[0]);   // <-- blind first-output fallback
```

A fanned-out `pl.spmd` scope is outlined into a Spmd wrapper whose body ends in the inner kernel call, so it has **no traceable top-level `ReturnStmt`** → `FindReturnedParamIndex` returns `nullopt` (or mis-traces through the first Out arg). The `out_indices[0]` fallback then aliases the result to the **first** output.

When such a scope writes multiple GM outputs and the function returns a **non-first** output (DeepSeek-V4 `hc_pre` with `mix_x` fused into `hc_post`), the generated orchestration was:

```cpp
params_t1.add_inout(pre_val_gm_inline83);      // [128, 8]    numel 1024  (out_indices[0])
params_t1.add_output(x_mixed_view_inline52);   // [128, 4096]  == the returned x_mixed
rt_submit_aiv_task(2, params_t1);
const Tensor& x_mixed_view_inline52__rv_v2 = pre_val_gm_inline83;          // WRONG output
Tensor x_mixed_v1 = x_mixed_view_inline52__rv_v2.reshape({64,2,4096}, 3);  // 1024 != 524288
```

`reshape([128,8] -> [64,2,4096])` fails `valid_reshape` (`runtime/.../tensor.h:435`) on the AICPU executor → the subtask aborts and never FINs → the scope never COMPLETEs → scheduler latches GLOBAL_STUCK (`sched_error_code=100`) → host `aclrtSynchronizeStreamWithTimeout (AICPU) failed: 507018`. A kernel-side shape bug masquerading as a scheduler deadlock.

## Fix

The returned SSA result var is an SSA rename of the **output arg** carrying the result, so its SSA base / buffer root is the authoritative signal for which Out/InOut arg the scope returns. Match the result var's SSA base against each Out/InOut arg; when it uniquely identifies one, use it over the unreliable `FindReturnedParamIndex`. Keep the `out_indices[0]` fallback only when there is no unique match (genuine single-Out kernels), plus an `INTERNAL_CHECK_SPAN` bound check mirroring `GenerateTupleReturnAliases`.

Same family as #1573 / #1693 (orchestration mis-aliasing multi-output scopes) but a distinct trigger and code path; the tuple / `__ssa_v1` / `__phi` paths are untouched.

## Validation

- On-board (Ascend a2a3): the DeepSeek-V4 `hc_pre` kernel fused with `mix_x` in the same multi-output spmd scope **no longer 507018-deadlocks** — it compiles and runs into runtime with this fix (without it, the `valid_reshape` abort hangs the AICPU).
- Local: builds clean (`pip install -e .`), no regression in `tests/ut/ir/verifier/` and `tests/ut/codegen/test_spmd_scope_tid_codegen.py`.

## Note on testing

A standalone `@pl.program` UT does not reproduce the exact mis-alias — it only triggers through the `pl.jit.inline` + tail-appended `create_tensor`-scratch path this kernel exercises (a minimal scope resolves the return to its param directly, bypassing `GenerateSingleReturnAlias`). On-board is the decisive validation here; a faithful regression/ST test is tracked as a follow-up.

Related: #1702, #1693, #1573, #1650

### DIFF
--- a/src/codegen/orchestration/orchestration_analysis.cpp
+++ b/src/codegen/orchestration/orchestration_analysis.cpp
@@ -297,11 +297,46 @@ void VarLineageCollector::VisitStmt_(const AssignStmtPtr& assign) {
         // returns (multi-Out kernels would otherwise be mis-traced to the
         // first Out, leaking scratch-buffer lineage onto the result Var).
         std::optional<size_t> returned_idx = FindReturnedParamIndex(callee, program_);
+
+        // Collect the Out/InOut arg positions once.
+        std::vector<size_t> out_positions;
         for (size_t i = 0; i < effective_dirs.size() && i < call->args_.size(); ++i) {
-          if (effective_dirs[i] != ParamDirection::Out && effective_dirs[i] != ParamDirection::InOut) {
-            continue;
+          if (effective_dirs[i] == ParamDirection::Out || effective_dirs[i] == ParamDirection::InOut) {
+            out_positions.push_back(i);
           }
-          if (returned_idx.has_value() && i != *returned_idx) {
+        }
+
+        // Same SSA-base disambiguation as GenerateSingleReturnAlias (issue
+        // #1702): for a multi-Out Group/Spmd wrapper FindReturnedParamIndex is
+        // unreliable — it returns nullopt (or mis-traces through the first Out)
+        // for outliner wrappers that end in the inner call, so the lineage
+        // would leak the FIRST Out (e.g. a GM scratch buffer) onto the result
+        // Var and every downstream scope reading the result would resolve to
+        // the wrong external tensor. The result Var is an SSA rename of the Out
+        // arg carrying the result, so its SSA base uniquely identifies that arg
+        // when it matches exactly one Out arg; trust that over returned_idx.
+        std::optional<size_t> matched_idx;
+        if (out_positions.size() > 1 && assign->var_) {
+          const std::string result_base = GetSSABaseName(assign->var_->name_hint_);
+          if (!result_base.empty()) {
+            for (size_t i : out_positions) {
+              auto arg_var = AsVarLike(call->args_[i]);
+              if (!arg_var) continue;
+              if (GetSSABaseName(arg_var->name_hint_) != result_base) continue;
+              if (matched_idx.has_value()) {  // ambiguous — abandon the match
+                matched_idx.reset();
+                break;
+              }
+              matched_idx = i;
+            }
+          }
+        }
+
+        // Prefer the unique SSA-base match, then the traced returned index,
+        // then (chosen == nullopt) the first Out — the legacy fallback.
+        std::optional<size_t> chosen = matched_idx.has_value() ? matched_idx : returned_idx;
+        for (size_t i : out_positions) {
+          if (chosen.has_value() && i != *chosen) {
             continue;
           }
           if (auto arg_var = AsVarLike(call->args_[i])) {
--- a/src/codegen/orchestration/orchestration_codegen.cpp
+++ b/src/codegen/orchestration/orchestration_codegen.cpp
@@ -2769,10 +2769,57 @@ class OrchestrationStmtCodegen : public CodegenBase {
     // result plus per-block GM scratch tensors used by a pl.spmd-dispatched
     // mixed kernel), out_indices[0] is the scratch buffer and aliasing the
     // call site SSA to it would route every downstream consumer into the
-    // scratch instead of the actual result. Fall back to out_indices[0] only
-    // when the return cannot be traced back to a Param.
+    // scratch instead of the actual result.
     auto returned_idx = FindReturnedParamIndex(callee, program_);
-    size_t param_idx = returned_idx.value_or(out_indices[0]);
+
+    // Group/Spmd wrappers produced by the outliner end in the inner kernel
+    // call, so they have no traceable top-level ReturnStmt and
+    // FindReturnedParamIndex either returns nullopt or — when an inner kernel's
+    // return is itself untraceable — mis-traces through the FIRST Out arg and
+    // reports the wrong index. For a multi-Out wrapper either outcome aliases
+    // the result to the wrong output (e.g. a small GM scratch), mis-routing
+    // every downstream consumer (issue #1702).
+    //
+    // The returned SSA result var is an SSA rename of the OUTPUT ARG carrying
+    // the result, so the result var's SSA base is the authoritative signal for
+    // which Out/InOut arg the call actually returns. Match it against each
+    // Out/InOut arg's SSA base; when it uniquely identifies one arg, trust it
+    // over FindReturnedParamIndex (which is unreliable for these wrappers).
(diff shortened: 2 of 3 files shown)


csy0225:
求跟踪下进展～

csy0225:
冲突啦～

PR #1722, opened 2026-06-09
fix(passes): outlined tile.store SSA scope violation via idempotent ConvertToSSA + escaping promotion (fixes #1693)

### LINKED ISSUE #1693: [Bug] Orchestration cross-assigns __ssa_v aliases for multiple in-place outputs of an inlined InCore scope

YunjiQin:
## Summary

Two stacked commits that together fix issue #1693 (SSA scope violation in outlined functions when ``tile.store`` is rewritten inside a loop body) by leaning on ConvertToSSA's existing escaping-variable promotion machinery.

### Commit 1 — ``fix(convert_to_ssa): make pass idempotent on already-SSA IR``

The pass doc advertises ``Preserves existing SSA structure`` and ``Keep existing SSA constructs unchanged``, but re-running on already-SSA IR with **dynamic shape / valid_shape / stride Var refs embedded in Tile/Tensor type annotations** dangled those refs. AssignStmt LHS is unconditionally re-versioned (``cur_[OLD] = NEW``), but every iter_arg / phi / return_var construction site read the type from the old Var verbatim — type-embedded refs to OLD survived and SSAVerify reported ``used before definition``.

Fix: thread every Var construction site that reads a type from an existing Var (or from ``TypeCollector.types[]``) through ``SubstType`` so renamed refs stay consistent. Sites: existing iter_args / loop_var / carried / escaping init types in ConvertFor and ConvertWhile; phi var, no-phi rv, mixed rv in ConvertIf. The previously-verbatim existing return_vars push is replaced with a freshen-and-update-cur_ loop; removed the now-redundant ``RegisterExistingReturnVars`` helper.

Tests: added ``TestIdempotence`` with two regression cases (IfStmt phi with dynamic valid_shape, ForStmt iter_arg with dynamic valid_shape). Both fail on baseline with ``used before definition`` and pass after the fix.

Doc: tighten ``04-convert_to_ssa.md`` to make the idempotence guarantee explicit (en + zh-cn).

### Commit 2 — ``fix(passes): rerun ConvertToSSA after outline passes (fixes #1693)``

OutlineScope's ``StoreEvalToAssignMutator`` rewrites EvalStmt-form ``pl.tile.store(t, ..., target)`` into ``_store_var = pl.tile.store(...)`` so the outlined function can return the post-store SSA name. When the store sits inside a for/while/if body, the synthesised AssignStmt is inside-loop while the ReturnStmt that references ``_store_var`` is at function-body depth — SSA scope violation, surfaced as the ``__FREE_VAR`` suffix the SSA verifier rejects.

Fix: re-run ``ConvertToSSA`` right after ``OutlineClusterScopes`` (the last outline pass). The escaping-variable promotion in ConvertToSSA detects "var defined inside loop, used after loop", synthesises IterArg + YieldStmt for each enclosing for/while, and ``FindInitValue`` picks the matching Out/InOut function param as the initial value. The outlined ``return _store_var`` becomes ``return _store_var__rv_vN`` referencing a function-scope yield result — SSA-legal, matches what the user would have written by hand with the equivalent native-DSL escaping store pattern. This relies on the idempotence guarantee from commit 1.

Pipeline-level effect: ``ConvertToSSA_post_outline`` is added between ``OutlineClusterScopes`` and ``ConvertTensorToTileOps``. The original 4th-position ``ConvertToSSA`` is untouched.

## Test plan

- [x] Idempotence regression tests pass (``TestIdempotence::test_rerun_on_if_with_dynamic_valid_shape``, ``test_rerun_on_for_loop_with_dynamic_valid_shape``)
- [x] Existing ConvertToSSA tests pass (65 cases) — no behavioural change on first invocation
- [x] Existing outline pass tests pass (74 cases) — they exercise ``outline_*()`` directly, unaffected by the new pipeline entry
- [x] Default strategy inventory test updated (``test_pass_manager_get_strategy_default`` — ``ConvertToSSA_post_outline`` added to ``TENSOR_ONLY_PASSES``)
- [x] Full ut suite: 5867 passed (3 pre-existing unrelated runtime failures match main — ``pypto.runtime.runtime_base.Worker.current`` attribute)
- [x] Issue #1693 repro: outlined function now emits ``for i, (a__iter_v0,) in pl.range(2, init_values=(a__ssa_v0,)): ... a__rv_v1 = pl.yield_(a__ssa_v2)`` followed by ``return a__rv_v1`` — SSA-legal, no ``__FREE_VAR``

## Related issues

Fixes #1693

### DIFF
--- a/docs/en/dev/passes/00-pass_manager.md
+++ b/docs/en/dev/passes/00-pass_manager.md
@@ -70,6 +70,7 @@ struct PassProperties {
 | NormalizeStmtStructure | TypeChecked | TypeChecked, NormalizedStmtStructure | — |
 | OutlineIncoreScopes | TypeChecked, SSAForm | SplitIncoreOrch | — |
 | OutlineClusterScopes | TypeChecked, SSAForm | ClusterOutlined | — |
+| ConvertToSSA (post-outline) | TypeChecked | TypeChecked, SSAForm | NormalizedStmtStructure |
 | ConvertTensorToTileOps | SplitIncoreOrch | IncoreTileOps | — |
 | LowerCompositeOps | — | — | — |
 | FlattenTileNdTo2D | SSAForm, IncoreTileOps | SSAForm, TileOps2D | — |
--- a/docs/en/dev/passes/00-pass_manager.md
+++ b/docs/en/dev/passes/00-pass_manager.md
@@ -71,6 +71,7 @@ struct PassProperties {
 | OutlineIncoreScopes | TypeChecked, SSAForm | SplitIncoreOrch | — |
 | OutlineClusterScopes | TypeChecked, SSAForm | ClusterOutlined | — |
 | ConvertToSSA (post-outline) | TypeChecked | TypeChecked, SSAForm | NormalizedStmtStructure |
+| NormalizeStmtStructure (post-outline) | TypeChecked | TypeChecked, NormalizedStmtStructure | — |
 | ConvertTensorToTileOps | SplitIncoreOrch | IncoreTileOps | — |
 | LowerCompositeOps | — | — | — |
 | FlattenTileNdTo2D | SSAForm, IncoreTileOps | SSAForm, TileOps2D | — |
--- a/docs/en/dev/passes/04-convert_to_ssa.md
+++ b/docs/en/dev/passes/04-convert_to_ssa.md
@@ -16,6 +16,11 @@ This pass transforms IR with multiple assignments to the same variable into SSA 
 
 **When to use**: Run this pass before any optimization or analysis that requires SSA form (e.g., OutlineIncoreScopes, memory optimization passes).
 
+**Pipeline positions**: The default tensor-optimization strategy invokes `ConvertToSSA` twice:
+
+1. **4th pass** — initial SSA conversion of user-written DSL (before any outlining).
+2. **Immediately after `OutlineClusterScopes`** (entry: `ConvertToSSA_post_outline`) — re-runs SSA on the outlined functions. Outlining lifts `tile.store` results from `EvalStmt` form into inside-loop `AssignStmt`s; the second pass uses **escaping-variable promotion** (see Algorithm step 4) to thread those defs through `IterArg` / `YieldStmt` to function scope, so the outlined `ReturnStmt` references SSA-valid Vars. This relies on the pass being **idempotent** (see "Mixed SSA/non-SSA" / "Idempotent" above): the first pass's outputs survive the second pass unchanged.
+
 ## API
 
 | C++ | Python | Level |
--- a/docs/en/dev/passes/04-convert_to_ssa.md
+++ b/docs/en/dev/passes/04-convert_to_ssa.md
@@ -10,6 +10,7 @@ This pass transforms IR with multiple assignments to the same variable into SSA 
 - **If statements**: Variables modified in one or both branches
 - **For loops**: Variables modified inside the loop body
 - **Mixed SSA/non-SSA**: Preserves existing SSA structure while converting non-SSA parts
+- **Idempotent**: Re-running on already-SSA IR produces SSA-equivalent IR with no use-before-def violations, including for Var refs embedded in dynamic Tile/Tensor type annotations (shape, valid_shape, stride). Every Var construction site that reads a type from an existing Var threads the type through `SubstType` so renamed refs stay consistent.
 
 **Requires**: `TypeChecked` property. `TypeChecked` is verified automatically at BASIC level once produced; use a `VerificationInstrument` via `PassContext` to validate required properties before this pass runs.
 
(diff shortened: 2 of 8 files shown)


YunjiQin:
Re: Gemini's review on `FindInitValue` — I applied the suggested fix in 69a0fab2 (comparing unsubstituted types). It locally passed the full ut suite (5904 tests), but on CI it regressed `tests/st/distributed/test_l3_ep_dispatch_combine.py::test_ep_dispatch_combine` (`max diff = 40.5`, previously the test passed with f3f518c3). Reverted in 768eea80.

The original concern is theoretical — `SubstType` does mint a new `shared_ptr` when type-embedded Vars are renamed, but in the current ConvertToSSA escape-promotion path the substituted Out/InOut param type and the substituted escape var type happen to share original Var pointers (per-function `cur_` only has outer-scope renames at scope entry), so pointer equality still holds for the common case. The dynamic-shape edge case Gemini described doesn't fire on any existing test, and the proposed "original-type comparison" path has at least one structural-equality assumption (`orig_params_[i]->GetType() == orig_type` for the param branch) that diverges from what's expected by some real distributed program — surfacing as the `ep_dispatch_combine` regression.

Keeping the original substituted-type comparison for now. Leaving this thread open so it stays visible if the dynamic-shape edge case ever surfaces — a future fix would need a separate test demonstrating the failure first.

YunjiQin:
Update on the FindInitValue thread: re-applied the fix in 86523a7c. Earlier I attributed the dist-system-tests regression (test_l3_ep_dispatch_combine, max diff = 40.5) to this change, but on closer inspection that test has a known intermittent synchronisation precision bug unrelated to ConvertToSSA. The original-type comparison is the correct semantics — keeping it.

YunjiQin:
Superseded by upstream #1747 (d86f86bc), which fixes #1693 via a different (and more thorough) approach: ScopeOutliner / ExpandMixedKernel / NormalizeReturnOrder now emit explicit param-return wrappers under a new ReturnParamsExplicit IR property + verifier, so outlined function ReturnStmts reference InOut params directly — the original SSA scope violation root cause is gone, no post-outline ConvertToSSA re-run needed.

Verified locally: the original #1693 repro on plain upstream/main with d86f86bc emits SSA-legal IR (`return a__ssa_v0, b__ssa_v0`) with no `__FREE_VAR` placeholder. Closing this PR.

PR #1726, opened 2026-06-09
feat(distributed): DimExpr IR node: composite dynamic dims (NR*SIZE) at all IR scopes + dynamic NR parity for AG/RS/Bcast STs

georgebisbas:
## Summary

Introduces **`DimExpr`** — a first-class IR node for composite dynamic dimensions — that makes `NR * SIZE` work at **every** IR scope (HOST, Orch, InCore, chip, distributed), not just the top-level HOST scope. 

## The Problem

Before this PR, `pl.dynamic("NR")` worked as a bare `Var` in shapes. But composite dims like `NR * 64` were raw `Mul(Var("NR"), ConstInt(64))` nodes sitting in a shape list — indistinguishable from runtime arithmetic. This caused three concrete failures:

1. Any attempt to use `NR * 64` in an InCore or Orch tensor shape failed because passes at those scopes walked into the `Mul` node and treated it as runtime code.

2. **SSA verifier rejected all-composite-dim kernels.** `Tensor[[M * 2, N * 3]]` → the verifier did `As<Var>(dim)` on each shape element → `As<Var>(Mul(...))` returned `null` → `M` and `N` were never registered as defined → "undefined variable" error.

3. **Distributed codegen couldn't extract dim values.** The orchestrator codegen had no way to find `Var("NR")` inside a composite dim expression to emit the runtime shape-extraction code.

## The Solution: `DimExpr`

`DimExpr` wraps a composite dimension expression (`NR * 64`) and signals: *"this is a compile-time type annotation, not a runtime value."*

```cpp
class DimExpr : public Expr {
 public:
  ExprPtr body_;  // IgnoreField — visitors, verifiers, serialization skip this

  static constexpr auto GetFieldDescriptors() {
    return std::tuple_cat(
        Expr::GetFieldDescriptors(),
        std::make_tuple(reflection::IgnoreField(&DimExpr::body_, "body")));
  }
};
```

### Key design decisions

**`body_` is `IgnoreField`.** Every automatic visitor walk, verifier, and serializer skips the wrapped expression by default. A new pass or verifier added tomorrow handles composite dims correctly with zero code — no per-pass workarounds needed.

**Intentional unwrap on demand.** Passes that need the inner expression use `As<DimExpr>(expr)->body_`:

| Consumer | How it unwraps |
|----------|---------------|
| `structural_equal` | Unwraps before TypeName check (so `Add` vs `DimExpr(Add)` normalize to `Add` vs `Add`) |
| `AreExprsEqual` | Unwraps before ConstInt/BinaryExpr comparison |
| `SSAVerifier::RegisterTypeVarsFromExpr` | Recursively unwraps `DimExpr` → `BinaryExpr` → `Var` to register dim Vars |
| `DistributedCodegen::EmitShapeDimVarDefs` | Extracts `Var` nodes from DimExpr bodies to emit Python orchestrator defs |
| Printer | Unwraps to inner expression, with `typed_index_consts_` enabled to prevent const-folding |
| Codegen (PTO, distributed) | Unwrap: `GenerateExprString(dim_expr->body_)` |

**Opaque to default IR traversal.** `IRVisitor::VisitExpr_(DimExpr)` is a no-op. `IRMutator::VisitExpr_(DimExpr)` returns identity. This keeps type-annotation variables out of runtime SSA scope checks.

**Full IR plumbing.** `DimExpr` has its own `ObjectKind::DimExpr`, `KindTrait`, visitor/mutator/functor dispatch, Python bindings (`ir.DimExpr`, `ir.dim_expr()`), and `.pyi` stubs.

### DSL impact

```python
# This now works at ANY scope — InCore, Orch, HOST, distributed:
NR = pl.dynamic("NR")
SIZE = 64

@pl.function(type=pl.FunctionType.InCore)
def kernel(self, a: pl.Tensor[[NR * SIZE, K], pl.FP32]) -> ...:
    out_tile = pl.load(a, [0, 0], [NR * SIZE, K])  # composite dim in load
    ...
```

The DSL user writes `NR * SIZE` naturally — the parser wraps the `Scalar` arithmetic in `ir.dim_expr(...)` automatically.


georgebisbas:
Superseded by https://github.com/hw-native-sys/pypto/pull/1781 #1781

PR #1747, opened 2026-06-11
fix(ir): Return params explicitly from outlined/wrapper functions

### LINKED ISSUE #1573: [Bug] Orchestration codegen scrambles phi assignments when one task writes 3+ outer-loop carries

### LINKED ISSUE #1693: [Bug] Orchestration cross-assigns __ssa_v aliases for multiple in-place outputs of an inlined InCore scope

### LINKED ISSUE #1702: [Bug] Orchestration codegen aliases a multi-output spmd scope's return value to the wrong output tensor (out_indices[0] fallback) -> numel-mismatched reshape -> 507018

lyfne123:
## Summary

Fixes #1702. Also resolves the same alias-miscompile class for #1693 and structurally closes #1573.

A multi-output `pl.spmd` scope whose return value is a **non-first** output was aliased by orchestration codegen to the **wrong** output tensor. The Spmd wrapper has no top-level `ReturnStmt` the codegen can trace, so `GenerateSingleReturnAlias` fell back to `out_indices[0]` (the first `Out`/`InOut`). Downstream the wrong tensor was reshaped; the numel mismatch tripped an AICPU `valid_reshape` assert, and the host surfaced it as a scheduler timeout (`507018`) — a kernel-side shape bug masquerading as a deadlock.

The fix moves the mapping from a codegen heuristic to an IR-level invariant.

## Approach — `ReturnParamsExplicit` invariant

Every `InCore`/`Group`/`Spmd` function returns the param `Var` each tensor result writes through, **by pointer identity**, so the return→param map is a lookup rather than a re-derivation. Kernel-allocated outputs and scalars are exempt.

- **New shared utility `ir::return_lineage`** (`return_lineage_utils`): traces a return value through var copies, loop carries, builtin writebacks (`tile.store` / `tensor.assemble` / `tensor.set_validshape`), `TupleGetItem` of tuple calls, and `Group`/`Spmd` wrappers ending in the inner call. Orchestration analysis `FindReturnedParamIndex/Indices` delegate to it.
- **Producers**: `ScopeOutliner` and `ExpandMixedKernel` build wrapper `ReturnStmt`s from params; `NormalizeReturnOrder` canonicalizes `InCore`/`Group`/`Spmd` returns to params (new step before reordering) and produces `ReturnParamsExplicit`.
- **Verification**: new `ReturnParamsExplicit` property + verifier, registered and added to the Basic verification set; the `InOutUse` verifier skips `Spmd` wrappers.
- **Codegen**: `GenerateSingleReturnAlias` now `INTERNAL_CHECK`s instead of guessing the first `Out` when a multi-output trace fails.

## Testing

- [x] Full unit-test suite green (5709 passed, 1 skipped)
- [x] New regression tests for #1702 and #1693 + a verifier unit test
- [x] Before/after expectations updated across the outlining, mixed-kernel, normalize-return-order, and orchestration-codegen suites (no assertions weakened)
- [x] clang-tidy and all pre-commit hooks pass
- [x] Docs updated (en + zh-cn): pass docs 10/11/21/24 and codegen 01

## Related Issues

Fixes #1702
Also resolves #1693; structurally closes the #1573 class


Hzfengsy (approved):
LGTM. The two gemini-code-assist suggestions (prefix-check via std::string::compare, splitting the bounds/type check) are minor style/efficiency nits and non-blocking.

Issue #975, opened 2026-04-12
[Pass Bug]SplitVectorKernel fails on singleton broadcast tile loads in split AIV kernels


ndleslx:
## Summary

`SplitVectorKernel` currently fails when an AIV split kernel contains a broadcast tile load whose split-axis extent is `1`, for example a tile like `[1, 128]` under `SplitMode.UP_DOWN`.

The pass currently tries to halve the split dimension unconditionally, which throws:

```text
SplitVectorKernel requires an even split dimension, got 1
```

## Repro pattern

This showed up in a Qwen decode scope where the AIV side loads RMS weight chunks as a broadcast tile and then uses `col_expand_mul`:

```python
gamma_tile: pl.Tile[[1, 128], pl.FP32, pl.MemorySpace.Vec] = pl.load(
    gamma, [0, 0], [1, 128], target_memory=pl.MemorySpace.Vec
)
result: pl.Tile[[16, 128], pl.FP32, pl.MemorySpace.Vec] = pl.col_expand_mul(prev, gamma_tile)
```

Under `SplitMode.UP_DOWN`, the split axis is dim `0`. The pass attempts to halve `gamma_tile` from `[1, 128]`, which is invalid.

## Expected behavior

Singleton broadcast dimensions on the split axis should be preserved, not halved.

Examples:
- normal tile `[16, 128]` with `UP_DOWN` should become `[8, 128]`
- broadcast tile `[1, 128]` with `UP_DOWN` should remain `[1, 128]`

## Actual behavior

The pass treats both cases the same and throws on the singleton case.

## Root cause

In `src/ir/transforms/split_vector_kernel_pass.cpp`, the pass halves split-dimension tile shapes and load shape tuples unconditionally for AIV tile loads / tile-producing ops. That is correct for real split tiles but incorrect for broadcast operands whose split-dimension extent is statically `1`.

## Suggested fix

When rewriting split AIV tile shapes / `tile.load` shapes, preserve singleton dimensions on the split axis instead of forcing halving. Offset adjustment and tracked split-tile metadata should only be applied when the split is actually performed.

## Regression test idea

Add a unit test for an `AIV` function with `attrs={"split": pl.SplitMode.UP_DOWN}` that does:

- `prev = pl.load(data, [0, 0], [16, 128], target_memory=pl.MemorySpace.Vec)`
- `gamma_tile = pl.load(gamma, [0, 0], [1, 128], target_memory=pl.MemorySpace.Vec)`
[...]

PR #1757, opened 2026-06-11
feat(st): add L3 ring allreduce ST (RS+AG schedule, monolithic InCore)

georgebisbas:
Port the ring allreduce algorithm from simpler allreduce_ring_distributed (#975) into the PyPTO DSL as a new L3 system test.

Single monolithic InCore kernel (ring_allreduce) collapses all four phases — stage-in, reduce-scatter loop, allgather loop, stage-out — into one AIV task dispatch per rank, eliminating 2*(P-1) kernel-launch overhead vs a multi-kernel chain.

Exchange publish buffer (1*CHUNK FP32 HCCL window) provides an explicit commit-point before each RS/AG barrier, matching simpler's publish pattern and ensuring data visibility across rounds.

Dynamic NR via pl.dynamic('NR'), compact signal layout [2*(NR-1), NR] int32, parameterized tests for P=2 and P=4.  Golden matches mesh allreduce (element-wise sum of all rank inputs).

References:
- simpler ring allreduce: hw-native-sys/simpler#975
- mesh allreduce for comparison: test_l3_allreduce.py
- PTOAS MTE drain fix (enables multi-round barriers): hw-native-sys/PTOAS#718

Docs: distributed_ops.md (en+zh) updated with ring ST entry.


georgebisbas:
Depends on #1758

Issue #1042, opened 2026-04-16
[Bug] Swimlane diagram not generated when precision comparison fails in run interface


lyfne123:
### Component

Codegen

### Description

When using the `run` interface (`python/pypto/runtime/runner.py`), if the precision comparison step fails (i.e., the output does not match the expected results within tolerance), the swimlane diagram (timeline visualization) is not generated.

The swimlane diagram is a valuable debugging and performance analysis tool. Even when precision comparison fails, users still need the swimlane diagram to understand kernel execution timing, identify performance bottlenecks, and debug the issue. The diagram generation should not be gated on precision comparison success.

### Steps to Reproduce

1. Write a program using the `run` interface with `runtime_profiling=True`
2. Ensure the kernel produces output that does not match the expected reference (precision mismatch)
3. Run the program
4. Observe that no swimlane diagram is generated in the output directory

### Expected Behavior

The swimlane diagram should still be generated even when precision comparison fails. The precision comparison failure should be reported as an error/warning, but should not prevent the swimlane diagram from being produced.

### Actual Behavior

When precision comparison fails, the swimlane diagram is not generated, losing the ability to visualize kernel execution timeline for debugging.

### Relevant Code

**`python/pypto/runtime/runner.py` — `run()` function (lines 298–364)**

The `run()` function wraps the entire execution in a `try/except` block. If any exception is raised during `_execute_on_device`, it is caught and the function returns `RunResult(passed=False)`:

```python
    try:
        # ...
        with _stage("device_execution"):
            _execute_on_device(
                work_dir,
                golden_path,
                chip_callable,
                runtime_name,
                config.platform,
                config.device_id,
                runtime_profiling=config.runtime_profiling,
            )

        profile_data = prof.to_dict() if prof is not None else None
        return RunResult(passed=True, execution_time=time.time() - start_time, profile=profile_data)

    except Exception:
        profile_data = prof.to_dict() if prof is not None else None
        return RunResult(
            passed=False,
            error=traceback.format_exc(),
            execution_time=time.time() - start_time,
            profile=profile_data,
        )
```

**`python/pypto/runtime/runner.py` — `_execute_on_device()` (lines 432–461)**

Inside `_execute_on_device`, the swimlane collection (line 446) is called **before** `validate_golden` (line 456). However, if `execute_on_device` itself raises an exception (e.g., device runtime error), the swimlane collection is skipped entirely:

```python
    # Execute
    if runtime_profiling:
[...]

PR #1769, opened 2026-06-15
chore(runtime): bump runtime to 2668be5a; type register() handles as Any

lyfne123:
## Summary
- Bump the `runtime` submodule `48980572..2668be5a` (33 commits).
- The only API-surface change touching pypto is runtime #891: `Worker.register()` now returns an opaque `CallableHandle` instead of an `int`. pypto threads that result straight back into `run()` / `submit_sub()` / `submit_next_level()`, which all accept the handle, so **no functional change is needed** — only the `int` type annotations on the cached/returned handles became inaccurate.
- Relax those annotations to `Any` (matching the files' existing style for simpler-side types):
  - `worker.py`: `_cid_cache`, `RegistrationHandle.__init__(cid)`, `RegistrationHandle.cid`
  - `distributed_runner.py`: `sub_ids` / `chip_cids` in `_register_callables` and `_dispatch`

The bump also brings additive features (per-task ring sizing via `CallConfig.runtime_env` #1042, typed/selective scalar dump, a5 SIMT launch, remote L3, etc.); none change the orchestration ABI that pypto codegen targets (`rt_submit_*`, `TaskOutputTensors`, `make_tensor_external` unchanged).

## Testing
- [x] Reinstalled `simpler` from the bumped runtime (fresh `_task_interface.so`) and confirmed the new API is live (`register() -> CallableHandle`, `CallConfig.runtime_env`).
- [x] `a2a3sim` end-to-end: `test_hello_world_add` PASS.
- [x] On-device `a2a3`: `test_hello_world_add` PASS.
- [x] Code review completed.

## Related Issues
N/A

### DIFF
--- a/python/pypto/runtime/distributed_runner.py
+++ b/python/pypto/runtime/distributed_runner.py
@@ -104,6 +104,25 @@ def _load_generated_module(path: Path) -> Any:
     module = importlib.util.module_from_spec(spec)
     sys.modules[module_name] = module
     spec.loader.exec_module(module)
+    # Generated modules live only in ``sys.modules`` — there is no
+    # ``_pypto_generated`` package on disk to re-import them by name. The
+    # runtime cloudpickles every registered callable to derive its hashid
+    # descriptor (runtime #891); without this, cloudpickle would serialize
+    # functions from this module *by reference* and fail to re-import
+    # ``_pypto_generated.<stem>`` (PicklingError). Force by-value pickling so
+    # the function code travels inside the payload.
+    #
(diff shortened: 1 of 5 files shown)


Hzfengsy (approved):
Runtime bump + handle re-typing looks correct. The two gemini-code-assist comments are minor: the return-type annotation is a consistency nit, and the id()-keyed cache is a pre-existing pattern (only the value type changed here) — both out of scope for this runtime-bump chore and can be addressed separately.

PR #1777, opened 2026-06-15
refactor(pass): fold Ascend910B load+tpop hazard into MemoryReuse, remove LegalizePTOBufferReuse

lyfne123:
## Summary

Removes the `LegalizePTOBufferReuse` pass and folds its one load-bearing job — the Ascend910B split-AIV `tile.load` + `tile.tpop_from_aic` in-place hazard split — into `MemoryReuse`, which already owns every buffer-coalescing decision.

## Motivation

`LegalizePTOBufferReuse` ran right after `MemoryReuse` to undo illegal PTO buffer sharing. An experiment (disabling the pass and reading CI) showed:

- Its **signature-legalisation path is unreachable in practice**: `MemoryReuse` only coalesces tiles with identical dtype *and* shape (`AreTileTypesCompatible`), so every sharing it produces is already PTO-materialisable. The pass's own unit tests constructed the illegal sharing by hand.
- The **only real dependency** was a single system test, `tests/st/runtime/ops/test_trans.py::TestTransposeColumnOperations::test_nd3d_transpose[a2a3]`, which silently produced wrong numerics on Ascend910B (cube transpose → `tpop_from_aic`) without the pass. Ascend950 was unaffected.

Rather than coalesce-then-split, `MemoryReuse` now never forms the hazardous sharing.

## What changed

**MemoryReuse** (`src/ir/transforms/memory_reuse_pass.cpp`)
- Collects load-derived and tpop-consuming vars in one IR walk — only when the function is Ascend910B split-AIV (gated by `RequiresSplitLoadTpopWorkaround()`).
- In the reuse decision, blocks a writer from in-place-reusing a load buffer it reads when the writer also consumes a `tile.tpop_from_aic` value. This reuses the existing in-place "buffer still occupied at def" detection, so the guard is precise. Off-910B the inputs are empty and reuse behaviour is byte-for-byte unchanged.

**Removal**
- `LegalizePTOBufferReuse` pass + factory + binding + stub + pipeline entry + CMake source + dedicated test, all deleted.
- `FoldNoOpReshape` / `BackendHandler` doc comments updated to point at `MemoryReuse`.

**Tests**
- Hazard coverage migrated to `test_memory_reuse.py`: Ascend910B split-AIV blocks the `tile.add(load, tpop)` output from reusing the load buffer; Ascend950 still reuses it.
- `test_pass_manager` strategy assertions updated.

**Docs**
- Deleted the pass doc, added an "Ascend910B load + tpop_from_aic hazard" section to the MemoryReuse doc, renumbered the passes-doc sequence, and synced zh-cn + the pass-doc-ordering rule.

## Testing

- Build: clean (`pypto_core` builds).
- `tests/ut/ir/transforms/`: **1262 passed, 25 skipped** (incl. the new hazard tests).
- pre-commit: clang-format, cpplint, ruff, pyright all pass.
- The end-to-end `test_nd3d_transpose[a2a3]` regression runs on CI (Ascend sim).


lyfne123:
Follow-up: the real root cause of the `test_nd3d_transpose[a2a3]` regression was **`tile.transpose` aliasing its output onto the source buffer** (in-place `pto.ttrans`, which ptoas cannot do — silent corruption). The removed LegalizePTOBufferReuse's signature-split path was the only thing fixing it. Added a root-cause fix: `tile.transpose` is now `not_inplace_safe()` and InitMemRef gives non-in-place-safe inherit-input ops their own buffer (inheriting only the memory space). The transpose output now lands on a distinct buffer, so `pto.ttrans` gets `src != dst`. Verified locally via pass dumps + full ir/codegen/transforms unit suites (green).

lyfne123:
Refined the transpose fix to the precise root cause. The in-place `pto.ttrans` corrupts only when the transpose **input is a sub-region of a larger live buffer** (nd3d: `trans_page_0 = slice(x_tile)` → the fractal-padded write smears into the neighbouring per-page slices). When the input is a **whole tile** (softmax_pool `[128,64]→[64,128]`) the scratch `tmp` stages the data and there's nothing else to clobber → in-place is safe and saves memory. InitMemRef now splits the transpose output to a fresh buffer only in the sub-region case, so `softmax_pool` keeps its in-place sharing (fixes the prior Vec +1280B overflow) and nd3d gets a distinct `src ≠ dst`. Verified locally: nd3d → fresh buffer, softmax_pool → in-place (compiles past AllocateMemoryAddr), 1645 transforms + 4294 ir/codegen unit tests green.

lyfne123:
Final iteration. Removed the `not_inplace_safe()` marker — it was self-contradictory with the intentional whole-tile in-place sharing and corrupted `decode_attention_swa` (a whole-tile [16,8]→[8,16] transpose that is correct in-place). The fix is now purely in InitMemRef: a transpose output gets a fresh buffer **only when its input is a sub-region of a larger live buffer** (nd3d's per-page slices of the flattened load); whole-tile transposes (swa, hca softmax_pool) keep their in-place sharing. MemoryReuse needs no marker — it can't re-create the hazard because the sub-region source is a slice of a still-live buffer, never a valid in-place reuse target (verified: nd3d trans_0 stays on its fresh buffer post-MemoryReuse). Net effect = baseline behavior except nd3d's sub-region transpose splits. Local: nd3d split, swa/hca in-place (no overflow), 1645 transforms + 4294 ir/codegen green.

lyfne123:
Rebased onto #1783 (re-enables the DeepSeek-V4 attention examples). For the record: the `decode_attention_swa` `x_out` ~22% mismatch is the **known external pypto-lib / pto-isa bf16 regression** that #1783's removed comment documents ("x_out ~17-22% mismatch ... introduced in pypto-lib / pto-isa, not in pypto") — not from this PR. The genuinely-mine issue (hca softmax_pool Vec overflow from an earlier all-split approach) is fixed by the final InitMemRef sub-region split. All pypto-owned CI was green pre-rebase (system-tests/nd3d fixed, unit/codegen/dist/a5sim).

Hzfengsy (approved):
Well-engineered refactor. The load+tpop hazard guard is strictly gated to Ascend910B AIV split-mode functions, so off-910B the HazardInputs stay empty and reuse behaviour is byte-for-byte unchanged. The collector is a single O(N) forward walk. Removing LegalizePTOBufferReuse is justified — its signature-legalization path was unreachable (MemoryReuse only coalesces identical dtype+shape), and the one real dependency (test_nd3d_transpose[a2a3]) is covered by folding the guard into the reuse decision. CI green incl. system-tests. LGTM.

PR #1781, opened 2026-06-15
feat(ir): add DimExpr IR node for composite dynamic dimensions

georgebisbas:
### Summary

Add `DimExpr` — a first-class compile-time IR expression node (`ObjectKind::DimExpr`) that wraps composite dimension expressions like `NR * SIZE` with `IgnoreField` semantics on `body_`. This prevents passes (ConvertToSSA, InitMemRef, SSAVerify) from walking into type-annotation Vars inside composite dims, which would otherwise fatally break tile allocation when composite shapes appear at the InCore body level.

### Motivation

PR #1765 (already merged) stores composite dims as bare `Mul(Var, ConstInt)` directly in `TensorType` shapes. While sufficient for parametric types (function signatures), it breaks when composite dims appear in InCore body expressions like `pl.create_tensor([NR * SIZE, 1])` — ConvertToSSA renames the `Var("NR")` inside the type-annotation Mul, and InitMemRef then rejects the tile because it can't find NR's definition in scope.

`DimExpr` solves this by wrapping the arithmetic in a node that every visitor/verifier/serializer treats as opaque by default (via `IgnoreField`), while expression functors (arith analyzers) auto-unwrap for free (via a non-pure-virtual default).

### What's in this PR

**Phase 1 — IR Foundation:**
- C++ `DimExpr` class (`include/pypto/ir/scalar_expr.h`) with `body_` as `IgnoreField`
- Visitor (no-op), Mutator (identity), ExprFunctor (auto-unwrapping default)
- Structural equality: DimExpr(Mul) ⇔ bare Mul
- Python printer: unwraps `DimExpr` to print its body
- Codegen: unwrap in `GenerateExprString`, `CollectVarsFromShapeExprImpl`, `GetExprAsCode`
- SSA verifier: register Vars inside `DimExpr` bodies in outermost scope

**Phase 2 — Python Bindings + Parser:**
- `nb::class_<DimExpr, Expr>` binding + `ir.dim_expr()` factory
- Type stubs in `ir.pyi`
- Parser auto-wraps composite Scalar dims in `_validate_dim_value` and `python_value_to_ir`

**Tests (differentiating from #1765):**
- `test_composite_dim_is_wrapped_in_dimexpr`: parser produces `DimExpr`, not bare `Mul`
- `test_dimexpr_survives_print_parse_roundtrip`: round-trip preserves structure
- `test_bare_dynvar_not_wrapped_in_dimexpr`: bare `pl.dynamic()` stays as `Var`
- `test_dimexpr_collect_vars_from_shape`: codegen correctly finds Vars inside `DimExpr`

### What's NOT in this PR (deferred to follow-up)

- Distributed ST generalization (allgather, reduce_scatter, broadcast → N-rank with `NR * SIZE`)
- Host orchestration codegen (`EmitShapeDimVarDefs` for DimExpr bodies)
- Dev docs for DimExpr

### Design decisions

| Decision | Rationale |
|----------|-----------|
| `body_` is `IgnoreField` | Prevents ConvertToSSA from renaming type-annotation Vars |
| ExprFunctor default recurses into `body_` | Arith analyzers auto-unwrap with 0 changes |
| IRVisitor is no-op, IRMutator is identity | Default traversal doesn't descend into composite dims |
| DimExpr is NOT a `VarLike` | Same rationale as MemRef (compile-time vs runtime semantics) |
| Compile-time only — fully resolved before codegen reaches NPU | Zero runtime impact |

### DIFF
--- a/include/pypto/ir/core.h
+++ b/include/pypto/ir/core.h
@@ -83,6 +83,12 @@ enum class ObjectKind {
   BitNot,
   Cast,
 
+  // Dimension expression — wraps arithmetic on pl.dynamic() dims.
+  // Treated as program-scoped by the SSA verifier so composite shape
+  // expressions (e.g. ``NR * 64``) can appear in type annotations at
+  // any IR function level.
+  DimExpr,
+
   // Statement kinds
   AssignStmt,
   IfStmt,
--- a/include/pypto/ir/kind_traits.h
+++ b/include/pypto/ir/kind_traits.h
@@ -81,6 +81,7 @@ DEFINE_KIND_TRAIT(Neg, ObjectKind::Neg)
 DEFINE_KIND_TRAIT(Not, ObjectKind::Not)
 DEFINE_KIND_TRAIT(BitNot, ObjectKind::BitNot)
 DEFINE_KIND_TRAIT(Cast, ObjectKind::Cast)
+DEFINE_KIND_TRAIT(DimExpr, ObjectKind::DimExpr)
 
 // Statement types
 DEFINE_KIND_TRAIT(AssignStmt, ObjectKind::AssignStmt)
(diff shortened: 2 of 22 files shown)


georgebisbas:
Closing this approach after discussion with @YunjiQin . Will open an issue on that.

PR #1793, opened 2026-06-16
feat(distributed): pl.nranks_dim — first-class DSL primitive for distributed rank-count dimensions

georgebisbas:
### Why

Distributed kernels express per-rank buffer shapes in terms of the world size — `signal: DistributedTensor[[NR, 1], INT32]` where `NR` is the number of ranks.  Before this PR, the only way to express that dimension was `pl.dynamic("NR")`, a string-keyed general-purpose mechanism that carries no semantic distinction between rank-count dimensions and ordinary dynamic axes.
This created three problems:

1. **No compiler awareness.** The compiler had no way to distinguish "this dimension is the rank count" from "this dimension is an arbitrary runtime value."  Downstream passes and codegen had to resort to fragile name matching (`"NR"` substring heuristics) that silently broke on any deviation.
2. **Leaky abstraction.** The rank count is a first-class distributed concept — it should be a named DSL primitive, not a string argument to a generic dynamic-dimension function.
3. **No misuse detection.** Writing `pl.load(data, [NR, 0], ...)` in a tile expression would silently produce broken IR because `NR` resolved like any ther dynamic dimension, leaking into tile shapes where only `ConstInt` is legal.

### What

This PR introduces `pl.nranks_dim` — a singleton `DynVar` that is semantically distinct from `pl.dynamic()`:

```python
NR = pl.nranks_dim  # singleton — no parentheses, not a function call

@pl.function
def kernel(self,
    signal: pld.DistributedTensor[[NR, 1], INT32],  # ← type annotations only
) -> ...:
    nranks = pld.nranks(ctx)
    for peer in pl.range(nranks):   # ← dynamic loop bound (runtime value)
        chunk = pl.load(..., [1, SIZE])  # ← tile shapes always ConstInt
```

Contract:

- pl.nranks_dim must only appear in type annotations (parameter shapes, return types). It carries an is_nranks_dim_ flag on the underlying ir::Var.

- Tile shapes (pl.load / pl.store / pld.tile.remote_load) remain ConstInt — the rank count only controls how many times the static-tile loop body runs, never what the tile looks like.
- The compiler rejects misuse in expression position with a clear error:
"pl.nranks_dim must only appear in type annotations, not in expressions."
- Resolution is structural — no name matching. The is_nranks_dim_ flag
is the single source of truth across all layers.

### DIFF
--- a/.claude/rules/pass-doc-ordering.md
+++ b/.claude/rules/pass-doc-ordering.md
@@ -16,6 +16,7 @@ Developers read pass docs sequentially to understand the compilation pipeline. I
 | 01 | `01-inline_functions.md` | 1st pass |
 | 02 | `02-unroll_loops.md` | 2nd pass |
 | 03 | `03-ctrl_flow_transform.md` | 3rd pass |
+| — | `resolve_distributed_shape_vars.md` | Runs after `CtrlFlowTransform`, before `ConvertToSSA` (distributed only — see `docs/en/dev/passes/resolve_distributed_shape_vars.md`). **TODO: assign slot 04, renumber 04–39 → 05–40.** |
 | 04 | `04-convert_to_ssa.md` | 4th pass |
 | 05 | `05-simplify.md` | 5th pass (also runs as the last pass of the tile pipeline) |
 | 06 | `06-flatten_call_expr.md` | 6th pass |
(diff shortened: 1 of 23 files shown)


georgebisbas:
Closing this apporach after discussion with @YunjiQin

PR #1802, opened 2026-06-17
fix(passes): track split axis through transpose/reshape in SplitVectorKernel

### LINKED ISSUE #1790: [Bug] SplitVectorKernel (pl.split UP_DOWN) rescales reshape result TYPE but not its shape ARGUMENT → memory_reuse OOB abort (size 64 into 32)
### Diagnosis

**pypto** (compile-time codegen) — `SplitVectorKernel` (the `pl.split(pl.SplitMode.UP_DOWN)` lowering) row-splits a vector scope by halving tile **types**, but leaves the explicit shape **argument** of `pl.tile.reshape` un-rescaled. A later `memory_reuse` pass then sizes the reshape output from the stale (un-split) argument and aborts when it cannot fit that into the split-sized reuse slot.

### Description

Adding `optimizations=[pl.split(pl.SplitMode.UP_DOWN)]` to the fused single-`pl.spmd` scope in `models/deepseek/v4/hc_pre.py` (DSv4 `hc_pre`) aborts at **compile time** (a2a3sim `--compile-only`, deterministic 6/6):

```
pypto.InternalError: Internal error: sharing-group member offset 0
rebased to rel=0 size=64 falls outside reuse target (size 32)
Check failed: rel >= 0 && (uint64_t)rel + old->size_ <= (*source_memref)->size_
  at src/ir/transforms/memory_reuse_pass.cpp:1328
C++ traceback: memory_reuse_pass.cpp:1859 -> fuse_create_assemble_to_slice_pass.cpp:124 (x3) -> memory_reuse_pass.cpp:1361
```

**Root cause (from the pass IR dump `30_after_InitMemRef.py`, i.e. after `SplitVectorKernel` at pass 24).** The `mix_x` epilogue builds 4 per-head scalar columns via transpose + reshape:

```python
pre_eps_t = pl.transpose(pre_eps, axis1=0, axis2=1)          # [HC_PAD, T_TILE]
pre0 = pl.mul(pl.reshape(pre_eps_t[0:1, 0:T_TILE], [T_TILE, 1]), 1.0)   # T_TILE == 16
pre1 = pl.mul(pl.reshape(pre_eps_t[1:2, 0:T_TILE], [T_TILE, 1]), 1.0)
pre2 = pl.mul(pl.reshape(pre_eps_t[2:3, 0:T_TILE], [T_TILE, 1]), 1.0)
pre3 = pl.mul(pl.reshape(pre_eps_t[3:4, 0:T_TILE], [T_TILE, 1]), 1.0)
```

After UP_DOWN the dumped IR shows the reshape **result type halved to `[8,1]`** (32 B) while the reshape **shape argument stays `[16, 1]`** (64 B):

```python
# 30_after_InitMemRef.py
t__tmp_v15: pl.Tile[[8, 1], pl.FP32, pl.MemRef(mem_vec_40, 0,   32), pl.Mem.Vec] = pl.tile.reshape(t__tmp_v14, [16, 1])
t__tmp_v17: pl.Tile[[8, 1], pl.FP32, pl.MemRef(mem_vec_40, 64,  32), pl.Mem.Vec] = pl.tile.reshape(...,        [16, 1])
t__tmp_v19: pl.Tile[[8, 1], pl.FP32, pl.MemRef(mem_vec_40, 128, 32), pl.Mem.Vec] = pl.tile.reshape(...,        [16, 1])
t__tmp_v21: pl.Tile[[8, 1], pl.FP32, pl.MemRef(mem_vec_40, 192, 32), pl.Mem.Vec] = pl.tile.reshape(...,        [16, 1])
pre0_inline10: pl.Tile[[8, 1], pl.FP32, pl.MemRef(mem_vec_51, 0, 32), pl.Mem.Vec] = pl.tile.reshape(..., [16, 1])
# ... pre1/pre2/pre3 identical
```

`memory_reuse` computes the reshape output size from the argument `[16, 1]` = **64 bytes**, but the split allocated a `[8, 1]` = **32-byte** slot → `64 > 32` → the assert (`size=64 falls outside reuse target (size 32)`). The `64` and `32` in the error map exactly to the stale `[16,1]` shape vs the split `[8,1]` type.

So `SplitVectorKernel` halves `reshape`'s result tile type but does not halve the `[T_TILE, 1]` literal it carries (because `T_TILE` is a Python constant baked into the op argument).

**Reproduction**
- File: `models/deepseek/v4/hc_pre_updown_draft.py` (the committed `hc_pre.py` with one line changed: `optimizations=[pl.split(pl.SplitMode.UP_DOWN)]` added to the `pl.spmd(...)`). Will attach.
- `python models/deepseek/v4/hc_pre_updown_draft.py -p a2a3sim --mode decode --compile-only` → aborts as above (compile-only, no device needed).
- Removing the transpose+reshape (using `pre_eps[:, h:h+1]` column slices instead, no `reshape(_, [T_TILE,1])`) makes this compile pass — confirming the `reshape` shape-argument is the trigger.

**Related**
- #839 — same class (SplitVectorKernel mishandling a `reshape` under split), but manifests as a runtime precision error rather than this compile-time memory_reuse abort. Likely the same underlying weakness in how the split rewrites `reshape`.
- #1789 — the column-slice workaround (which sidesteps *this* bug) hits a separate **runtime** 507018 fence-allocation deadlock on the same kernel.

### Environment

| Component | Version |
|---|---|
| pypto-lib | `6f0304d` (branch `perf/dsv4-hc-pre-sinkstack-mixx`; repro = `hc_pre_updown_draft.py`) |
| pypto | `a18c4cf9` (branch: main) |
| pypto runtime (submodule) | `48980572` |
| pto-isa | `8e436661` (branch: main) |
| ptoas | `0.45` |
| CANN | `9.0.0` |

### Host Platform

Linux (aarch64)

### Additional Context

- The crash is in `memory_reuse_pass` reached via `fuse_create_assemble_to_slice_pass`; the offending allocs are `mem_vec_40/51/52/53/54` (the transposed `pre_eps_t` rows and `pre0..pre3`).
- Pass dump used: enable via `run_jit(..., compile_cfg=dict(dump_passes=True))`; inspect `build_output/.../passes_dump/30_after_InitMemRef.py` (last dump before the failing pass).
- Fix direction: when `SplitVectorKernel` rescales a `reshape`'s result tile type, it must also rescale the op's target-shape argument (and any baked-in `T_TILE`-derived literal) consistently.

Little-oil:
## Summary

Fixes #1790. `SplitVectorKernel` (`pl.split` UP_DOWN / LEFT_RIGHT) assumed a single global split axis (dim0 for UP_DOWN) in every rewrite branch. When a vector scope transposes/reshapes the split (token) axis into the free axis — RMS `row_sum -> [1,N]` row vector, `tile.transpose`, sinkhorn row-major reshapes — the pass halved the wrong axis, producing numel-inconsistent tiles (an already-split `[8,1]` reshaped to a `[1,16]` claiming 16 elems) and a mis-halved transpose (`[8,8] -> [4,16]`). Those oversized views then aborted `memory_reuse`:

```
sharing-group member offset 0 size=64 falls outside reuse target (size 32)
  at src/ir/transforms/memory_reuse_pass.cpp:1461
```

The DeepSeek-V4 `hc_pre` fused cube+vec kernel hits exactly this. #1796 patched only the already-split plain result-halving reshape path and did not cover the migrating reshapes this kernel actually hits, so #1790 stayed open.

## Fix

Resolve a **per-tile split axis** before rewriting:

- A fixpoint analysis (`SplitAxisAnalysis`) seeds `tile.load` / `tpop_from_aic` at the default axis and propagates it across `transpose` (swap), `reshape` (axis map), reduce and elementwise ops.
- Input-less accumulators (`tile.full` / `tile.create` / pipeline `iter_arg`) take their axis from their consumer; `init` / `iter_arg` / `yield` are unified.
- `ProcessStmt` halves each tile on its **resolved** axis instead of the global one, dropping the full-width-reshape + per-lane-slice workaround.
- A numel-consistency `INTERNAL_CHECK_SPAN` on split view ops fails fast in this pass instead of deep in `memory_reuse`.

**Safety invariant:** a tile that never crosses a migrating transpose/reshape resolves to the default axis, so existing UP_DOWN / LEFT_RIGHT codegen is byte-identical (verified: `qwen3` scope3 post-pass IR unchanged; the 31 existing UTs unchanged).

## Tests

- New UT `test_reshape_migrates_split_axis_for_full_accumulator` (before/after structural).
- New self-contained ST `test_split_vector_kernel_axis_migration` reproducing the `memory_reuse` abort: crashes on the original pass, compiles on the fixed pass (compile-only on `a2a3sim`).
- Local: `hc_pre` repro compiles (decode + prefill); 32 split UTs pass; new ST passes.

## Validation

On-device numerical validation (real `a2a3`) is deferred to CI.

### DIFF
--- a/docs/en/dev/passes/23-split_vector_kernel.md
+++ b/docs/en/dev/passes/23-split_vector_kernel.md
@@ -117,7 +117,21 @@ reads only this attribute and never re-derives from `SplitMode`.
 3. (AIV only) InjectSubblockIdx prepends:
        subblock_idx = tile.get_subblock_idx()
    to the body, picking a fresh name if 'subblock_idx' is already taken.
-4. Walk every statement via ProcessStmt:
+4. (AIV only) SplitAxisAnalysis resolves the split axis of every tile
+   before any rewrite (phase 1). Most tiles split on `split_dim`, but a
+   tile.transpose / numel-preserving tile.reshape MIGRATES the axis between
+   dims (a per-row reduction transposed to a `[1, N]` row moves the split
+   from rows to columns), and an input-less accumulator (tile.full /
+   tile.create / pipeline iter_arg) takes its axis from a consumer, not a
+   producer. A fixpoint seeds loads / tpop_from_aic at `split_dim` and
+   propagates the axis both ways across op operand<->result relations
+   (transpose swaps it, reshape maps it, elementwise/reduce keep it,
+   iter_args unify init/arg/return); any tile left unresolved defaults to
+   `split_dim`. INVARIANT: a tile that never crosses a transpose or a
+   migrating reshape resolves to `split_dim`, so the emitted IR is
+   byte-identical to halving on the global axis.
+5. Walk every statement via ProcessStmt, halving each tile on its
+   resolved axis:
 
    tile.tpush_to_aiv / tile.tpush_to_aic / tile.tpop_from_aiv:
      RebuildCallWithSplit — sync the `split=` kwarg only. AIC keeps the
@@ -149,21 +163,23 @@ reads only this attribute and never re-derives from `SplitMode`.
      per-channel scale [D] -> [1, D] — must give each lane its own half.
      Because reshape is an offsetless view, halving only its result type
      would leave BOTH lanes reading the first half of the full buffer. So
-     when the reshape input is NOT already split and the result's split-axis
-     dim is a static non-singleton extent, emit the reshape at full width
-     and follow it with a per-subblock tile.slice selecting
-     `[..., subblock_idx * half : +half]` on the split axis (the slice
-     result, and the original var, are both tracked in tile_vars). If the
-     reshaped split-axis dim is singleton (e.g. [D] -> [1, D] under UpDown)
-     the singleton rule above keeps it full — both lanes need it. If the
-     input is already split, the reshape falls through to result-halving,
-     which also halves the explicit target-shape argument on the split axis
-     (leaving the literal stale would make memory_reuse size the output from
-     the un-split shape and abort against the split-sized slot).
+     when the reshape input is NOT split and the result's resolved-axis dim
+     is a static non-singleton extent, emit the reshape at full width and
+     follow it with a per-subblock tile.slice selecting
+     `[..., subblock_idx * half : +half]` on the resolved axis (the slice
+     result, and the original var, are both tracked in tile_vars). When the
+     input IS split, the reshape halves its result type AND its explicit
+     target-shape argument on the resolved axis — which may MIGRATE (e.g.
+     `[N, 1] -> [1, N]` moves the split from dim0 to dim1); halving the
+     stale full literal instead would make memory_reuse size the output
+     from the un-split shape and abort against the split-sized slot.
 
    any other tile.* op producing a TileType (AIV only):
-     Halve the result shape on split_dim. For tile.full / tile.create the
-     static shape arg is also halved. Reduce ops on the split axis raise
+     Halve the result shape on the tile's resolved axis. For tile.full /
+     tile.create the static shape arg is also halved — an accumulator with
+     no producer takes its axis from the consumer that pairs it with a split
+     tile (phase 1). Tiles that resolve to no axis (broadcasts both lanes
+     need) are left full. Reduce ops on the input's split axis raise
      ValueError via IsReduceOnSplitAxis (partial reduction would be
      incorrect).
 
@@ -176,11 +192,11 @@ reads only this attribute and never re-derives from `SplitMode`.
    IfStmt / SeqStmts:
      Recurse into branches and stmt sequences.
 
-5. After rewriting, transform_utils::Substitute applies var_replacements
+6. After rewriting, transform_utils::Substitute applies var_replacements
    so every reference (param, iter_arg, return_var, tpop result) sees the
    rewritten Var node.
-6. DeepClone is applied to detach from any shared IR sub-trees.
-7. WithSplitAttrs stamps the resolved SplitMode onto Function::attrs
+7. DeepClone is applied to detach from any shared IR sub-trees.
+8. WithSplitAttrs stamps the resolved SplitMode onto Function::attrs
    (overwriting any prior `split` entry). For AIV functions whose
    resolved mode is non-None it *also* writes `dual_aiv_dispatch=true`
    so orchestration codegen has a single attribute to read instead of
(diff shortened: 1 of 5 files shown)


Little-oil:
Addressed the review feedback on `kAxisNotSplit` (-2) propagation in `SplitAxisAnalysis`:

- **Generic constraint**: when no operand carries a valid split axis but one is explicitly non-split, propagate `kAxisNotSplit` to the result and operands instead of letting the result default to the split axis (fixes the result-halved-vs-inputs-full mismatch when all operands are non-split).
- **Transpose / Reshape constraints**: relaxed the `>= 0` guards to `!= kAxisUnknown` so the non-split status propagates. (Kept `sw()` on the transpose result/scratch axis — gemini's snippet predated the transpose-scratch fix.)
- **MapReshapeAxis**: early-returns for `axis < 0`, so the sentinel is never used as a shape index; also dropped the unused `from` parameter (coderabbit nit).

Validated with real ptoas: the 32 split UTs, the new ST, and hc_pre (a2a3, decode+prefill) all pass; the change is a structural no-op on already-passing kernels.

Little-oil:
Converting to draft: on-device numerical validation of the actual issue kernel (DeepSeek-V4 `hc_pre` decode, a2a3) shows `x_mixed` mismatches (~85%) **when UP_DOWN split is enabled**, while `post`/`comb` pass and the same kernel passes fully **un-split**. So the `tile.transpose` split-axis migration is numerically incorrect (post-transpose the per-lane data is column-local, but the downstream per-lane slice offset treats it as globally indexed). CI is green only because hc_pre's numerics aren't exercised on device (the in-tree ST is compile-only; swa/qwen3 transposes don't actually migrate). Holding until the transpose migration is corrected.

Little-oil:
Closing: withdrawing this axis-migration approach.

Device evidence (a2a3) shows the axis-migration IR this PR produces is actually correct (transpose result [8,8], correct per-lane slices, no spurious offset). The real defect is independent of the split pass: an [8,8] FP32 hardware transpose (validRow=8 < the pto-isa row tile of 16) only runs the TTRANS tail path and miscomputes on device, while the un-split [16,8] transpose (validRow=16, full path) is correct. A minimal pure-transpose ST (no pl.split) reproduces this: [8,8] fails, [16,8]->[8,16] passes.

So splitting a kernel that contains such a transpose cannot be made correct by the pass alone. Direction instead: have SplitVectorKernel detect that a split would produce an FP32/FP16 transpose whose source valid row count is not a multiple of the backend transpose row tile, and reject the split with a clear error (rather than silently miscompiling). Tracked under #1790.

Issue #1807, opened 2026-06-18
Feature: indexed GM-row gather with loop-local / per-iteration indices (paged_gather #1644 is single-shot only)


lwDavid:
## Summary

We need an indexed GM-row gather (gather scattered rows of a paged KV pool into UB/L1) that can be driven by **loop-local / per-iteration indices** — usable inside an `spmd` or `pl.range` / `CORE_GROUP` loop where each iteration gathers a different set of rows. The recently merged `pl.paged_gather` (#1644) is exactly the right primitive conceptually, but it is **single-shot only**: its `indices` must be a kernel-input (or otherwise tile-lowering-surviving) tensor used in one call with no enclosing runtime loop. That does not fit per-token sparse attention.

## Motivation

In the DeepSeek-V4 unified decode sparse-attention kernel (`models/deepseek/v4/decode_sparse_attn.py`), the compressed-KV gather is ~half the decode wall. Per token it does a paged lookup `phys = block_table[idx // block_size] * block_size + idx % block_size` and gathers the selected rows — byte-for-byte what `paged_gather` does internally. Fusing it (gather straight into L1, feed the QK/PV matmul, drop the ~84 MB `sparse_kv` GM round-trip) is the single biggest remaining decode optimization.

## What we tried (all rejected)

Calling `pl.paged_gather` with a per-(token, block) index, three index forms:

1. a runtime slice `cmp_logical_idx[g_t:g_t+1, off:off+128]` inside an AIV `pl.spmd`
2. the same slice inside `pl.at(level=pl.Level.CORE_GROUP)` + `pl.range`
3. a freshly **materialized** index — `pl.create_tensor([1, 128], INT32)` written from the slice — inside the CORE_GROUP loop

All three fail at lowering with:

```
pypto.InternalError: tensor.paged_gather: indices must be TensorType
Check failed: idx_type at src/ir/transforms/op_conversion_registry.cpp:1482
```

## Root cause

`paged_gather`'s lowering (op_conversion) runs **after** the tile-lowering passes, so any tensor created or sliced inside a runtime loop has already become a tile / `partition_view` by then and is no longer a `TensorType`. Only tensors that survive tile-lowering (kernel inputs, or top-level tensors used whole with no per-iteration indexing) stay `TensorType`. The ST test (`tests/st/runtime/ops/test_paged_gather.py`) passes precisely because it calls `paged_gather` once, outside any loop, on a kernel-input `idx`.

## Request

Either of:

[...]

PR #1812, opened 2026-06-18
feat(op): add tile.mgather (GM-row indexed gather-load into VEC tile) [#1807]

Little-oil:
## Summary

Adds `tile.mgather` (`pto.mgather`) — indexed gather-load from a GM table into a fresh VEC tile, usable with a loop-local index inside `spmd` / `pl.range` (route B for #1807). Two modes via `coalesce`:

- **row** (default): `dst[r, j] = mem[idx[r], j]`, `idx` a `[1, R]` index vector, output `[R, mem_cols]`.
- **elem**: `dst[i, j] = mem[idx[i, j]]`, output = `idx` shape.

Implemented across all layers: C++ op (`src/ir/op/tile_ops/memory.cpp`), IR wrapper, DSL (`pl.tile.mgather` / `pl.mgather`), codegen (`MakeTileMgatherCodegenPTO`), UT, and en/zh docs.

## Key fix — row-mode 507018 hang (`not_inplace_safe`)

`tile.mgather` initially defaulted to in-place-safe, so `MemoryReuse` could place the output tile at the **same UB address as the index tile** (both at offset 0). Row-mode gather reads the whole `idx` vector across its per-row DMA loop while writing `dst`, so an aliased `dst` corrupts indices mid-loop → garbage GM addresses → intermittent AICPU **507018** hangs that worsen with gathered row count (16 rows clean, 32 rows ~always hung).

Fix: mark the op `.not_inplace_safe()` (mirroring `tile.gather` / `tile.scatter`) so `dst` gets its own buffer. Verified on a2a3 (pto-isa e722679b + PTOAS 0.47): the case that deterministically hung (32×32) now passes 5/5 with zero hangs across repeated runs; index/dst tiles allocate at disjoint UB offsets.

## Also included

- Derive the output tile's `valid_shape` from the index tile's effective valid (`DeduceTileMgatherType`), with UTs.
- Row regression ST cases (`16×64`, `32×32`) that reproduced the hang before the fix.
- Doc note: PTOAS requires the index tile row byte size to be 32-byte aligned (`none_box` constraint) — INT32 index width a multiple of 8; pad + `valid_shapes` for other counts.

## a2a3 ST is skipped pending a toolchain bump

`TestMgatherRow` / `TestMgatherElem` are `@pytest.mark.skip`. a2a3 `tile.mgather` needs a toolchain newer than CI currently pins (ptoas v0.45, pto-isa 016396b5 via runtime 09f1546f):

- **PTOAS 0.47** for the row-mode `MTE2→S` index sync (PTOAS#861) — not yet officially released.
- a **pto-isa with the templated `MGATHER<Coalesce>`** (e722679b) and its **matching runtime** (SDMA layout). These move in lockstep via `runtime/pto_isa.pin`; overriding the pto-isa alone regresses the distributed SDMA path, so it can't be done piecemeal.

The op and the in-place fix are verified locally on that toolchain (row + elem pass). Un-skip once CI's ptoas / pto-isa / runtime are bumped together.

Refs #1807.


Little-oil:
mgather的同步插入有Bug，等pto-isa删掉手动插入的同步，等PTOAS正确插入同步

PR #1815, opened 2026-06-19
feat(ir): M/N output tiling in AutoTileMatmulL0

tonibohnlein:
## Summary

`AutoTileMatmulL0` lowered an oversized `tile.matmul` to a pipelined K-loop but only handled **K-tiling**. When `ChooseL0Tile` returned `m < M` or `n < N` (the `[M, N]` output Acc exceeds L0c) it emitted `PH-AT-006` and left the matmul untouched — so the full `[M, N]` accumulator overflowed L0c at `AllocateMemoryAddr`.

This PR implements **M/N output tiling**. The operands are already Mat-resident, so only the output overflows. For a plain `tile.matmul` whose result is consumed by exactly one 2D `tile.store(c, base, out)`, the pass now tiles the output:

- unrolls into a `ceil(M/m) × ceil(N/n)` grid of `[m, n]` sub-tiles (partial on the boundary via **static** extents — `m`/`n` need not divide `M`/`N`),
- computes each with the existing pipelined K-loop,
- stores each `[m, n]` Acc sub-tile straight to `out[mi:, ni:]` (the direct-store / DDR-output path),
- chains the output tensor through the per-sub-tile stores in SSA form.

Every Acc tile is then ≤ L0c, so the matmul lowers through `AllocateMemoryAddr` with no overflow.

## Why

A cross-core matmul scheduler emits per-tile `[w, h]` output regions whose per-tile InCore kernels must fit L0c for any tile size. We deliberately do not bound that scheduler by L0c (it would distort the L1 + cross-core region balance), so on-core M/N tiling here is the mechanism that lets oversized output tiles lower.

## Design notes

- **Unrolled static sub-tiles, not a runtime loop.** `ChooseL0Tile` optimizes L1↔L0 traffic and picks `m`/`n` that rarely divide `M`/`N` (e.g. 256×256 → `m=192, n=160`), so partial boundary tiles are the norm. Unrolling with static partial extents handles boundaries cleanly — no `valid_shape` masking, no runtime `min()` — and is trivially correct.
- The K-loop builder is parameterized by sub-tile offset `(mi, ni)` + extent `(m, n)`; the **K-only path (`mi = ni = 0`, `m = M`, `n = N`) emits byte-identical IR** (existing structural tests unchanged).
- A per-`SeqStmts` use-count + store-consumer index is built lazily so M/N folding stays **O(N)**.

## Deferred (still `PH-AT-006`, left untouched)

`tile.matmul_acc` (needs per-sub-tile slicing of the caller's `[M, N]` accumulator), a Vec left operand (PV path), a single K block (`K/k < 2`), and non-store consumers (need the Mat-scratch / `tile.assemble` path for on-chip downstream ops).

## Testing

- **End-to-end acceptance** — 256×256 FP32 on Ascend910B (output 256 KB > 128 KB L0c) compiles through the full pipeline (passes `AllocateMemoryAddr` with no L0c overflow) and matches `torch.matmul` (`rtol=atol=1e-3`).
- Pass-level numeric checks via `torch_codegen` for clean (512×512) and **partial-boundary** (384×384) tiles.
- Structural before/after for the 2×2 grid; deferred-case and idempotency coverage.
- Full suite green (`tests/ut/ir/transforms`, `tests/ut/jit`, `tests/ut/debug`: 2025 passed).

A `tile.extract` handler was added to `debug.torch_codegen` (a 2D slice) so tiled matmuls — K-only and M/N — can be numerically validated; the op was previously never exercised through `torch_codegen`.

Docs updated (`docs/en` + `docs/zh-cn`).

## Note: separate prerequisite commit

`fix(backend): import Traversable from importlib.resources.abc on Python 3.11+` — `importlib.abc.Traversable` was deprecated in 3.12 and removed in 3.14, breaking all pypto imports on a 3.14 interpreter. Required for the test suite to run; kept as its own focused commit.

### DIFF
--- a/docs/en/dev/passes/16-auto_tile_matmul_l0.md
+++ b/docs/en/dev/passes/16-auto_tile_matmul_l0.md
@@ -1,10 +1,12 @@
 # AutoTileMatmulL0 Pass
 
-L0 tiling for `tile.matmul` / `tile.matmul_acc` ops with a Mat right operand (and a Mat- or Vec-resident left operand): pick an L0 tile shape `(m, n, k)` from the active backend's L0 capacities and rewrite the call into a 2-stage pipelined K-loop with per-iter Mat→Left/Right `tile.extract`s.
+L0 tiling for `tile.matmul` / `tile.matmul_acc` ops with a Mat right operand (and a Mat- or Vec-resident left operand): pick an L0 tile shape `(m, n, k)` from the active backend's L0 capacities and rewrite the call into a 2-stage pipelined K-loop with per-iter Mat→Left/Right `tile.extract`s. When the `[M, N]` output itself exceeds L0c, the output is additionally tiled (M/N tiling) into a grid of `[m, n]` sub-tiles, each stored straight to the output tensor.
 
 ## Overview
 
 Mat-resident matmuls produced upstream by `ConvertTensorToTileOps` + [`FlattenTileNdTo2D`](15-flatten_tile_nd_to_2d.md) carry full `(M, N, K)` operand shapes — almost always larger than the cube unit's L0a/L0b/L0c capacity. This pass picks an L0-fitting `(m, n, k)` and rewrites the matmul into a K-loop whose body extracts `[m, k]` and `[k, n]` slabs into `Left` / `Right` and accumulates into an `Acc`-resident iter-arg. The loop is marked `ForKind::Pipeline` with `pipeline_stages=2` so the downstream [`LowerPipelineLoops`](26-lower_pipeline_loops.md) pass produces a 2-deep ping-pong on the per-iter operand extracts.
+
+**K-tiling vs M/N-tiling.** When the chooser returns `m == M` and `n == N` the output already fits L0c, so only the K dimension is tiled (one K-loop). When it returns `m < M` or `n < N` the `[M, N]` output Acc would overflow L0c. The operands are already Mat-resident, so *only* the output overflows: the pass tiles the **output** into a `ceil(M/m) × ceil(N/n)` grid of `[m, n]` sub-tiles (partial on the boundary — `m`/`n` need not divide `M`/`N`), computes each with the same pipelined K-loop, and stores each `[m, n]` Acc sub-tile straight to `out[mi:, ni:]` (the direct-store / DDR-output path). Every Acc tile is then ≤ L0c, so the matmul lowers through `AllocateMemoryAddr` with no overflow. The output tensor is chained through the per-sub-tile stores in SSA form (`out → out_t0 → out_t1 → …`).
 
 **Pipeline position**: After [`FlattenTileNdTo2D`](15-flatten_tile_nd_to_2d.md), before [`InferTileMemorySpace`](18-infer_tile_memory_space.md). All tile ops are already 2D and memory spaces have not yet been inferred.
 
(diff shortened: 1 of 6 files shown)


tonibohnlein:
@Hzfengsy this PR implements the M/N output-tiling TODO you left in `AutoTileMatmulL0` (`auto_tile_matmul_l0_pass.cpp`, 86e9d76dd — "M/N tiling needs a Mat-resident output scratch + per-iter assemble that is not yet implemented").

It takes the direct-store route rather than the Mat-scratch one: for a plain `tile.matmul` whose oversized `[M, N]` output is consumed by a single 2D `tile.store`, the output is unrolled into a `ceil(M/m) × ceil(N/n)` grid of sub-tiles (static partial extents on the boundary), each running the existing pipelined K-loop and storing straight to `out[mi:, ni:]`. The Mat-scratch + `tile.assemble` path (for on-chip / chained-matmul consumers) plus `matmul_acc`, Vec-lhs, and single-K-block M/N are deferred and still emit `PH-AT-006`.

Would appreciate your review since it's your TODO — happy to adjust the approach.

tonibohnlein:
Thanks for the reviews. Addressed the `gemini-code-assist` feedback in c6372b16: both internal invariants now use `INTERNAL_CHECK_SPAN` so failures carry the IR source location.

One correction on the suggested snippets — the macro is `INTERNAL_CHECK_SPAN(expr, span)` (condition first, span second), so the applied form is:
- `INTERNAL_CHECK_SPAN(store_call, store_stmt->span_)`
- `INTERNAL_CHECK_SPAN(tiling->K / tiling->k >= 2, tiling->assign->span_)`

Both span sources (`store_stmt`, `tiling->assign`) are guaranteed non-null at the failure point, so this is safe per the span-safety rule in the project's error-handling conventions.

Re: the CodeRabbit docstring-coverage warning — the uncovered functions are the `@pl.program` `Before`/`Expected` `kernel` fixtures in the test file, which are intentionally undocumented (consistent with the existing tests in this file); the test methods themselves are all docstring'd.

tonibohnlein:
Thanks for the thorough re-review — the SSA bug is real and is now fixed in 97234d1b.

**Reproduced exactly as described.** With the reversed def/store ordering, the printer flagged the dangling reference:
```
out1_t0 = tile.store(c2_t0, [0, 512], out1__FREE_VAR)   # before the fix
```
fold-2 (for `c2`) captured the raw `out1` before fold-1's rewrite of `out1` existed, and `out1`'s defining store was then dropped.

**Fix.** I went with the "apply the current remap at the consumer-store emission site" approach. One nuance that bit the first attempt — exactly your *"carefully order the original store-result mapping"* caveat: for a single matmul, `out = tile.store(c, …, out)` the store's **input tensor and result are the same SSA var** (output-param aliasing), so `remap[store_result] = fold_final` aliases the chain start, and a blanket `Substitute` rewrites the fold's own first store onto its final output (a self-referential use-before-def — the after-pass verifier caught it). So the emission applies the remap **excluding the fold's own `store_result_var`**:
```cpp
auto self = remap.extract(it->second.store_result_var.get());
for (auto& s : it->second.stmts)
  out.push_back(remap.empty() ? s : transform_utils::Substitute(s, remap));
if (!self.empty()) remap.insert(std::move(self));  // restore for downstream uses
```
This rewrites the chain start to the producing fold's output in the reverse-order case (input ≠ result) while leaving the output-param chain start intact.

**Regression test** added (`test_mn_tiling_reversed_def_store_chain_stays_ssa`): the exact reversed def/store ordering, two oversized matmuls writing disjoint halves of a `[512, 1024]` output; it runs `verify_properties(SSAForm + UseAfterDef)` after the pass and checks numerical correctness via `torch_codegen`. (The pass also runs under the after-pass verifier instrument in these tests, so SSA is checked automatically.)

**On scope** — fully agree this is the **direct-to-store subset**, not the general Mat-scratch TODO. I've added an explicit `TODO(M/N tiling)` in the file header documenting the remaining work: the Acc→Mat `tile.assemble` (→ `pto.tinsert`) path for on-chip / non-store consumers, plus Mat-capacity checking and per-sub-tile accumulator slicing for `tile.matmul_acc`. Those cases emit `PH-AT-006` today. Your `pto.tinsert` / `tile.insert` design notes are excellent and I'll use them as the basis for the follow-up PR rather than expanding this one. Happy to adjust the PR title to make the "direct-store subset" framing explicit if you'd prefer.

tonibohnlein:
Closing in favor of the general M/N solution. This PR implemented the direct-to-store specialization; per review we'll deliver the full TODO — a Mat-resident output scratch + per-output-tile assembly (Acc→Mat insertion via pto.tinsert), with the direct-store path retained for single-store consumers under a consumer-aware dispatcher, plus matmul_acc per-sub-tile accumulator slicing, the single-K-block path, Mat-capacity checking, and the SSA reverse-order regression fix. A new PR will supersede this one.

tonibohnlein:
Reopening as the **direct-GM (terminal-output) M/N slice** — a deliberately narrow first step, not the general M/N TODO.

**What it covers:** an oversized `tile.matmul` whose `[M,N]` output exceeds L0c and is consumed by a single 2D `tile.store` is tiled into `ceil(M/m)×ceil(N/n)` sub-tiles, each completing its full K-reduction in one L0C accumulator and storing **directly to the final GM coordinates** (each output element written exactly once — no scratch, no whole-tensor copy). Partial M/N boundaries use static extents; the SSA reverse-order fold bug is fixed (`97234d1b`) with a regression covering reversed def/store ordering + SSA/UseAfterDef + numerics.

**Scope boundary (intentional):** successful M/N tiling is *always* K-split (`K/k ≥ 2`); the **full-K single-block** form (`k == K`, `m<M`/`n<N`) is deferred (`PH-AT-006`), as are `matmul_acc`, Vec-lhs, and non-store consumers. One L0C accumulator (no L0C double-buffer) is the intended default.

**Follow-up roadmap (separate PRs):**
1. **Full-K single-block M/N path** — straight-line `extract [m,K]/[K,n] → matmul → store` (no K-loop); the natural home for A/B-stationary + serpentine operand reuse. Self-contained, no allocator dependency.
2. **Schedule-aware chooser** — make `k` a real optimization variable (current traffic model ignores `k`); add `stationary_operand`/traversal + a calibrated cost model.
3. **Split-K accumulator strips** — A/B-stationary reuse bounded by live L0C slots.
4. **Acc→Mat / L1 materialization** for on-chip consumers — empirically `tile.assemble` already does Acc→Mat correctly; the real gate is a **MemoryReuse peak-aware larger-reuses-smaller** fix (the naive relaxation regresses qwen3 `rmsnorm`).
5. Conditional (not unconditional) L0C double-buffering.

This PR should be reviewed strictly within its documented subset.

Hzfengsy:
@codex review

tonibohnlein:
The earlier CI failure was `pre-commit` only (everything else — unit-tests 3.10, clang-tidy, codegen/system/dist/a5sim — passed). Fixed in `2a7f501e`: clang-format wrapping, ruff import-sort (I001) + `PLR0915`/`PLC0415` noqa, and a pyright fix (dropped `out_dtype` on a tile-operand `pl.matmul` so it resolves the Tile overload). Verified locally — `ruff`, `clang-format`, and `pyright` all pass, and the 21 unit tests stay green.

The re-run shows **`action_required`** (fork-PR workflow approval). @Hzfengsy — could you approve the workflow run when you get a chance? Thanks.

Hzfengsy (approved):

Issue #1697, opened 2026-06-08
[Bug] Orchestration codegen does not hoist a manual_scope-internal tensor's SSA alias, breaking after-scope readers ("was not declared in this scope")


Hzfengsy:
### Component

Codegen

### Description

The orchestration C++ codegen emits a tensor's SSA-versioned `const Tensor&` alias at the **current indentation** — i.e. inside the deep `PTO2_SCOPE(MANUAL) { ... }` block when the producing write is inside a `pl.manual_scope` — with **no live-out / hoisting analysis**. When a task placed **after** the `manual_scope` reads that tensor, the generated `add_input(<name>__ssa_v<N>)` is emitted at the outer indent and references the now-out-of-scope alias, so the orchestration `.cpp` fails to C++-compile:

```
error: '<name>__ssa_v<N>' was not declared in this scope
```

This blocks any optimization that moves a consumer of a `manual_scope` `pl.assemble` output to after the scope (e.g. fusing an MLP-tail residual-add / `down_cast_residual`-style writer into an after-scope auto-dep writer).

Atomic/inout tensors are unaffected: their producer-call alias is bound in the outer function body (atomic-add does not mint a deep downstream SSA version), so the after-scope reader resolves to an outer alias. A plain (non-atomic) `pl.assemble` inside the scope mints a new SSA version aliased at deep indent, which dies when the block closes.

**Why it is easy to miss:** `compile_for_test` / `--smoke` generate the orchestration `.cpp` but do **not** C++-compile it, so they pass. The error only surfaces when the `.cpp` is C++-compiled — i.e. a real device dispatch / orchestration C++ build.

Related: #1693 (orchestration codegen `__ssa_v` alias cross-assignment).

### Steps to Reproduce

Self-contained minimal repro (`repro_manual_scope_hoist.py`): a tensor created **before** `pl.manual_scope`, written **inside** it, read by a `pl.at` task **after** it.

```python
import torch
import pypto.language as pl
from pypto.backend import BackendType, set_backend_type
from pypto.runtime import RunConfig

M, N = 16, 256

@pl.jit
def repro(x: pl.Tensor, out: pl.Out[pl.Tensor]):
    acc = pl.create_tensor([M, N], dtype=pl.FP32)  # atomic-written  (works)
    buf = pl.create_tensor([M, N], dtype=pl.FP32)  # plain-assemble-written (breaks)
[...]

PR #1828, opened 2026-06-22
refactor(codegen): use CodeEmitter in orchestration + add precondition safety net

georgebisbas:
## Summary

Migrate `OrchestrationStmtCodegen` from raw `std::ostringstream code_` + `int indent_` to the shared `CodeEmitter`, consolidate four duplicated task-dispatch sequences into a single `TaskDispatchPlan::Emit()`, add a codegen-entry precondition safety net, and extract GM-pipe workspace analysis into its own module.

**Branch:** `refactor/orch-use-codeemitter`  
**HEAD:** `77fd678d`  
**Rebased on:** `origin/main`  
**Files:** 13 changed (+909 / -408)  
**Sim Docker:** 527 codegen UT passed (incl. 3 emission-order regression tests). All 163 pre-existing orchestration codegen blobs verified byte-for-byte identical to `origin/main`.

---

### 1. Replace `ostringstream` + `indent_` with `CodeEmitter`

`OrchestrationStmtCodegen` was the last codegen class with its own raw emission machinery. Every line was:

```cpp
code_ << Indent() << "..." << "\n";
```

And every block boundary required matching `indent_ += 4` / `indent_ -= 4`. This is trivially unbalanced on early returns, and there was no guard against negative indent — the old code would silently produce malformed C++.

The migration replaces this with:

- **`EmitIndentedLine(text)`** — emits one indented line; no explicit newline or indent string needed
- **`IndentGuard guard(Active())`** — RAII; the destructor always fires, and `CodeEmitter::DecreaseIndent()` has an `INTERNAL_CHECK` that catches underflow
- **`Active()` / `active_emitter_`** — pointer to the current emission target; used by `EmitBatchedAllocTensors` to swap in a temporary emitter for the scope-hoist path without disrupting the main stream

The `scope_hoist_indent_` field was a `std::string` that captured the indent string at scope-entry time. This is now an `int scope_hoist_indent_level_` — the string representation is computed once via `IndentAtLevel()` where needed, and the level value supports identity-safe comparison in `IsAtManualScopeBodyIndent()`.

### 2. `TaskDispatchPlan` — consolidate four dispatch sequences

`GenerateDirectCallCode`, `GenerateSpmdCallCode`, `GenerateGroupCallCode` (AIV-only), and `GenerateGroupCallCode` (Mixed) each contained the same ~15-line emission block:

```cpp
std::string ind = Indent();
std::string task_var = "params_t" + ...;
code_ << "\n";
code_ << ind << "// ...";
EmitTaskParamsDecl(ind, task_var);
for (const auto& p : params) { code_ << ...; }
EmitSelectiveDumpCall(ind, task_var, params);
// ... ctx scalars, launch spec, manual deps, submit
```

These are now built via builder functions (`BuildDirectCallDispatchPlan`, `BuildSpmdCallDispatchPlan`, `BuildAivOnlyGroupDispatchPlan`, `BuildMixedGroupDispatchPlan`) that all return a `TaskDispatchPlan` value. A single `Emit(cg)` method emits the sequence, preserving the historical per-path order exactly:

| Path | Order |
|------|-------|
| Direct call | `params → dump → ctx scalars → deps → launch_spec → early_resolve → submit` |
| Spmd / Group / Mixed | `params → dump → ctx scalars → pre_lines (MixedKernels) → launch_spec → early_resolve → deps → submit` |

The difference between the two paths is encoded in a `deps_before_launch` flag on `TaskDispatchPlan`, set only by `BuildDirectCallDispatchPlan`. This is the minimal expression of the two orderings that existed separately in the four original functions — no new logic is introduced.

The builder functions structurally enforce the ctx-scalar dispatch:

| Path | `dist_ctx_callee` | `ctx_scalar_names` | Emitted |
|------|-------------------|-------------------|---------| 
| Direct call | `callee_func` | `{}` | `EmitDistTensorCtxScalars` (walks callee params, 1:1 safe) |
| Spmd/Group/Mixed | `nullptr` | from `BuildWrapperReorderedParams` | `EmitCtxScalars` (uses pre-collected inner-kernel names) |

### 3. `WrapperParams` / `EmitCtxScalars` — two ctx-scalar paths

Wrapper functions (Spmd, Group) reorder arguments before reaching the inner kernel, so the inner callee's `DistributedTensor` param positions don't map 1:1 to the wrapper's args. `BuildWrapperReorderedParams` walks the inner callee and collects `ctx_scalar_names` directly from the `DistributedTensor` formals. `EmitCtxScalars` emits `task_var.add_scalar(ext_<name>_ctx)` per entry — the wrapper-path counterpart of `EmitDistTensorCtxScalars` (which walks a directly-called callee's own params at emit time).

### 4. Codegen-entry precondition barrier

Pass-level property verifiers (from #1700) fire *after* a pass runs. Nothing previously checked invariants *before codegen entry*. If a pass is omitted from a strategy, codegen silently produces invalid output.

This PR adds two precondition checks, each called as the very first thing in the codegen entry point:

| Entry point | Checks |
|---|---|
| `GenerateOrchestration` | `SplitIncoreOrch`, `OrchestrationReferencesResolved`, `RuntimeScopesMaterialized` |
| `DistributedCodegen::Generate` | `CommDomainScopesMaterialized` (gated: only when distributed tensors + windowing ops are present) |

Both call `pass::VerifyProperties`, which raises `pypto::InternalError` — same error type as #1700's post-pass verifiers, with the message pattern `"Internal error: <property> not established — pass <N> should have <done>. The pass pipeline is incomplete."` — immediately actionable for anyone debugging a strategy configuration.

### 5. Extract GM-pipe workspace analysis

`ComputeGMPipeWorkspaceElements` and `GetGMPipeSlotCount` were anonymous-namespace helpers in the 3800-line `orchestration_codegen.cpp`. They are now in `orchestration_analysis.cpp` / `.h`, alongside the other analysis visitors. The extraction is byte-for-byte identical; no logic changed.

### 6. Replace fragile inline string concatenation with loop helpers

Hot-path loop emissions (array-carry init/yield copies, compiler-dep collection fills, non-trivial parallel range headers) were repeated inline string concatenations at 6+ call sites. Now replaced with single-call helpers (`EmitForLoopHeader`, `EmitArrayCopyLoop`, `EmitArrayFillLoop`) that produce identical output but are auditable in one place.

---

## Review fixes applied

- **`SetInitialIndent` missing guard** — added `INTERNAL_CHECK(indent >= 0 && indent % 4 == 0)` before the `/ 4` division; without it, `indent = -1` truncated to `0` and silently passed
- **Hoisted carry decl missing `\n`** — `scope_hoist_sink_->push_back(...)` entry was missing the `\n` terminator; the manual-scope flush loop uses `AppendRaw` (no implicit newline), so each entry must carry its own; this caused the hoisted decl to concatenate with the `PTO2_SCOPE(MANUAL)` header, producing uncompilable code
- **Per-path `deps` / `launch_spec` emission order** — `main` used two different orderings across its four per-path functions: direct calls emitted `set_dependencies` before `launch_spec`; wrapper paths (Spmd/Group/Mixed) emitted it after. The initial unified `Emit()` accidentally applied the wrapper order to all paths, reordering one line vs `main`. Restored via the `deps_before_launch` flag on `TaskDispatchPlan`. Verified by running a before/after codegen dump across the full suite: 163/163 pre-existing blobs are now byte-for-byte identical to `origin/main`. Three ordering tests lock this against future regression.
- **`codegen_preconditions.cpp` used `CHECK`** — a failed precondition is a compiler/strategy bug, not user input; changed to `INTERNAL_CHECK` per `error-checking.md`
- **`WrapperParams`/`EmitCtxScalars` accidentally omitted** — the initial refactoring commit dropped `WrapperParams` and `EmitCtxScalars`; restored from PR #1917. Without these, wrapper-path `DistributedTensor` CommContext scalars were not emitted for Spmd/Group/Mixed task launches

---

## Testing

- **`tests/ut/codegen/test_codegen_preconditions.py`** (new) — 2 tests: distributed precondition fires when `CommDomainScopesMaterialized` is deliberately omitted; orchestration precondition entry point is wired correctly
- **`tests/ut/codegen/test_orchestration_codegen_more.py`** — 2 new emission-order tests: `test_mixed_group_emission_order` (asserts `MixedKernels` before `launch_spec`), `test_direct_call_with_deps_emission_order` (asserts `set_dependencies` before submit)
- **`tests/ut/codegen/test_orchestration_task_deps.py`** — added position assertion to `test_spmd_submit_emits_launch_spec_and_captures_task_id`: verifies that the direct-call path emits `set_dependencies` before `launch_spec`, locking the per-path byte-identical order
- **Byte-identity verification** — ad-hoc before/after codegen dump across the full codegen suite: 163/163 pre-existing orchestration blobs match `origin/main` byte-for-byte. The only textual variance across the full run is order-dependent SSA temp numbering (a process-global counter in the inlining pass, not in codegen), which is fully deterministic for any single input program

### DIFF
--- a/src/codegen/codegen_preconditions.cpp
+++ b/src/codegen/codegen_preconditions.cpp
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) PyPTO Contributors.
+ * This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+ * CANN Open Software License Agreement Version 2.0 (the "License").
+ * Please refer to the License for details. You may not use this file except in compliance with the License.
+ * THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+ * See LICENSE in the root of the software repository for the full text of the License.
+ * -----------------------------------------------------------------------------------------------------------
+ */
+
+#include "pypto/codegen/codegen_preconditions.h"
+
+#include <string>
+
+#include "pypto/ir/transforms/base/visitor.h"
+#include "pypto/ir/transforms/ir_property.h"
+#include "pypto/ir/transforms/passes.h"
+#include "pypto/ir/type.h"
+
+namespace pypto {
+namespace codegen {
+
+using namespace pypto::ir;  // NOLINT(build/namespaces)
+
+namespace {
+
+class DistributedTensorUseCollector : public IRVisitor {
+ public:
+  bool uses_distributed_tensor = false;
+  bool has_comm_domain_scope = false;
+  bool has_windowing_ops = false;
+
+ protected:
+  void VisitStmt_(const CommDomainScopeStmtPtr& op) override {
+    has_comm_domain_scope = true;
+    uses_distributed_tensor = true;
+    IRVisitor::VisitStmt_(op);
+  }
+
+  void VisitExpr_(const VarPtr& var) override {
+    if (As<DistributedTensorType>(var->GetType())) {
+      uses_distributed_tensor = true;
+    }
+    IRVisitor::VisitExpr_(var);
+  }
+
+  void VisitExpr_(const CallPtr& call) override {
+    const std::string& op_name = call->op_->name_;
+    if (op_name.rfind("pld.tensor.", 0) == 0 || op_name.rfind("tensor.window", 0) == 0 ||
+        op_name.rfind("tensor.alloc_window_buffer", 0) == 0 ||
+        op_name.rfind("tensor.window_buffer", 0) == 0 || As<DistributedTensorType>(call->GetType())) {
+      uses_distributed_tensor = true;
+      has_windowing_ops = true;
+    }
+    IRVisitor::VisitExpr_(call);
+  }
+};
+
+}  // namespace
+
+void VerifyOrchestrationCodegenPreconditions(const ProgramPtr& program, const FunctionPtr& func) {
+  INTERNAL_CHECK(program != nullptr)
+      << "Internal error: GenerateOrchestration preconditions — program must not be null";
+  INTERNAL_CHECK(func != nullptr)
+      << "Internal error: GenerateOrchestration preconditions — function must not be null";
+
+  // Codegen assumes hierarchy references resolved and explicit
+  // RuntimeScopeStmt materialization.
+  pass::VerifyProperties(
+      IRPropertySet{IRProperty::SplitIncoreOrch, IRProperty::OrchestrationReferencesResolved,
+                    IRProperty::RuntimeScopesMaterialized},
+      program, "GenerateOrchestration preconditions");
+}
+
+void VerifyDistributedCodegenPreconditions(const ProgramPtr& program) {
+  INTERNAL_CHECK(program != nullptr)
+      << "Internal error: DistributedCodegen preconditions — program must not be null";
+
+  DistributedTensorUseCollector collector;
+  collector.VisitProgram(program);
+  if (!collector.uses_distributed_tensor) {
+    return;
+  }
+
+  INTERNAL_CHECK(!collector.has_windowing_ops || collector.has_comm_domain_scope)
+      << "Internal error: DistributedCodegen preconditions — MaterializeCommDomainScopes must run before "
+         "DistributedCodegen when window-buffer/distributed-tensor ops are present. "
+         "The pass pipeline is incomplete.";
+
+  // Comm-domain materialization is required when DistributedTensor values are
+  // present in host orchestration paths.
+  pass::VerifyProperties(IRPropertySet{IRProperty::CommDomainScopesMaterialized}, program,
+                         "DistributedCodegen preconditions");
+}
+
+}  // namespace codegen
+}  // namespace pypto
--- a/src/codegen/orchestration/orchestration_codegen.cpp
+++ b/src/codegen/orchestration/orchestration_codegen.cpp
@@ -1929,11 +1929,9 @@ class OrchestrationStmtCodegen : public CodegenBase {
   /// Wrapper functions may omit constants or otherwise expose a different
   /// parameter order than the callee binary expects. Submit args using the
   /// inner callee's order, not the wrapper's order.
-  WrapperParams BuildWrapperReorderedParams(const CallPtr& outer_call,
-                                                      const FunctionPtr& wrapper_func,
-                                                      const CallPtr& inner_call,
-                                                      const FunctionPtr& inner_callee,
-                                                      const WrapperBridge& bridge = {}) {
+  WrapperParams BuildWrapperReorderedParams(const CallPtr& outer_call, const FunctionPtr& wrapper_func,
+                                            const CallPtr& inner_call, const FunctionPtr& inner_callee,
+                                            const WrapperBridge& bridge = {}) {
     std::unordered_map<const Var*, size_t> wrapper_param_to_outer_idx;
     for (size_t i = 0; i < wrapper_func->params_.size(); ++i) {
       wrapper_param_to_outer_idx[wrapper_func->params_[i].get()] = i;
@@ -2355,8 +2353,7 @@ class OrchestrationStmtCodegen : public CodegenBase {
   /// ``BuildWrapperReorderedParams`` (inner-kernel param order); this appends the
   /// ``_ctx`` suffix. The wrapper-path counterpart of ``EmitDistTensorCtxScalars``
   /// (which walks a directly-called callee's own params).
-  void EmitCtxScalars(const std::string& task_var,
-                      const std::vector<std::string>& ext_names) {
+  void EmitCtxScalars(const std::string& task_var, const std::vector<std::string>& ext_names) {
     for (const auto& ext_name : ext_names) {
       EmitIndentedLine(task_var + ".add_scalar(" + ext_name + "_ctx);");
     }
@@ -2452,8 +2449,7 @@ class OrchestrationStmtCodegen : public CodegenBase {
 
   TaskDispatchPlan BuildAivOnlyGroupDispatchPlan(const CallPtr& call, const FunctionPtr& launch_func,
                                                  const std::string& group_name, int aiv_id,
-                                                 std::vector<ParamEntry>&& params,
-                                                 bool capture_plain_task_id,
+                                                 std::vector<ParamEntry>&& params, bool capture_plain_task_id,
                                                  std::vector<std::string>&& ctx_scalar_names = {}) {
     std::string task_var = CurrentTaskVarName();
     auto [launch_core_num, launch_sync_start] = EffectiveLaunchSpec(call, launch_func);
@@ -2738,10 +2734,10 @@ class OrchestrationStmtCodegen : public CodegenBase {
       std::string emit_var = ReserveVarEmitName(assign->var_.get());
       if (IsInjectedGMPipeCreateVar(assign->var_)) {
         auto create_use = ResolveGMPipeCreateUse(stmts, stmt_idx, assign->var_);
-        CHECK(create_use.has_value())
+        INTERNAL_CHECK(create_use.has_value())
             << "Internal error: injected gm_pipe_buffer tensor.create is not passed to a callee";
         int64_t workspace_elems = GetGMPipeWorkspaceElements(create_use->callee);
-        CHECK(workspace_elems > 0)
+        INTERNAL_CHECK(workspace_elems > 0)
             << "Internal error: injected gm_pipe_buffer tensor.create found without initialize_pipe ops";
         std::string size_expr = std::to_string(workspace_elems);
         const std::string& core_num_expr = create_use->core_num_expr;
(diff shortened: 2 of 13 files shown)


src/codegen/orchestration/orchestration_codegen.cpp
```diff
@@ -3011,13 +3020,15 @@ class OrchestrationStmtCodegen : public CodegenBase {
   /// Caller guarantees the decl type is ``Tensor``.
   void EmitMutableTensorCarryDecl(const std::string& name, const std::string& init_expr) {
     if (scope_hoist_sink_ != nullptr && IsAtManualScopeBodyIndent() && IsEnclosingScopeValid(init_expr)) {
-      scope_hoist_sink_->push_back(scope_hoist_indent_ + "Tensor " + name + " = " + init_expr + ";\n");
+      scope_hoist_sink_->push_back(IndentAtLevel(scope_hoist_indent_level_) + "Tensor " + name + " = " +
+                                   init_expr + ";");
```
YunjiQin: **Missing trailing newline on the hoisted carry decl — generates malformed C++.**

This hoist entry now ends in `";"` with no `\n`, whereas `main` pushed `... + init_expr + ";\n"`. The hoist list is still flushed verbatim with no separator:

```cpp
for (const auto& line : hoisted) {
  Active().AppendRaw(line);   // AppendRaw adds no newline
}
Active().AppendRaw(parent_indent + "PTO2_SCOPE(PTO2ScopeMode::MANUAL) {\n");
```

The batch path (`EmitBatchedAllocTensors`) pushes `batch_emitter.GetCode()`, which *is* newline-terminated, so the flush loop relies on every entry carrying its own `\n`. This carry-decl entry breaks that invariant.

When a `pl.manual_scope` body hoists a mutable tensor carry decl through this path (the cross-scope tensor case, #1697), the output becomes:

```cpp
    Tensor foo = bar;    PTO2_SCOPE(PTO2ScopeMode::MANUAL) {
```

i.e. the hoisted decl is glued onto the `PTO2_SCOPE` header — uncompilable orchestration code. Suggest restoring the `";\n"` terminator (or flushing the list via `EmitLine` instead of `AppendRaw`).

src/codegen/orchestration/orchestration_codegen.cpp
```diff
@@ -2291,30 +2247,140 @@ class OrchestrationStmtCodegen : public CodegenBase {
...
+      }
+      cg.EmitTaskParamsDecl(task_var);
+      for (const auto& p : params) {
+        cg.EmitIndentedLine(task_var + "." + ArgDirectionToMethodName(p.direction) + "(" + p.value + ");");
+      }
+      cg.EmitSelectiveDumpCall(task_var, params);
+      if (dist_ctx_callee != nullptr) {
+        cg.EmitDistTensorCtxScalars(call, dist_ctx_callee, task_var);
+      }
+      cg.EmitManualDeps(call, task_var);
```
YunjiQin: **Unified `Emit` reorders emission for the SPMD / group / mixed paths.**

The old `GenerateSpmdCallCode` / `GenerateGroupCallCode` / `GenerateMixedGroupCallCode` emitted `LaunchSpec → EarlyResolveHint → ManualDeps`; this unified plan emits `ManualDeps → LaunchSpec → EarlyResolveHint`. For the mixed path, the `MixedKernels mixed_N = {...}` line (now `pre_lines`) also moves *ahead* of the `L0TaskArgs` decl rather than after the params/dump block.

These setters are independent and all still precede the submit, so runtime semantics are unchanged — but the **generated text changes** for those three paths. The PR states output is unchanged and the 147/147 snapshot gate passes; the direct-call path order is preserved, so direct-call snapshots can't cover this. Could you confirm the reorder is intended and that the SPMD/group/mixed paths are actually exercised by the snapshot suite?

src/codegen/codegen_preconditions.cpp
```diff
@@ -0,0 +1,94 @@
...
+void VerifyDistributedCodegenPreconditions(const ProgramPtr& program) {
+  CHECK(program != nullptr) << "DistributedCodegen preconditions: program must not be null";
+
+  DistributedTensorUseCollector collector;
+  collector.VisitProgram(program);
+  if (!collector.uses_distributed_tensor) {
+    return;
+  }
+
+  CHECK(!collector.has_windowing_ops || collector.has_comm_domain_scope)
```
YunjiQin: Per `.claude/rules/error-checking.md`, a failed precondition here means a pass was omitted from the strategy — a compiler/configuration bug, not user input — so this should be `INTERNAL_CHECK` (raises `InternalError`) rather than `CHECK` (raises `pypto::ValueError`, user-facing). The PR description also says each check "fires `INTERNAL_CHECK_SPAN` with the same message pattern as #1700", but the implementation uses plain `CHECK` + `pass::VerifyProperties` — worth aligning the code with the description.

YunjiQin (commented):
A few notes from reviewing the CodeEmitter migration. One looks like a real codegen regression (missing newline on the hoisted carry decl); the other two are correctness-adjacent things worth confirming.

src/codegen/orchestration/orchestration_codegen.cpp
georgebisbas: Good catch — the reorder was unintentional. When the four paths (`GenerateSpmdCallCode` / `GenerateGroupCallCode` / `GenerateMixedGroupCallCode` / direct-call inline) were collapsed into the single `TaskDispatchPlan::Emit`, the unified order accidentally swapped `ManualDeps` ahead of `LaunchSpec`/`EarlyResolveHint`, and `MixedKernels` ahead of `L0TaskArgs`.

**Fixed.** The canonical order is now restored to match what was on `main`:
```
L0TaskArgs → params → dump → MixedKernels → LaunchSpec → EarlyResolveHint → ManualDeps → submit
```

Two changes in `TaskDispatchPlan::Emit`:
- Moved `pre_lines` loop after `EmitDistTensorCtxScalars` (so `MixedKernels` lands back after params, not before `L0TaskArgs`)
- Moved `EmitManualDeps` after `EmitEarlyResolveHint` (restoring `LaunchSpec → EarlyResolveHint → ManualDeps` order for all paths)

**Two new ordering tests guard against regression:**
- `test_mixed_group_emission_order` — asserts `MixedKernels` after `L0TaskArgs`, `launch_spec` before submit, using `code.index()` comparisons
- `test_direct_call_with_deps_emission_order` — asserts `L0TaskArgs` before `set_dependencies` before submit

Both test the generated C++ text order directly — a future reorder in `Emit` will fail with a clear `assert l0_index < mixed_index` message, not silently pass.

Existing SPMD/mixed tests (`test_standalone_spmd_dispatches_group_with_spmd_launch_spec`, `test_spmd_mixed_multi_out_single_return_alias_targets_actual_return`) continue to pass. The one pre-existing failure (`test_no_split_mixed_group`) is a `pl.auto_chunk` parser issue in the Docker image, unrelated.

src/codegen/orchestration/orchestration_codegen.cpp
georgebisbas: fixed via restoring the `;\n` terminator

georgebisbas: Also added a small regression test for that.

src/codegen/orchestration/orchestration_codegen.cpp
```diff
@@ -300,7 +215,7 @@ class OrchestrationStmtCodegen : public CodegenBase {
 
   void SetTupleVarToKey(std::map<const Var*, std::string> mapping) { tuple_var_to_key_ = std::move(mapping); }
 
-  void SetInitialIndent(int indent) { indent_ = indent; }
+  void SetInitialIndent(int indent) { emitter_.SetIndentLevel(indent / 4); }
```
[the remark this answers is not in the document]
georgebisbas: fixed

src/codegen/codegen_preconditions.cpp
georgebisbas: updated

src/codegen/orchestration/orchestration_codegen.cpp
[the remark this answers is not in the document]
georgebisbas: check again

src/codegen/orchestration/orchestration_codegen.cpp
```diff
@@ -3012,13 +3021,15 @@ class OrchestrationStmtCodegen : public CodegenBase {
   /// Caller guarantees the decl type is ``Tensor``.
   void EmitMutableTensorCarryDecl(const std::string& name, const std::string& init_expr) {
     if (scope_hoist_sink_ != nullptr && IsAtManualScopeBodyIndent() && IsEnclosingScopeValid(init_expr)) {
-      scope_hoist_sink_->push_back(scope_hoist_indent_ + "Tensor " + name + " = " + init_expr + ";\n");
+      scope_hoist_sink_->push_back(IndentAtLevel(scope_hoist_indent_level_) + "Tensor " + name + " = " +
+                                   init_expr + ";");
```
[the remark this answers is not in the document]
georgebisbas: check again

src/codegen/orchestration/orchestration_codegen.cpp
```diff
@@ -302,7 +217,7 @@ class OrchestrationStmtCodegen : public CodegenBase {
 
   void SetTupleVarToKey(std::map<const Var*, std::string> mapping) { tuple_var_to_key_ = std::move(mapping); }
 
-  void SetInitialIndent(int indent) { indent_ = indent; }
+  void SetInitialIndent(int indent) { emitter_.SetIndentLevel(indent / 4); }
```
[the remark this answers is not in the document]
georgebisbas: check again

src/codegen/codegen_preconditions.cpp
```diff
@@ -0,0 +1,94 @@
...
+      << "DistributedCodegen preconditions: MaterializeCommDomainScopes must run before "
+         "DistributedCodegen when window-buffer/distributed-tensor ops are present.";
+
+  // Comm-domain materialization is required when DistributedTensor values are
+  // present in host orchestration paths.
+  pass::VerifyProperties(IRPropertySet{IRProperty::CommDomainScopesMaterialized}, program,
+                         "DistributedCodegen preconditions");
```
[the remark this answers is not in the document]
georgebisbas: check again

lyfne123 (approved):

PR #1834, opened 2026-06-23
fix(runtime): default to tensormap_and_ringbuffer runtime

lyfne123:
## Summary

`pto_backend` bakes `"runtime": "tensormap_and_ringbuffer"` into every generated `kernel_config.py`, so `CompiledProgram.runtime_name` is always `tensormap_and_ringbuffer`. But `ChipWorker._DEFAULT_RUNTIME` and the `compile_and_assemble` fallback in `device_runner.py` defaulted to `host_build_graph` — which is not a real runtime directory in `simpler`.

As a result, a default-constructed `with ChipWorker():` never bind-matched a freshly compiled program:
- the reuse lookup in `execute_on_device` silently fell through to a one-shot worker (the active worker was never reused), and
- explicit `ChipWorker.run(compiled)` raised a spurious runtime-mismatch `ValueError`.

This change aligns both defaults with the runtime the backend actually emits, so the reuse and explicit-dispatch paths bind-match.

## Changes

- `worker.py` — `_DEFAULT_RUNTIME` → `tensormap_and_ringbuffer` (+ corrected the explanatory comment).
- `device_runner.py` — `compile_and_assemble` fallback → `tensormap_and_ringbuffer` (+ docstring examples).
- `kernel_compiler.py`, `compiled_program.py` — docstring example updates.
- `test_worker_reuse.py`, `test_chip_worker_explicit_dispatch.py` — sync fixtures and `current()` lookups to the new default.

Retained `host_build_graph` references are intentional: the explicit-mismatch negative test (`test_runtime_mismatch_returns_none`) and the `host_dlopen_count` variant docstring.

## Testing

- [x] `ruff check` / `ruff format` / `pyright` clean (pre-commit hooks pass)
- [x] Full unit suite: 6195/6219 pass. The remaining failures are confined to `tests/ut/runtime/` and stem from a local `simpler` install missing `RunTiming` — they reproduce identically on `main` and are unrelated to this change (CI installs a matching `simpler`).
- [x] Code review completed

### DIFF
--- a/python/pypto/ir/compiled_program.py
+++ b/python/pypto/ir/compiled_program.py
@@ -648,7 +648,7 @@ class CompiledProgram:
 
     @property
     def runtime_name(self) -> str:
-        """Runtime ABI name baked into ``kernel_config.py`` (e.g. ``"host_build_graph"``)."""
+        """Runtime ABI name baked into ``kernel_config.py`` (e.g. ``"tensormap_and_ringbuffer"``)."""
         self._ensure_runtime_loaded()
         assert self._runtime_name is not None
         return self._runtime_name
--- a/python/pypto/runtime/device_runner.py
+++ b/python/pypto/runtime/device_runner.py
@@ -302,7 +302,7 @@ def compile_single_kernel(
         compiler: Configured :class:`KernelCompiler` instance.
         platform: Target execution platform.
         pto_isa_root: Resolved PTO-ISA root directory.
-        runtime_name: Runtime name (e.g. ``"host_build_graph"``).  Passed to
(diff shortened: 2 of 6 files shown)


Hzfengsy (approved):
Clean, well-scoped fix. Aligning the default runtime name with the value `pto_backend` bakes into every `kernel_config.py` (`tensormap_and_ringbuffer`) is correct — it fixes both the silent reuse fall-through and the spurious explicit-dispatch mismatch. Docstrings, comments, and the two runtime test files are kept in sync, and the retained `host_build_graph` references (negative test + variant docstring) are intentional. LGTM.

PR #1842, opened 2026-06-23
feat(ir): full-K snake M/N tiling with operand reuse in AutoTileMatmulL0

tonibohnlein:
## Summary

Follow-up to #1815 (on-core M/N output tiling). When `AutoTileMatmulL0`'s M/N tiling fires on a plain `tile.matmul` whose `[M, N]` output exceeds L0c **and** the chooser returns `k == K` (the full K panel fits L0a/L0b), this PR:

1. **Adds the missing `k == K` coverage** — each output sub-tile is a single straight-line `tile.matmul` over the full K, instead of bailing (the old `K/k < 2` skip, which left the matmul to overflow L0c at `AllocateMemoryAddr`) or forcing a degenerate 1-trip pipelined K-loop.
2. **Implements cross-sub-tile operand-content reuse** via a serpentine **snake** traversal with **auto row/column scheduling**.

## The snake + reuse

In the full-K case the whole operand panel fits L0, so the `ceil(M/m) × ceil(N/n)` grid is emitted as a serpentine snake that keeps one panel resident and reuses it across sub-tiles. The left `[m, K]` panel depends only on the M-row and the right `[K, n]` panel only on the N-col, so a fresh `tile.extract` is emitted **only when that index changes**. Reversing the inner sweep each step also reuses the moving panel across every turn:

> operand extracts: **`P·Q + 1`** instead of `2·P·Q`  (`P = ceil(M/m)`, `Q = ceil(N/n)`)

The same Left/Right SSA Var feeds multiple matmuls, extending the L0 tile's lifetime so the allocator keeps it resident.

### Row vs column — analytic decision

Keep the **more expensive** panel stationary (loaded `P` or `Q` times instead of `~P·Q`). With `A_cost = m·K·bytes_a`, `B_cost = K·n·bytes_b`:

```
T_row − T_col = (P−1)(Q−1) · (B_cost − A_cost)
```

`(P−1)(Q−1) ≥ 0`, so the sign is just `sign(B_cost − A_cost)` → **row snake** (A-stationary) when the left panel is larger, else **column snake** (B-stationary). For equal dtypes this reduces to the tile aspect ratio (`m` vs `n`).

## Scope (precise)

- **Reuse is `k == K` only.** The split-K M/N path (panel does **not** fit L0) keeps the existing per-sub-tile K-loop — cross-sub-tile reuse there needs multi-accumulator strips, a separate change.
- **The L0-tile chooser is unchanged** (still schedule-unaware — its traffic model ignores `k`). This PR optimizes the *traversal* of the already-chosen full-K tile; a schedule-aware chooser that co-optimizes `(m, n, k)` with the stationary choice is the next step.
- Deferred regimes are unchanged and still emit `PH-AT-006`: `matmul_acc`, `Vec` left operand (PV), and non-store consumers (Mat-scratch path).

## Testing

Both snake directions, asserting exact reuse counts, `SSAForm` + `UseAfterDef`, and numerical agreement with `torch.matmul`:

| Case (BF16, 910B) | Direction | Extracts | Grid |
| --- | --- | --- | --- |
| 256×64 @ 64×256 | row snake (A-stationary) | 2 Left + 3 Right (`P·Q+1 = 5`) | 2×2 |
| 384×64 @ 64×256 | column snake (B-stationary) | 5 Left + 2 Right (`P·Q+1 = 7`) | 3×2 |

Validated end-to-end (operand reuse survives `InferTileMemorySpace → MemoryReuse → AllocateMemoryAddr → codegen`, maxdiff 0.0). Full local suite: 23 pass-tests + 1992 transforms/jit regression green; `pre-commit` all hooks pass; pyright 0 errors. Docs updated (en + zh-cn).


tonibohnlein:
Follow-up to #1815. @Hzfengsy — would appreciate a review when you have a chance. This adds the `k == K` full-K M/N coverage **and** implements cross-sub-tile operand-content reuse (serpentine snake with auto row/column scheduling). Scope is deliberately precise: reuse is full-K-only (split-K keeps the per-sub-tile K-loop), and the L0-tile chooser is unchanged — this optimizes the traversal of the already-chosen tile; the schedule-aware chooser is the next PR.

tonibohnlein:
Thanks for the reviews.

- **CodeRabbit**: no actionable comments — 👍
- **@gemini-code-assist**: applied. Replaced the defensive `lhs_tile ? … : 1` / `rhs_tile ? … : 1` fallbacks in `BuildFullKSnake` with an `INTERNAL_CHECK_SPAN(lhs_tile && rhs_tile, sp)`. `AnalyzeMatmul` already bails (`return std::nullopt`) unless both operands are `TileType` before the `MatmulTiling` is built, so the casts can't fail here — and a silent `bytes = 1` fallback would have skewed the panel-cost comparison and picked the wrong snake direction. This matches the project's `error-checking.md` rule (pass-invariant failures use `INTERNAL_CHECK`, not silent defaults).

Pushed as `ae921973`; CI re-running.

tonibohnlein:
Closing to redesign the full-K path.

The cost model (pto-isa A2A3: cube ≈ 4096 bf16 MAC/cyc; L1→L0 extract 220–441 B/cyc) shows the full-K sub-tile is **compute-bound** — the operand extract is only ~14–19% of the matmul. Two consequences:

1. The serpentine **operand-reuse** that is this PR's centerpiece saves *moving-operand traffic*, which is hidden in the compute-bound regime → negligible. The wins that actually matter are A-within-row reuse and **double-buffering** (~15%, latency-hiding).
2. Double-buffering on this hardware is realized only via `ForKind::Pipeline` loops (LowerPipelineLoops); an unrolled straight-line snake cannot be double-buffered.

So the right full-K emitter is **nested pipelined loops** (outer rows / inner cols, both `stage=2`) driving the placement callback — which delivers the A-reuse + double-buffering and makes the serpentine moot. That's a redesign of the emitter, not an increment on this PR, so reworking it here would invalidate the review. Will reopen as a fresh PR with the nested-pipelined-loop design + its cost-model justification. The placement-callback decomposition and tests from this work carry over.

PR #1897, opened 2026-06-30
feat(jit): @pl.jit.aic/aiv/group for hand-written mixed kernels

lyfne123:
## Summary

Adds `@pl.jit.aic` / `@pl.jit.aiv` / `@pl.jit.group(split=...)` flavors so a
cube/vector **mixed kernel can be hand-written inside the `@pl.jit` world**.

The automatic split (one mixed block + `optimizations=[pl.split(UP_DOWN)]`)
injects `tpush`/`tpop` boundary tiles the user never wrote, so they cannot be
pinned with `tile.alloc_buffer` and the whole-space-manual contract rejects
them — meaning an auto-split attention kernel cannot place its on-chip buffers
by hand. Authoring the two halves directly fixes this:

- `@pl.jit.aic` / `@pl.jit.aiv` → `FunctionType.AIC` / `AIV`
- `@pl.jit.group(split=...)` → `FunctionType.Group` with the split attr,
  dispatching the AIC + AIV pair (called from `with pl.spmd(N): group(...)`)

Each half carries its own pipe setup (`reserve_buffer` / `import_peer_buffer` /
`{aic,aiv}_initialize_pipe`) and explicit `tpush`/`tpop`, so every real tile is
`alloc_buffer`-pinnable and the cross-core `tpop` results stay buffer-less —
preserving the cube/vector overlap while giving full manual layout control.

## Changes

- `python/pypto/jit/decorator.py` — `_SubFunctionDecorator` gains `allow_split`;
  new `aic` / `aiv` / `group` flavors; `JITFunction._split`; `_discover_deps`
  admits the new func types.
- `python/pypto/jit/specializer.py` — `_emit_decorator` maps the flavors to the
  right `@pl.function(type=...)` (+ split attr); `split` threaded through
  `build_specialize_context` / `SpecializeContext`.
- `tests/ut/jit/test_manual_mixed.py` — end-to-end compile of a hand-written
  AIC+AIV kernel with `alloc_buffer` pins, plus a `split=`-on-non-group guard.

## Stacked on #1886

This branch is stacked on #1886 (`tile.alloc_buffer` + L0 pinning). Until #1886
merges, this PR's diff includes those 3 commits; the net-new commit here is the
jit extension. Checking out `feat-jit-manual-mixed` gives **both** features,
which is what hand-writing + pinning a mixed kernel needs.

## Verification

- `tests/ut/jit/test_manual_mixed.py` passes; full `tests/ut/jit/` (271) pass —
  no regression.
- A hand-written `fa_fused` (AIC + AIV halves, every tile `alloc_buffer`-pinned)
  compiles through the **entire PyPTO compiler** (all IR passes + codegen →
  valid `.pto`). The only remaining step, `ptoas` assembly, is environment-gated
  (a local toolchain lacking `pto.load_scalar`); it is not affected by this PR.


lyfne123:
Folding the jit extension into #1886 per request — commit 32000e9e is now on feat-tile-user-buffers. Closing this stacked PR.

Issue #1233, opened 2026-04-30
Frontend batch_matmul lowering emits invalid ND reshape/slice for 4-D matmul operands


ndleslx:
## Problem

When using 4-D tensor operands for `pl.matmul` in the Qwen3 32B decode kernel, frontend lowering converts the rank >= 3 tensor matmul to `tile.batch_matmul`, but the later ND-to-2D tile flattening path preserves an unnecessary local reshape/slice chain around the lhs operand.

In the model source, the problematic frontend pattern is equivalent to:

```python
a_chunk0 = pl.reshape(attn_proj_tile[0:1, :, :], [1, 1, BATCH, OUT_PROJ_K_CHUNK])
a_acc = pl.matmul(a_chunk0, wo[...])
```

The lowered IR before C++ codegen contains:

```python
t__tile = pl.tile.load(..., [1, 16, 128], ...)
a_chunk0__tile = pl.tile.reshape(t__tile, [1, 1, 16, 128])
lhs_nd_slice_0 = pl.tile.slice(a_chunk0__tile, [1, 1, 16, 128], [0, 0, 0, 0])
lhs_2d_0 = pl.tile.reshape(lhs_nd_slice_0, [16, 128])
pl.tile.tpush_to_aic(lhs_2d_0, split=0)
```

For zero-valid subblocks, the same lowering path produces ND tile views with valid shape `[0, 0, 0, 0]`. C++ codegen then emits an invalid small row-major BF16 tile, for example:

```cpp
Tile<TileType::Vec, bfloat16_t, 1, 1, BLayout::RowMajor, ...>
```

This fails static alignment checks during kernel compilation.

## Expected behavior

The frontend should recognize that the reshape is only changing batch dimensions for batch matmul. For this case:

```text
[1, 16, 128] -> [1, 1, 16, 128]
```

the actual matmul page remains `[16, 128]`, so the batch matmul lowering should use the loaded 2-D page directly instead of emitting `tile.reshape -> tile.slice -> tile.reshape`.

Expected lowered shape flow should be closer to:

```python
lhs_2d = pl.tile.load(..., [1, 16, 128], ...)
# or directly the flattened [16, 128] page
pl.tile.tpush_to_aic(lhs_2d, split=0)
```

## Suggested frontend fix

The fix likely belongs in:

```text
src/ir/transforms/flatten_tile_nd_to_2d_pass.cpp
```

[...]

PR #1922, opened 2026-07-01
chore(runtime): bump runtime submodule to simpler main 0a8bc943 (pick up pto-isa 83d01313)

chenshengxin2026:
## What

Move the pinned `runtime/` submodule pointer from `02bd0c4f` → **simpler main HEAD `0a8bc943`** (clean fast-forward, +37 commits).

This is the sanctioned mechanism for advancing pypto's **pto-isa**: `versions.env` does not declare pto-isa — the `toolchain` job derives `PTO_ISA_COMMIT` from the pinned runtime's `runtime/pto_isa.pin`. Bumping the submodule moves it in lockstep with the runtime pypto builds against (build == run).

| | before | after |
| --- | --- | --- |
| `runtime/` submodule | `02bd0c4f` | `0a8bc943` |
| derived `runtime/pto_isa.pin` | `e722679` | `83d01313d9bfc247c4b7c8bcf969d1019f0d106f` |

## pto-isa range

- `e722679` (2026-06-25) → `83d01313` (2026-07-01): **fast-forward, 33 ISA commits**.
- The pin itself was already advanced on simpler main in **simpler #1233**; this PR just points the runtime submodule at a simpler commit that carries it.

## Impact / notes

- **pto-isa only** from a toolchain standpoint; ptoas is unrelated and tracked separately in `toolchain/versions.env`.
- The runtime submodule advances +37 simpler commits (runtime refactors/fixes since 06-28), so this is a genuine runtime bump, not just a pin flip — CI (build + system-tests) is the gate on whether any pypto-side adaptation is needed.
- Downstream **pypto-lib** inherits the new pto-isa automatically on its next CI run (reads `runtime/pto_isa.pin` from pypto HEAD); no pypto-lib change needed.


chenshengxin2026:
Closing — this PR is mis-scoped.

pypto has no independent pto-isa knob: `versions.env` does not declare pto-isa, and the `toolchain` job derives `PTO_ISA_COMMIT` **only** from `runtime/pto_isa.pin` (the simpler commit the `runtime/` submodule points at), with a guard that fails the build on any hard-coded pto-isa commit. So the only way to move pypto's pto-isa is to move the runtime submodule pointer — which points at a whole simpler commit.

The runtime pointer currently lags simpler main by ~37 commits, so this bump is really a **runtime version jump** (HostApi refactor, arena-layout split, register/run vocabulary, …) that merely happens to carry the new pin — not a clean pto-isa bump. That belongs to the routine `chore(runtime): bump runtime submodule to simpler main …` work (with any needed pypto adaptation), not to a toolchain version bump.

The pto-isa version change itself is the source-of-truth pin on **simpler**, already merged in simpler#1233 (`pto_isa.pin` → 83d01313). The new pto-isa will reach pypto on the next routine runtime-submodule bump.

PR #1924, opened 2026-07-01
fix(ir): coalesce peeled loop-carried accumulator if-phis in MemoryReuse

tonibohnlein:
## Summary
Fixes a pre-existing MemoryReuse bug that **halves the usable L0C capacity for pipelined-K matmuls**. Reproduces on `main`; not chooser-specific.

`LowerPipelineLoops` peels the `stage=2` K-loop into `if`-phis whose live branch is an in-place `matmul_acc` (on the accumulator buffer) and whose dead `if k==0` branch is a fresh `matmul` seed on a *different* Acc buffer. `YieldFixupMutator` reconciled them by copying the accumulator onto the seed's buffer via an `acc->acc tile.move`. Two consequences:

1. **Only half of L0C is usable.** The reconciliation copy is the same size as the accumulator and **co-live** with it, so a single pipelined-K accumulator can occupy at most **half** of L0C — the other half is the phantom duplicate. On a2a3 (128 KB L0C) that caps the accumulator at 64 KB (e.g. `128×128` fp32); the `512×512×192 bf16` case wants the full 128 KB tile (`256×128` fp32 = 131072 B) and fails to compile with `Acc buffer usage (262144 bytes) exceeds platform limit (131072 bytes)` — **exactly 2×**. This directly bounds matmul tiling: the L0 roofline tiler (#1912) sizes L0C tiles against the *full* L0C, so without this fix its larger tiles overflow and it is forced into smaller, less efficient ones (more M/N sub-tiles → more accumulator drains and operand reloads).
2. **ptoas rejects the move.** Acc->Acc `tmov` has no legal address-space pair on any target (no legal Acc->Acc pair), so the peeled K-loop path fails to compile even when the duplicate would fit.

## Fix (`src/ir/transforms/memory_reuse_pass.cpp`)
- **`TopDownRetargeter::CoalesceAccumulatorIfPhis`** (Step 3.75): identifies the in-place-accumulator branch by its `reuses_input` producer and retargets the *other* branch's seed onto the accumulator buffer, so both branches share it and no move is emitted (matches `mad_acc`'s shared-`%dst` semantics). The retarget is **mandatory** — a declined retarget is an `INTERNAL_CHECK`, never a move, since no legal Acc->Acc move exists. A single coalesced accumulator can then use the **full** L0C.
- **Branch-locality + branch-scoped liveness guard.** The `check_liveness=false` bypass is sound only when (a) the seed producer is a `Call` lexically *inside* the non-accumulator branch — a pre-`if` value yielded through the branch runs unconditionally and would clobber the accumulator the sibling branch reads — and (b) the accumulator buffer is dead *within the branch* after the seed. A branch-scoped `IsTargetDeadAtAssign` (new `stop_at` param halts the scan at the enclosing `if`) enforces (b) while ignoring the mutually-exclusive sibling and the legitimate post-`if` phi consumers; when either (a) or (b) fails the phi is left to `YieldFixup`. The branch-tail scan rejects a later **read _or_ write-only clobber** of the accumulator base (`SubtreeWritesBase`, the dual of the reads-only collector) — a fresh producer aliasing the base without reading it would otherwise overwrite a retargeted seed before the phi consumes it.
- **`NormalizeIdentityCopyBuffersMutator`** (Step 4.5): reconciles a downstream bare-`Var` SSA identity copy of the retargeted return_var (which otherwise keeps the pre-coalesce buffer), and re-anchors an in-place op's output onto its reused input's new buffer when the input was retargeted (`ReanchorInplaceOutput`) — see below.

## Non-divisor K-peel
When the K tile does not divide K, `AutoTileMatmulL0` pipelines the full K-blocks and peels a straight-line `matmul_acc` tail. After `CoalesceAccumulatorIfPhis` retargets the loop accumulator, that tail's output — an in-place op aliasing the accumulator — stayed declared on the pre-coalesce buffer, orphaning it on a never-written Acc base (compiles, but reads a stale buffer → wrong result). `ReanchorInplaceOutput` re-anchors the tail onto the accumulator's coalesced buffer, mirroring the existing bare-`Var` and view-input rename handling.

## Testing
- **Unit:** structural before/after (`assert_structural_equal`) on a minimal accumulator if-phi (verified to regress pre-fix: 2 Acc buffers + phantom Acc->Acc move); a gating test proving the Acc-only scope (a Vec `fillpad_inplace` if-phi is left to the legal Vec->Vec move path); a branch-locality test (a pre-`if` seed stays on a distinct buffer); and a write-only-clobber test (a later write-only def of the accumulator base blocks the seed retarget). Full MemoryReuse suite **64 pass**; transforms **1778** unaffected.
- **Device (a2a3):**
  - `512x512x192 bf16` matmul — compiles through the full pipeline with **one `loc=acc` buffer of the full 128 KB L0C and zero `tmov`** in the emitted `.pto`; on-device numerics vs an f32 golden (Frobenius rel_err < 2e-2). Pre-fix this same shape overflowed at 2× (262144 > 131072).
  - Non-divisor K-peel `128x688x128 bf16` (k=128 peel: 5 full K-blocks + 48 tail) — **compiles and is numerically correct: 0 / 16384 mismatches** at rtol=atol=2e-3 (residuals at pure fp32-accumulation rounding, ~1e-5 abs), versus a compile-but-wrong 16378 / 16384 without the `ReanchorInplaceOutput` fix.
  - Broader AutoL0 sweep (the `_AUTOL0_SHAPES` set, bf16, A-/B-stationary, chained mat-scratch, both K-peels) — 11 / 11 pass, no regression.

## Docs
Updated `docs/en|zh-cn/dev/passes/29-memory_reuse.md` (Algorithm + Implementation).


tonibohnlein:
Thanks — addressed in 0faf7a7d.

**@chatgpt-codex-connector (P1) and @coderabbitai (Major) — `check_liveness=false` bypass unsound beyond the peeled-loop shape.** Both are correct: branch exclusivity only justifies skipping the *global* liveness check when (a) the seed producer is lexically **inside** the non-accumulator branch, and (b) the accumulator buffer is dead **within that branch** after the seed. A pre-if Acc value yielded through the branch would otherwise be retargeted onto the accumulator and clobber it before the `if`; a same-branch tail read was likewise unprotected.

`TryCoalesceAccIfPhi` now, before the (still-mandatory) retarget:
- requires the seed to be a `Call` producer lexically inside the branch (checked via `seed_def.ancestors` containing the `IfStmt`); otherwise the phi is left to `YieldFixupMutator` — this also removes the spurious hard-error path on a non-`Call` seed;
- runs a **branch-scoped** `IsTargetDeadAtAssign` (new `stop_at` boundary that halts the scan at the enclosing `if`) so a same-branch tail read of the accumulator buffer is caught, while the mutually-exclusive sibling branch and the legitimate post-`if` phi consumer are correctly ignored.

The production peeled-loop path is unchanged (both preconditions hold), so `512x512x192 bf16` still coalesces to one L0C buffer. Added `test_pre_if_acc_seed_not_coalesced_onto_accumulator` (the pre-if seed stays on a distinct buffer — no clobber), and docs step 5 (en + zh-cn) now states the two preconditions rather than "exclusivity directly".

**@coderabbitai (nitpick) — s-test regression coverage implicitly depends on the tiler picking 176x176.** Valid observation; leaving the s-test as-is by design. The s-test's role is on-device numerics; the *deterministic* regression coverage of the peeled accumulator if-phi lives in the unit tests, which run `lower_pipeline_loops` directly and assert the coalescing shape independent of the tiler heuristic — `test_pipelined_kloop_accumulator_coalesces_to_one_acc_buffer` (one Acc buffer, no `tile.move`) and `test_accumulator_if_phi_seed_retargets_to_accumulator_buffer` (structural before/after). So a future tiler change can't silently drop coverage of the buggy path.

tonibohnlein:
Hi @lyfne123 — a question on this PR given the level2/level3 memory-planning split.

**#1924 fixes a level3-only bug.** On the default path (`memory_planner=PYPTO`), the peeled loop-carried accumulator — the `if k==0` seed branch and the non-divisor K-tail — lands on a *second* Acc buffer, so `YieldFixup` emits an illegal `Acc→Acc tile.move` and the accumulator is capped at half L0C.

**On level2 (`memory_planner=PTOAS`, #1934) it doesn't reproduce.** `MaterializeSemanticAliases` handles the loop-carry, and PlanMemory's lifetime coalescing merges the peeled seed + K-tail on its own — I verified `init` and the tail buffer get the same address. So #1924 only matters as long as the level3 planner path exists.

**So:** do you want to keep level3 as a supported path (→ I'll finish/merge #1924), or is level2/PlanMemory the long-term single path (→ I'll close #1924 as moot)? It's rebased and green either way.

lyfne123 (approved):

PR #1965, opened 2026-07-07
fix(distributed): Correct L3 put atomic add barrier

hashiqiqixian:
## Summary
- Fix the `test_atomic_add_accumulate` program in `tests/st/distributed/test_l3_put.py`.
- Change phase 3 so every rank notifies the root, but only the root waits on the root-local signal cell.
- Re-enable the atomic-add put ST by removing the outdated skip marker.

## Testing
- [x] `python -m py_compile tests/st/distributed/test_l3_put.py`
- [x] `git diff --check`
- [ ] Full ST execution not run locally: current environment is missing `cmake`, `pytest`, and `torch`.

## Related Issues
N/A


tests/st/distributed/test_l3_put.py
```diff
@@ -361,7 +369,6 @@ def test_ring_shuffle_pipeline(self, test_config, device_ids):
             f"pipeline ring put mismatch: max diff = {(outputs - expected).abs().max().item()}"
         )
 
-    @pytest.mark.skip(reason="atomic-add put still fails on the current runtime/PTOAS stack")
```
georgebisbas, line 364: Can you mention in the PR description the related PTOAS fix, just to have it for the backlogs? Thanks.

hashiqiqixian: Thanks for the suggestion. I checked this again and I don't think this PR should reference a PTOAS fix.

The failure was caused by the test program's phase-3 synchronization protocol: all ranks notified the root signal, but every rank then waited on its own local signal for `nranks`. In a reduce-to-root atomic-add put, only the root-local signal receives all notifications, so only the root should wait before reading the accumulated result.

So this PR fixes the program-level notify/wait protocol and re-enables the test; it is not tied to a PTOAS bug/fix.

georgebisbas: Sounds good!

lyfne123 (approved):

PR #1967, opened 2026-07-07
feat(distributed): push-based allgather composite intrinsic

georgebisbas:
## Summary

This PR rewrites `pld.tensor.allgather` from a **pull-based** 4-argument composite intrinsic into a **push-based** 3-argument one. Instead of each rank pulling every other rank's chunk via `pld.tile.get`, each rank now **pushes** its own chunk to all peers via `pld.tile.put`. The staging window becomes the gathered result (`window-as-result`), dropping the separate `out` Tensor. This matches the same push-based / return-in-place idiom already used by `reduce_scatter`.

**footprint** drops from $(NR+1) \times SIZE$ to $NR \times SIZE$, since the separate output buffer is eliminated. The API is also unified across the HOST builtin and InCore composite paths — both use the same 3-argument signature with identical arg positions.

---

## Motivation

YunjiQin's review on #1937 proposed the push-based approach for `allgather`, mirroring the `reduce_scatter` pattern. Each rank pushes its chunk to every peer's window slot; the window itself holds the full `[NR, SIZE]` gathered result after the barrier.

### 1. Core lowering — push-based decomposition (`lower_composite_ops_pass.cpp`)

**Before** (5 phases, pull-based):
1. `tile.load(local_data)` — materialize tile from Tensor input
2. `tile.store(stage, target)` — stage this rank's chunk into window
3. `notify-all` / `wait-all` — cross-rank barrier
4. `for r in 0..NR-1: pld.tile.get(out, peer=r, target, stage, ...)` — pull each peer's chunk from their window slot into `out`
5. Return `out` (Tensor rebind)

**After** (4 phases, push-based):
1. `tile.create [1, SIZE]` — VEC staging tile for auto-chunking
2. `for peer in 0..NR-1: pld.tile.put(target, peer, local_data, stage, ...)` — push this rank's chunk into every peer's window at `target[my_rank, :]`. Self-store uses HCCL identity mapping.
3. `notify-all` / `wait-all` — cross-rank barrier
4. Return `target` (DistributedTensor rebind) — window IS the result

Key design decisions:
- `pld.tile.put` auto-chunks when SIZE exceeds staging-tile capacity — a single `[1, SIZE]` tile works regardless of problem size
- No `tile.load` or `tile.store` in the decomposition; `pld.tile.put` reads from a Tensor or Tile source directly
- Dropped the loop-variable→column-offset arithmetic (`r*SIZE`), since each rank writes to a fixed row `my_rank`

### 2. Unified type deducer (`collective.cpp`)

- **4-arg InCore** → **3-arg** (both paths): `(local_data, target, signal)`
- Fixed arg roles across both paths: `args[1]=target`, `args[2]=signal` (previously the 2-arg HOST path used `args[0]=target`, `args[1]=signal`)
- Both paths now return `DistributedTensorType` (window-as-result)
- Removed type-dispatch on `args[0]` — the deducer trusts the lowering passes to validate per their context
- REGISTER_OP description updated from 2-arg/4-arg to "unified 3-arg"

### 3. Alias analysis (`convert_tensor_to_tile_ops_pass.cpp`)

- `GetWriteTargetExpr`: distinguishes 2-arg HOST (`args[0]`) vs 3-arg InCore (`args[1]`)
- `AnalyzeCallAccess`: both `target` and `signal` → InOut, `local_data` → In (no more `out` → Out)
- Verified via `test_convert_tensor_to_tile_ops.py` (1/1 pass)

### 4. Host lowering (`lower_host_tensor_collectives_pass.cpp`)

- `MakeBuiltinAllGather`: signal from `args[2]` (was `args[1]`)
- Rule table updated: `scope_buffers` uses `args[1]`/`args[2]`, `signal_expr` uses `args[2]`, `alias_source` uses `args[1]`

### 5. Window buffer analysis (`materialize_comm_domain_scopes_pass.cpp`)

- Removed type-dispatch branch on `args[0]`
- Always uses `args[1]=target`, `args[2]=signal`

### 6. Python DSL (`python/pypto/language/distributed/op/tensor_ops.py`)

- Removed `isinstance(local_data, DistributedTensor)` dispatch branch — the HOST path is detected at compile time
- Always forwards `(local_data_expr, target_expr, signal_expr)` — 3-arg
- Returns `DistributedTensor` (window-as-result) instead of `Tensor`
- Removed the `@overload` for 2-arg HOST convenience form per reviewer feedback
- Docstring updated with push-based semantics

### 7. Python IR builder (`python/pypto/ir/op/distributed/tensor_ops.py`)

- Unified to 3-arg form
- HOST dispatch guard tightened

### 8. Documentation

- `docs/en/dev/passes/12-lower_composite_ops.md` — updated allgather section with 4-phase push decomposition
- `docs/zh-cn/dev/passes/12-lower_composite_ops.md` — Chinese translation updated

### 9. Tests — all passing in Docker sim (7040 UT, 0 failures)

| Test | What changed |
|------|-------------|
| `test_lower_composite_ops.py` | `_build_allgather_before` → 3-arg; ops set: added `put`, removed `get`/`store`/`create` |
| `test_convert_tensor_to_tile_ops.py` | 3-arg form; direction inference (target/signal InOut, local_data In) |
| `test_l3_tensor_allgather_intrinsic.py` | Rewritten for push-based; stage-out via `pl.load`+`pl.store` from window |
| `test_host_orch_distributed.py` | Updated call: `allgather(data, data, signal)`, signal shape `[4]` |
| `test_lower_host_tensor_collectives.py` | Updated call: `allgather(data, data, signal)` |
| `test_l3_host_tensor_allgather.py` | Updated call |

## Review feedback addressed

All 7 review threads resolved:
- **gemini-code-assist**: `GetWriteTargetExpr` now handles 2-arg vs 3-arg overloads correctly (was incorrectly aliasing `signal` as write target for HOST path)
- **copilot-reviewer**: 2-arg HOST dispatch now requires both args be `DistributedTensor`; IR builder raises clear `TypeError` on missing `target`
- **copilot-reviewer**: `@overload` signatures added for type safety, mirroring `allreduce`
- All 7 threads resolved


src/ir/transforms/convert_tensor_to_tile_ops_pass.cpp
```diff
@@ -850,11 +850,12 @@ ExprPtr GetWriteTargetExpr(const CallPtr& call) {
...
-  // is the primary write target; target (args_[1]) is used for staging only.
-  if (IsOp(call, "pld.tensor.allgather") && call->args_.size() >= 4) {
-    return call->args_[3];
+  // pld.tensor.allgather(local_data, target, signal): push-based — writes
+  // gathered chunks into target on every rank (Phase 1 remote_store loop).
+  // target (args_[1]) is the primary write target and also the result
+  // (window-as-result).  signal (args_[2]) is used for barrier only.
+  if (IsOp(call, "pld.tensor.allgather") && call->args_.size() >= 2) {
+    return call->args_[1];
   }
```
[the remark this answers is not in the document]
georgebisbas: Fixed in `f2fe1032` — `GetWriteTargetExpr` now distinguishes the two overloads: 2-arg HOST returns `args_[0]` (target), 3-arg InCore returns `args_[1]`. `AnalyzeCallAccess` got a matching 2-arg branch (target+signal both InOut).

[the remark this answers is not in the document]
georgebisbas: Fixed in `f2fe1032` — `GetWriteTargetExpr` now handles both overloads explicitly (2-arg HOST → `args_[0]`, 3-arg InCore → `args_[1]`).

src/ir/transforms/convert_tensor_to_tile_ops_pass.cpp
```diff
@@ -850,11 +850,12 @@ ExprPtr GetWriteTargetExpr(const CallPtr& call) {
...
-  // chunks into out on every rank (Phase 3 per-peer pld.tile.get).  out (args_[3])
-  // is the primary write target; target (args_[1]) is used for staging only.
-  if (IsOp(call, "pld.tensor.allgather") && call->args_.size() >= 4) {
-    return call->args_[3];
+  // pld.tensor.allgather(local_data, target, signal): push-based — writes
+  // gathered chunks into target on every rank (Phase 1 remote_store loop).
+  // target (args_[1]) is the primary write target and also the result
+  // (window-as-result).  signal (args_[2]) is used for barrier only.
+  if (IsOp(call, "pld.tensor.allgather") && call->args_.size() >= 2) {
+    return call->args_[1];
```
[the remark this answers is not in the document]
georgebisbas: Fixed in `f2fe1032` — both `GetWriteTargetExpr` (lines ~853) and `AnalyzeCallAccess` (lines ~1031) now branch on `args_.size()`: 2-arg HOST maps target=`args_[0]` (both args InOut, like allreduce); 3-arg InCore keeps target=`args_[1]`.

python/pypto/language/distributed/op/tensor_ops.py
```diff
@@ -534,30 +534,29 @@ def allgather(
...
-            out: InCore only: :class:`pl.Tensor` [1, NR*SIZE] output.
 
         Returns:
-            InCore: the ``out`` :class:`pl.Tensor`.  HOST: the rebound
-            :class:`pld.DistributedTensor`.
+            :class:`pld.DistributedTensor` — the ``target`` window holding the
+            gathered ``[NR, SIZE]`` result.
     """
-    if isinstance(target, DistributedTensor) and signal is None and out is None:
+    if isinstance(target, DistributedTensor) and signal is None:
```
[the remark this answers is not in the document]
georgebisbas: Fixed in `f2fe1032` — the 2-arg HOST dispatch now requires both `local_data` and `target` be `DistributedTensor` (plus `signal is None`), and the InCore path raises a descriptive `TypeError` when `target`/`signal` are missing.

python/pypto/ir/op/distributed/tensor_ops.py
```diff
@@ -306,24 +305,26 @@ def allgather(
...
         signal: Window-bound INT32 barrier tensor.
-        out: For 4-arg: Tensor [1, NR*SIZE] output.
     """
     actual_span = _get_span_or_capture(span, frame_offset=1)
-    if signal is None and out is None:
+    if signal is None:
         # 2-arg HOST builtin form: allgather(data, signal)
+        # Positional mapping: data→local_data, signal→target
         _args: list[Expr] = [local_data, target]  # type: ignore[assignment]  # target is non-None here
         return _ir_core.create_op_call("pld.tensor.allgather", _args, {}, actual_span)
```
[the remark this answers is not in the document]
georgebisbas: Fixed in `f2fe1032` — the 2-arg path now raises a clear `TypeError` when `target` (the positional signal) is `None`, and the `# type: ignore` was removed since `target` is proven non-None.

python/pypto/language/distributed/op/tensor_ops.py
```diff
@@ -515,14 +515,14 @@ def allgather(
     local_data: Tensor | DistributedTensor,
     target: DistributedTensor | None = None,
     signal: DistributedTensor | None = None,
-    out: Tensor | None = None,
-) -> Tensor | DistributedTensor:
+) -> DistributedTensor:
```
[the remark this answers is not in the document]
georgebisbas: Fixed in `f2fe1032` — added `@overload` signatures for the HOST (2-arg) and InCore (3-arg) forms, mirroring `allreduce`.

python/pypto/language/distributed/op/tensor_ops.py
```diff
@@ -511,18 +511,30 @@ def broadcast(
...
 
-        **InCore composite (4 args):** ``pld.tensor.allgather(local_data, target, signal, out)`` —
-        ``local_data`` is a plain Tensor [1, SIZE] with this rank's chunk.
-        The intrinsic handles ``pl.load``, stage-in, notify/wait, and per-peer
-        ``remote_load`` into ``out``.
+        **InCore composite (3 args):** ``pld.tensor.allgather(local_data, target, signal)`` —
+        push-based: each rank remote_stores its chunk into every peer's window
+        slot at ``target[my_rank, :]``, then notify/wait barrier; the window
+        itself becomes the gathered ``[NR, SIZE]`` result (window-as-result).
+        No separate output tensor — the window IS the result.
```
YunjiQin, line 561: unify incore and host level api to 3 args.

src/ir/transforms/lower_composite_ops_pass.cpp
```diff
@@ -709,41 +709,40 @@ ExprPtr LowerTensorBroadcastRule(const CallPtr& call, const std::vector<ExprPtr>
...
+//       trust model as pld.tile.get self-path).
 //   2a. notify-all (Set 1)
 //   2b. wait-all  (Ge 1)
-//   3.  for r in 0..NR-1:
-//         pld.tile.get(out, peer=r, target, stage,
-//                      dst_offsets=[0, r*SIZE], src_offsets=[0,0], shape=[1,SIZE])
-//       — transfer each peer's chunk directly into out at column offset
-//         [0, r*SIZE]; one shared [1, SIZE] VEC staging tile, no tile.concat
-//       return out  (Tensor rebind)
+//   return target  (DistributedTensor rebind) — window IS the gathered result
```
YunjiQin, line 1027: Phases: 
0. for r in 0..NR-1:        pld.tile.put(local_data, peer = r, target, stage, dst_offsets=[my_rank, 0], src_offsets=[0, 0], shape=[1, SIZE])
1. notify/wait all

put can auto chunk to support SIZE over UB with a small stage tile

python/pypto/language/distributed/op/tensor_ops.py
georgebisbas: Updated

georgebisbas: Done in 491de34 + 9e692f3. The C++ deducer now requires exactly 3 args and dispatches by args[0] type (DistributedTensor → HOST, Tile/Tensor → InCore). GetWriteTargetExpr, AnalyzeCallAccess, and ShouldSkipHostCollective all follow the same pattern. The 2-arg convenience wrapper remains in the DSL

src/ir/transforms/lower_composite_ops_pass.cpp
georgebisbas: Done in 491de34 + a6ece28. Replaced pld.tile.remote_store with pld.tile.put(local_data, peer=r, target, put_stage, dst_offsets=[my_rank,0], src_offsets=[0,0], shape=[1,SIZE]). local_data is passed directly as the Tensor source; pld.tile.put handles tile-load/chunking internally through the stage tile. The push loop is a single ForStmt over NR ranks, followed by notify-all/wait-all. Updated docs (en+zh) and test expectations. Passing all sim tests.

georgebisbas:
All Copilot AI review suggestions addressed in commit [`386252a`](https://github.com/hw-native-sys/pypto/pull/1967/commits/386252aa3adc5aae6722d60daab156af9da8959a):

**🔴 High:**
- ✅ `INTERNAL_CHECK_SPAN` now accepts both `TensorType || TileType` — `LowerCompositeOps` runs after `ConvertTensorToTileOps` in the default pipeline, so `local_data` may already be a Tile.

**🟡 Medium:**
- ✅ Removed misleading "ConvertTensorToTileOps runs later" comment — replaced with neutral statement that `pld.tile.put` reads from either Tensor or Tile source.
- ✅ 2-arg HOST builtin guard tightened: now also checks `isinstance(target, DistributedTensor)`.
- ✅ EN+ZH docs: replaced stale `tile.load` Phase 0 with `tile.create(stage)` description matching actual lowering.
- ✅ ST module docstring: `pld.tile.remote_store` → `pld.tile.put (TPUT-based)`.

**🟢 Low:**
- ✅ IR builder docstring: `tile.load` → `tile.create(stage)`.
- ✅ Lowering UT docstring: `remote_store` → `pld.tile.put`.

Sim Docker build+test: **7040 passed, 2 skipped, 0 failures** ✅

src/ir/transforms/lower_composite_ops_pass.cpp
```diff
@@ -962,49 +962,45 @@ ExprPtr LowerTensorBroadcastRule(const CallPtr& call, const std::vector<ExprPtr>
...
-  // handles the load internally).  Record which so we can emit tile.load
-  // after the shared setup expressions are available.
-  bool is_tensor_input = (As<TensorType>(local_data->GetType()) != nullptr);
-  INTERNAL_CHECK_SPAN(As<TileType>(local_data->GetType()) || is_tensor_input, span)
-      << "pld.tensor.allgather local_data must be TileType or TensorType, got "
-      << local_data->GetType()->TypeName();
+
+  // local_data must be TensorType at this point (ConvertTensorToTileOps runs later).
+  INTERNAL_CHECK_SPAN(As<TensorType>(local_data->GetType()), span)
+      << "pld.tensor.allgather local_data must be TensorType, got " << local_data->GetType()->TypeName();
```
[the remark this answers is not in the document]
georgebisbas: Fixed in commit 386252a: INTERNAL_CHECK_SPAN now accepts both TensorType and TileType via `As<TensorType>(...) || As<TileType>(...)`. Updated the comment to note that ConvertTensorToTileOps runs BEFORE LowerCompositeOps in the default pipeline, so local_data may already be converted to a TileType.

src/ir/transforms/lower_composite_ops_pass.cpp
```diff
@@ -1021,74 +1017,58 @@ ExprPtr LowerTensorAllGatherRule(const CallPtr& call, const std::vector<ExprPtr>
...
-  // The loop bound is the runtime nranks (comm.nranks_idx), matching the
-  // notify/wait phases.  This keeps the gather consistent with the barrier
-  // phases regardless of the comm-group size, avoiding a mismatch between
-  // compile-time target.shape[0] and the actual world size.
-
-  auto gather_stage =
+  // ---- Phase 0: tile.load removed — pld.tile.put handles tile load internally ----
+  // local_data is always a Tensor at this point in the pass pipeline
+  // (ConvertTensorToTileOps runs later).  pld.tile.put reads from the
+  // tensor source and auto-chunks into the stage tile.
```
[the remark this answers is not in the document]
georgebisbas: Fixed in commit 386252a: Removed the misleading comment claiming ConvertTensorToTileOps runs later. Replaced with a neutral statement that pld.tile.put can read from either a Tensor or Tile source and auto-chunks through the VEC staging tile.

python/pypto/language/distributed/op/tensor_ops.py
```diff
@@ -544,53 +544,80 @@ def broadcast(
...
+            (InCore only; None for HOST).
+
+    Returns:
+        :class:`pld.DistributedTensor` — the ``target`` window holding the
+        gathered ``[NR, SIZE]`` result.
     """
-    if isinstance(target, DistributedTensor) and signal is None and out is None:
-        # 2-arg HOST builtin path: allgather(data, signal)
-        # Positional mapping: data→local_data, signal→target
+    if isinstance(local_data, DistributedTensor) and signal is None:
```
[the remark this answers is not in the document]
georgebisbas: Fixed in commit 386252a: Tightened the 2-arg HOST builtin guard to also check isinstance(target, DistributedTensor). Now allgather(DistributedTensor, non-DistributedTensor) will fall through to the clear 3-arg error message instead of failing inside _unwrap_distributed_tensors.

tests/ut/ir/transforms/test_lower_composite_ops.py
```diff
@@ -990,20 +987,17 @@ def test_allgather_is_decomposed_to_primitives():
...
-    the barrier phases) instead of a compile-time unrolled loop — this
-    keeps the gather consistent with the notify/wait bounds regardless of
-    the actual comm-group size.  The gather ForStmt body emits pld.tile.get
-    for every peer (self-read via HCCL identity mapping), so there is no
-    per-rank IfStmt inside the gather loop."""
+    """Push-based allgather emits 3 ForStmts + 2 IfStmts: push loop, notify-all, wait-all.
+
+    Phase 1 (push) uses a runtime ForStmt over nranks_idx — every peer gets a
+    remote_store (self-store via HCCL identity mapping, no per-rank IfStmt).
+    Phase 2a/2b are the standard notify-all/wait-all loops with per-peer IfStmts."""
```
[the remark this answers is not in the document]
georgebisbas: Fixed in commit 386252a: Updated docstring to use pld.tile.put instead of remote_store, matching the actual lowering and required-op set (which asserts pld.tile.put, not remote_store).

tests/st/distributed/collectives/test_l3_tensor_allgather_intrinsic.py
```diff
@@ -12,11 +12,12 @@
...
-``target`` (DistributedTensor [NR, SIZE] staging window), ``signal``, and
-``out`` (plain Tensor [1, NR*SIZE]).  It handles the ``pl.load`` internally,
-synchronises, uses ``pld.tile.get`` to transfer from peers, and writes directly
-into ``out``.
+Push-based (3-arg): ``pld.tensor.allgather(local_data, target, signal)`` —
+each rank pushes its chunk to every peer's window slot via
+``pld.tile.remote_store``; after the notify/wait barrier, the window itself
+holds the gathered ``[NR, SIZE]`` result (window-as-result).  The InCore
+function then reads from the window and writes into the output tensor for
+host-side verification.
```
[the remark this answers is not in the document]
georgebisbas: Fixed in commit 386252a: Updated module docstring to pld.tile.put (TPUT-based), matching the actual lowering primitive used by LowerCompositeOps.

docs/en/dev/passes/12-lower_composite_ops.md
```diff
@@ -201,15 +201,15 @@ Only `ReduceOp::kSum` is supported in the first version; the C++ deducer rejects
...
+Signature: `allgather(local_data, target, signal)`. `local_data` is this rank's chunk (`Tensor` or `Tile` `[1, SIZE]`), `target` is a window-bound `DistributedTensor[NR, SIZE]` staging area that also serves as the result, and `signal` is the INT32 barrier. Push-based decomposition:
 
 - Phase 0: `tile.load(local_data, [0, 0], [1, SIZE])` — emit a Tile from the plain input when `local_data` is a `Tensor`; skipped when it is already a Tile
-- Phase 1: `tile.store(stage_tile, [0, 0], target)` — stage this rank's chunk into its private HCCL window at local row 0
+- Phase 1: for `peer` in `0..NR-1`, `pld.tile.put(target, peer, stage_tile, stage, [my_rank, 0], [0, 0], [1, SIZE])` — push this rank's chunk into every peer's window at row `my_rank`. Self-store (`peer == my_rank`) uses HCCL identity mapping. `pld.tile.put` auto-chunks when SIZE exceeds the staging-tile capacity
```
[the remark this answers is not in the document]
georgebisbas: Fixed in commit 386252a: Replaced the stale Phase 0 tile.load bullet with a description of tile.create for the VEC staging tile plus a note that pld.tile.put reads directly from the local_data source — no explicit tile.load is emitted. Updated Phase 1 to pass local_data directly to pld.tile.put.

docs/zh-cn/dev/passes/12-lower_composite_ops.md
```diff
@@ -201,15 +201,15 @@ allreduce 规则把单个组合 Call 展开成两道跨卡屏障，复用同一
 
 ### `pld.tensor.allgather`
 
-签名：`allgather(local_data, target, signal, out)`。`local_data` 是本 rank 的 chunk（`Tensor` 或 `Tile` `[1, SIZE]`），`target` 是窗口绑定的 `DistributedTensor[NR, SIZE]` 暂存区，`signal` 是 INT32 屏障，`out` 是接收结果的普通 `Tensor[1, NR*SIZE]`。展开序列对齐 simpler allgather 参考实现 (`simpler/examples/workers/l3/allgather_distributed/`)：
+签名：`allgather(local_data, target, signal)`。`local_data` 是本 rank 的 chunk（`Tensor` 或 `Tile` `[1, SIZE]`），`target` 是窗口绑定的 `DistributedTensor[NR, SIZE]` 暂存区同时又是结果，`signal` 是 INT32 屏障。基于推送 (push-based) 的展开：
 
 - Phase 0：`tile.load(local_data, [0, 0], [1, SIZE])` — 当 `local_data` 是 `Tensor` 时从普通输入发射一个 Tile；已是 Tile 时跳过
-- Phase 1：`tile.store(stage_tile, [0, 0], target)` — 将本 rank 的 chunk 写入私有 HCCL 窗口的本地行 0
+- Phase 1：对 `peer` 从 `0` 到 `NR-1`，`pld.tile.put(target, peer, stage_tile, stage, [my_rank, 0], [0, 0], [1, SIZE])` — 将本 rank 的 chunk 推送到每个 peer 窗口的第 `my_rank` 行。自推送 (`peer == my_rank`) 通过 HCCL 恒等映射实现。`pld.tile.put` 在 SIZE 超过 staging tile 容量时自动分块
```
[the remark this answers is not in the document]
georgebisbas: Fixed in commit 386252a: Updated the Chinese docs to match the English — replaced the stale tile.load Phase 0 with tile.create description, and updated Phase 1 to reflect local_data passed directly to pld.tile.put.

python/pypto/ir/op/distributed/tensor_ops.py
```diff
@@ -302,33 +302,46 @@ def allgather(
...
+    **HOST builtin:** ``allgather(local_data, target, signal=None)`` — each rank's
+    chunk is pre-staged in the window-bound ``local_data`` ([NR, SIZE]
+    DistributedTensor), ``target`` is the INT32 DistributedTensor signal.
+    The 2-arg convenience form expands to 3-arg IR internally.
+
+    **InCore composite:** ``allgather(local_data, target, signal)`` —
+    push-based: each rank pushes its chunk into every peer's window via
+    ``pld.tile.put``, then notify/wait barrier; the window itself becomes
+    the gathered [NR, SIZE] result (window-as-result). Lowered by
+    LowerCompositeOps into tile.load + pld.tile.put loop + notify/wait.
```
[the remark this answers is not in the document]
georgebisbas: Fixed in commit 386252a: Updated docstring from tile.load + pld.tile.put to tile.create(stage) + pld.tile.put, matching the actual lowering which does not emit an explicit tile.load.

python/pypto/language/distributed/op/tensor_ops.py
YunjiQin: the 2 arags host builtin still exist. don't keep it.

python/pypto/language/distributed/op/tensor_ops.py
```diff
@@ -603,8 +630,6 @@ def reduce_scatter(
 
     ``target`` has shape [NR, SIZE] — one row per chunk.  Each rank must
     stage all NR chunks before calling::
-
-        for j in range(nranks):
             data = pl.store(chunk_j, [j, 0], data)
```
YunjiQin, line 610: fix this doc's indent

python/pypto/language/distributed/op/tensor_ops.py
```diff
@@ -546,51 +546,47 @@ def broadcast(
...
+    ``builtin.tensor.barrier`` per chip to synchronise the pre-staged data.
+
+    The C++ deducer validates ``target`` and ``signal`` by type and
+    accepts ``Tile``, ``Tensor``, or ``DistributedTensor`` for ``local_data``;
+    the lowering passes handle path-specific validation per their context.
+
+    Args:
+        local_data: InCore: :class:`pl.Tensor` [1, SIZE] with this rank's
+            chunk.  HOST: :class:`pl.Tensor` [1, SIZE] or
+            :class:`pld.DistributedTensor` [NR, SIZE] pre-staged window.
```
YunjiQin: HOST should only keep the `pl.Tensor` [1, SIZE] for local_data. don't accept `pld.DistributedTensor` [NR, SIZE] pre-staged window. each rank also don't need to pre-stage self data to window in HOST.

src/ir/transforms/lower_composite_ops_pass.cpp
```diff
@@ -962,48 +962,47 @@ ExprPtr LowerTensorBroadcastRule(const CallPtr& call, const std::vector<ExprPtr>
...
-  // handles the load internally).  Record which so we can emit tile.load
-  // after the shared setup expressions are available.
-  bool is_tensor_input = (As<TensorType>(local_data->GetType()) != nullptr);
-  INTERNAL_CHECK_SPAN(As<TileType>(local_data->GetType()) || is_tensor_input, span)
-      << "pld.tensor.allgather local_data must be TileType or TensorType, got "
+
+  // local_data may be a Tensor or Tile at this point — the default pass pipeline
+  // runs ConvertTensorToTileOps before LowerCompositeOps, so local_data may
+  // already be converted to a TileType.
+  INTERNAL_CHECK_SPAN(As<TensorType>(local_data->GetType()) || As<TileType>(local_data->GetType()), span)
```
YunjiQin: **`local_data`'s Tile contract is broken in the push rewrite.**

Phase 1 below passes `local_data` straight to `pld.tile.put` as `src`, but that op's `src` only accepts a Tensor/DistributedTensor — `AsTensorTypeLike` excludes `TileType` (deducer `put.cpp:109` + codegen `pto_ops_distributed.cpp:482`, both user-facing `CHECK`). So a `TileType` `local_data` passes this `INTERNAL_CHECK` but then crashes downstream inside `pld.tile.put`, with the error surfacing at a different op — hard to diagnose.

The old pull-based rule genuinely supported Tile (it went through `tile.load`/`tile.store`, which accept tiles); dropping those in the push rewrite silently broke Tile support while the C++ op description and Python docstrings still advertise "Tile/Tensor [1, SIZE]".

Also the comment on the line above ("...so local_data may already be converted to a TileType") is misleading: if `ConvertTensorToTileOps` actually produced a Tile here, the push path would crash in `pld.tile.put`. ST passing implies `local_data` stays a Tensor.

Suggestion: drop the `|| As<TileType>(...)` branch (InCore accepts only `TensorType`), fix the comment, and update the op description + both docstrings to drop the "Tile" wording.

src/ir/transforms/lower_composite_ops_pass.cpp
```diff
@@ -1491,10 +1472,11 @@ class LowerCompositeOpsMutator : public IRMutator {
...
   [[nodiscard]] static bool ShouldSkipHostCollective(const CallPtr& call) {
     if (!call || !call->op_) return false;
-    // pld.tensor.allgather overloads: skip only the 2-arg HOST builtin form;
-    // the 4-arg InCore composite form must still be lowered by this pass.
+    // pld.tensor.allgather unified 3-arg: skip the HOST builtin path
+    // (args[0] is a DistributedTensor — pre-staged window); the InCore
+    // composite path (args[0] is Tile/Tensor) must still be lowered here.
     if (IsOp(call, "pld.tensor.allgather")) {
-      return call->args_.size() == 2;
+      return call->args_.size() == 3 && As<DistributedTensorType>(call->args_[0]->GetType()) != nullptr;
```
YunjiQin: **HOST vs InCore should be decided by the caller's function context, not by arg[0]'s type.**

This dispatch is already two-level: the outer `skip_host_collectives_` is decided authoritatively by the caller's level/role (see `TransformLowerCompositeOps`: `Level::HOST && (Orchestration || Orchestrator)`), and `ShouldSkipHostCollective` is only consulted when that is already true (L1486). The other four collectives just `return true` unconditionally here.

Re-deriving the context from `As<DistributedTensorType>(args[0])` is redundant and strictly less robust: inside a host-orch function, an allgather whose arg[0] isn't a DT would wrongly fall back to composite lowering. Suggest folding allgather into the same unconditional list as the others:

```cpp
return IsOp(call, "pld.tensor.allgather") || IsOp(call, "pld.tensor.allreduce") ||
       IsOp(call, "pld.tensor.barrier")   || IsOp(call, "pld.tensor.broadcast") ||
       IsOp(call, "pld.tensor.reduce_scatter");
```

src/ir/transforms/convert_tensor_to_tile_ops_pass.cpp
```diff
@@ -1028,24 +1030,32 @@ void AnalyzeCallAccess(const CallPtr& call, const AliasOriginMap& origin_map, st
...
-      // out: write only
-      MarkAccess(CollectReferencedOrigins(call->args_[3], origin_map), has_write);
+    // Unified 3-arg:
+    //   HOST (args[0] is DistributedTensor): args[0]=pre-staged window (InOut),
+    //     args[1]=signal (InOut), args[2]=unused.
+    //   InCore (args[0] is Tile/Tensor): args[0]=local_data (In, read-only),
+    //     args[1]=target window (InOut, push target and result),
+    //     args[2]=signal (InOut, barrier).
+    if (call->args_.size() >= 3) {
+      if (As<DistributedTensorType>(call->args_[0]->GetType())) {
```
YunjiQin: **This HOST branch is dead code.**

`AnalyzeCallAccess` is reached only via `TransformIncoreFunction`, which the program-pass driver invokes solely for `func_type_ == FunctionType::InCore` (L2007 in this file). Inside an InCore function, an allgather's arg[0] is `local_data` (Tile/Tensor), never a DistributedTensor, so this `As<DistributedTensorType>(call->args_[0])` HOST branch can never execute.

Same root cause as the `ShouldSkipHostCollective` comment in `lower_composite_ops_pass.cpp`: HOST vs InCore is a function-context property, not an arg[0]-type property. Suggest deleting the HOST branch and unconditionally applying the InCore direction inference (local_data=In; target/signal=InOut).

python/pypto/language/distributed/op/tensor_ops.py
georgebisbas: Done in e645c2b — the two stale 2-arg references in collective.cpp are now removed:
- File header comment: merged the two allgather entries (2-arg HOST + 3-arg push) into a single (unified 3-arg) entry.
- builtin.tensor.allgather deducer: removed the 2-arg conditional path (args.size() == 2 ? 0 : 1 target/signal dispatch). Now requires exactly 3 args with fixed positions.
The Python DSL and IR builder were already strict 3-arg only.

src/ir/transforms/convert_tensor_to_tile_ops_pass.cpp
georgebisbas: Fixed in 54afce59 — removed the HOST dispatch. AnalyzeCallAccess is only reached via TransformIncoreFunction, so the branch checking As<DistributedTensorType>(args[0]) was dead code. Now unconditionally treats all three args as InCore (local_data=In, target/signal=InOut).

src/ir/transforms/lower_composite_ops_pass.cpp
georgebisbas: Fixed in 54afce59 — allgather is now unconditional like the other four collectives. The outer skip_host_collectives_ flag already determines context by caller level/role; re-inferring from arg[0] type was redundant and fragile.

src/ir/transforms/lower_composite_ops_pass.cpp
georgebisbas: Fixed in 54afce59 — removed || As<TileType> from the INTERNAL_CHECK, corrected the misleading comment about ConvertTensorToTileOps producing Tile, and updated the REGISTER_OP + both Python docstrings to remove "Tile" mentions. pld.tile.put's src only accepts Tensor/DistributedTensor — the old pull-based path genuinely supported Tile via tile.load/tile.store, but the push rewrite dropped those steps.

python/pypto/language/distributed/op/tensor_ops.py
georgebisbas: Fixed in 54afce59 — removed "DistributionTensor [NR, SIZE]" from HOST local_data descriptions in the REGISTER_OP, IR builder, and DSL docstrings. HOST now only advertises Tensor [1, SIZE].

src/ir/op/distributed/collective.cpp
```diff
@@ -219,93 +218,68 @@ REGISTER_OP("pld.tensor.broadcast")
...
-  CHECK(out_type) << "pld.tensor.allgather out must be a Tensor (not a DistributedTensor), got "
-                  << args[3]->GetType()->TypeName();
-  CHECK(out_type->shape_.size() == 2)
-      << "pld.tensor.allgather out must be 2D [1, NR*SIZE], got " << out_type->shape_.size() << " dims";
-
-  // Return the output Tensor type — the intrinsic writes directly into it.
-  return out_type;
+
+  // Both paths return target's DistributedTensorType (window-as-result).
+  return target_type;
```
YunjiQin, line 269: **Lost validation: `signal` first dim must equal NR.**

The old 4-arg deducer checked `AreExprsEqual(signal_type->shape_[0], target_type->shape_[0])`; the unified deducer dropped it. The notify/wait barrier indexes `signal` per rank (`0..NR-1`), so an undersized `signal` now passes type deduction and only faults at runtime inside `EmitNotifyAll`/`EmitWaitAll` (the InCore path has no other static check on the barrier size — HOST still has `CheckStaticSignalCapacity`). The test change from `signal[SIZE]` -> `signal[NR]` confirms this constraint is real.

Suggest restoring it right before `return target_type;`:

```cpp
// The notify/wait barrier indexes `signal` per rank (0..NR-1), so its first
// dimension must equal target's NR.  Without this guard an undersized signal
// passes type deduction and only faults at runtime inside EmitNotifyAll /
// EmitWaitAll (InCore path has no other static check on the barrier size).
CHECK(AreExprsEqual(signal_type->shape_[0], target_type->shape_[0]))
    << "pld.tensor.allgather signal first dimension must equal target first "
       "dimension (NR)";
```

src/ir/transforms/lower_composite_ops_pass.cpp
```diff
@@ -962,49 +962,47 @@ ExprPtr LowerTensorBroadcastRule(const CallPtr& call, const std::vector<ExprPtr>
...
-  // after the shared setup expressions are available.
-  bool is_tensor_input = (As<TensorType>(local_data->GetType()) != nullptr);
-  INTERNAL_CHECK_SPAN(As<TileType>(local_data->GetType()) || is_tensor_input, span)
-      << "pld.tensor.allgather local_data must be TileType or TensorType, got "
-      << local_data->GetType()->TypeName();
+
+  // local_data is always a Tensor at this point.  pld.tile.put only accepts
+  // Tensor/DistributedTensor for its src parameter; a TileType would crash
+  // downstream in pld.tile.put's own deducer/codegen CHECKs.
+  INTERNAL_CHECK_SPAN(As<TensorType>(local_data->GetType()), span)
```
YunjiQin: **Wrong macro for a user-facing contract + docstring over-claims.**

`local_data` is user-provided (the DSL accepts `Tensor | DistributedTensor`) and the deducer *intentionally* no longer validates it ("validated by the lowering passes per their context"). So this is a user-facing contract check, not a compiler invariant — per `error-checking.md` it should be `CHECK_SPAN`, not `INTERNAL_CHECK_SPAN`. A `DistributedTensor` local_data on the InCore path currently crashes as an "Internal error" instead of a clean `ValueError`.

```cpp
// local_data is user-provided (DSL allows Tensor | DistributedTensor) and the
// deducer defers its validation to the lowering passes, so this is a user-facing
// contract check -> CHECK_SPAN. The InCore push path only supports a plain
// Tensor source: pld.tile.put reads its `src` via AsTensorTypeLike.
CHECK_SPAN(As<TensorType>(local_data->GetType()), span)
    << "pld.tensor.allgather local_data must be a plain Tensor [1, SIZE] on the "
       "InCore path, got " << local_data->GetType()->TypeName();
```

Also the `allgather` DSL docstring in `python/pypto/language/distributed/op/tensor_ops.py` says the deducer "accepts `Tile`, `Tensor`, or `DistributedTensor` for `local_data`" — that's now inaccurate; the InCore path requires a plain `Tensor`. Please tighten the docstring to match.

YunjiQin (commented):
Two follow-ups on the unified 3-arg allgather — both are validation gaps introduced by the rewrite. Inline comments below.

src/ir/op/distributed/collective.cpp
georgebisbas: Done. Restored the `AreExprsEqual(signal_type->shape_[0], target_type->shape_[0])` guard with a `CHECK` right before `return target_type`, with the comment explaining that the notify/wait barrier indexes signal per rank so the first dimension must match target NR. CI 11/11 green.

src/ir/transforms/lower_composite_ops_pass.cpp
georgebisbas: Done. Changed `INTERNAL_CHECK_SPAN` to `CHECK_SPAN` for the `local_data` check in `lower_composite_ops_pass.cpp` (user-facing contract, not compiler invariant per error-checking.md). DSL docstring in `tensor_ops.py` tightened to accurately describe path-specific validation: InCore requires plain Tensor, HOST accepts either. CI 11/11 green.

src/ir/op/distributed/collective.cpp
georgebisbas: Follow-up on the AreExprsEqual guard: after restoring the check, the HOST distributed ST (test_l3_host_tensor_allgather.py) failed with this exact assertion because signal.shape_[0] and target.shape_[0] come from separate pld.world_size() invocations at different pld.window() call sites — each produces a distinct ir::Call pointer.

Root cause: AreExprsEqual handled ConstInt (by value) and BinaryExpr (structurally), but fell through to return false for Call nodes — only pointer identity was accepted. The old InCore tests never hit this because they used a single NR = pl.dynamic("world_size") whose DynVar.unwrap() is idempotent (same Var pointer), whereas the HOST test's separate pld.world_size() calls produce different Call pointers for the same semantic value.

Fix (0d8b1f30): extended AreExprsEqual with structural Call comparison — same op_ pointer (registry singleton per op name) plus recursively equal positional args. This follows the existing BinaryExpr structural comparison pattern and is sound because OpPtr comes from a singleton registry. 7280 UT passed (sim Docker).

src/ir/expr.cpp
```diff
@@ -66,6 +66,19 @@ bool AreExprsEqual(const ExprPtr& e1, const ExprPtr& e2) {
   if (b1 && b2 && e1->GetKind() == e2->GetKind()) {
     return AreExprsEqual(b1->left_, b2->left_) && AreExprsEqual(b1->right_, b2->right_);
   }
+  // Call nodes produced by different invocation sites have different pointer
+  // identities but may be structurally identical (e.g. two pld.world_size()
+  // calls). Compare by op pointer identity (registry singleton per op name)
+  // plus recursive arg equality.
+  auto call1 = As<Call>(e1);
+  auto call2 = As<Call>(e2);
+  if (call1 && call2 && call1->op_ == call2->op_) {
```
YunjiQin: **Op compared by pointer here violates `operator-identity-checks.md` ("Match by name, not by pointer").**

`call1->op_ == call2->op_` relies on pointer identity, but the rule is explicit that two `Op`s sharing a name can be *distinct pointers* — the `.pto` deserializer and the MemRef alloc builders each construct their own `Op`. For structurally-identical Calls whose ops came from different construction sites (e.g. after a `.pto` round-trip), this returns a false negative. Since `AreExprsEqual` is a core utility used broadly (e.g. `TensorValue` `valid_shape`/`stride` comparison in `type.cpp`), that surfaces as spurious type-inequality post-deserialization.

It works today only because both operands are registry-sourced in-memory (same singleton pointer). Suggest matching by canonical name instead:

```cpp
if (call1 && call2 && call1->op_ && call2->op_ &&
    call1->op_->name_ == call2->op_->name_) {
```

and mirror it in `HashExprForAreExprsEqual` (hash the op *name*, not `call->op_.get()`), so equality and hash stay consistent across deserialized/synthesized ops. The equality/hash pair is internally consistent right now — this is about surviving distinct-pointer same-name ops, not a hash-map bug.

YunjiQin (commented):
Thanks — the signal-NR check and the `local_data` CHECK_SPAN are both correctly addressed. One note on the connected `AreExprsEqual` extension that the NR check pulled in.

src/ir/expr.cpp
georgebisbas: Matching by name applied

YunjiQin (commented):
LGTM ✅

All three substantive findings from the earlier review are resolved:

1. **signal-NR check** restored in `DeduceTensorAllGatherType` (`CHECK(AreExprsEqual(signal[0], target[0]))`) — undersized barriers now caught at type-deduction time.
2. **`local_data` validation** switched from `INTERNAL_CHECK_SPAN` to `CHECK_SPAN` (user-facing contract), and the DSL docstring tightened to state the InCore path requires a plain `Tensor [1, SIZE]`.
3. **`AreExprsEqual` Call branch** now matches ops by canonical name (`op_->name_`) instead of pointer identity, with `HashExprForAreExprsEqual` synced to hash the name — no false negatives across `.pto` round-trips / pass-synthesized ops, and equality/hash stay consistent.

The push-based rewrite is clean and well-tested (docs EN+ZH updated, per-pass tests adjusted). Nice work.

YunjiQin (approved):
Approving — all three review findings resolved (signal-NR check, local_data CHECK_SPAN + docstring, AreExprsEqual by-name op match with hash synced). Push-based allgather is clean and well-tested. 🚀

PR #1984, opened 2026-07-08
fix(ir): remap device= attr in IRMutator Call visitor for P=1 unroll

### LINKED ISSUE #1983: [Pass Bug] IRMutator drops kAttrDevice remap during loop unrolling (P=1 host-orch codegen)
### Pass Name

IRMutator (`IRMutator::VisitExpr_(const CallPtr&)` in `src/ir/transforms/mutator.cpp`) — the generic attribute-remapping logic that `UnrollLoops` relies on when specializing a single-iteration `pl.range` host-orchestration loop.

### Description

Reported by @vloncar while working on a more advanced example of distributed linear attention.
Reproduced by @georgebisbas.

When a `host_orch` (`pl.Level.HOST`, `pl.Role.Orchestrator`) function contains a loop `for r in pl.range(P): ... self.chip_orch(..., device=r)` and `P == 1`, the loop gets unrolled (single iteration), substituting the loop variable `r` with the literal `0` via `DeepClone`.

`IRMutator::VisitExpr_(const CallPtr&)` re-mutates a whitelist of `Var`-typed dependency attributes during this substitution, but the whitelist omits `kAttrDevice` — an `ExprPtr`-valued attribute holding the device expression (`device=r`). As a result, every whitelisted attribute and all call arguments get correctly rewritten to reference the literal `0`, but `kAttrDevice` keeps pointing at the now-stale/unbound `r` variable.

This produces IR — and consequently generated `host_orch.py` orchestration code — that references an undefined Python variable, crashing at runtime with a `NameError`.

The same program compiled with `P == 2` (no unrolling, loop variable stays a real loop-carried var) works correctly, which is what makes this specifically a **loop-unrolling / single-iteration specialization** bug rather than a general `device=` attribute bug.

### Git Commit ID

10d6b12

### Before IR (Input)

```python
import pypto.language as pl

@pl.program
class Before:
    @pl.function(type=pl.FunctionType.Orchestration)
    def chip_orch_stage2(
        self,
        Q: pl.Tensor[[L, DK], pl.FP32],
        # ... other per-chip tensor args ...
        O: pl.Out[pl.Tensor[[L, DV], pl.FP32]],
    ) -> pl.Tensor[[L, DV], pl.FP32]:
        return self.gla_stage2(Q, ..., O)

    @pl.function(level=pl.Level.HOST, role=pl.Role.Orchestrator)
    def host_orch(
        self,
        Q: pl.Tensor[[P, L, dk], pl.FP32],
        # ... other per-chip tensor args ...
        O: pl.Out[pl.Tensor[[P, L, dv], pl.FP32]],
    ) -> pl.Tensor[[P, L, dv], pl.FP32]:
        for r in pl.range(P):          # P == 1 -> single-iteration loop, gets unrolled
            Q_r = Q[r]
            O_r = O[r]
            self.chip_orch_stage2(Q_r, ..., O_r, device=r)   # device attr references loop var `r`
        return O
```

Full runnable repro (compiled with `ir.compile(program, platform="a2a3sim", distributed_config=DistributedConfig(device_ids=[6], num_sub_workers=0))`) is available on request — attaching the full script here would exceed the template's scope, but the loop/`device=r` shape above is the exact reproducing pattern (`L, C, dk, dv = 32, 16, 16, 16`, `P = 1`).

After `UnrollLoops` specializes the single-iteration loop, **all** references to `r` — including the `kAttrDevice` attribute on the `Submit`/`Call` node — should be substituted with the literal `0`:

```python
@pl.program
class Expected:
    @pl.function(level=pl.Level.HOST, role=pl.Role.Orchestrator)
    def host_orch(self, Q, ..., O) -> pl.Tensor[[P, L, dv], pl.FP32]:
        Q_r = Q[0]
        O_r = O[0]
        self.chip_orch_stage2(Q_r, ..., O_r, device=0)   # device attr correctly folded to literal 0
        return O
```

Generated `host_orch.py` should emit, e.g.:

```python
_submit_chip(orch, callables["chip_orch_stage2"], _ta_0, config, 0)
```

### Actual IR/Error

Generated `host_orch.py` keeps the stale unbound loop variable in the device slot:

```python
_submit_chip(orch, callables["chip_orch_stage2"], _ta_0, config, r__idx_v0)
```

Running it raises:

```text
NameError: name 'r__idx_v0' is not defined
```

### NPU Kind (if relevant)

Reproduces on the `a2a3sim` simulator platform and a2a3 910B

N/A (not hardware-specific)

### Additional Context

- `P == 2` (`--devices 6,7`, no unrolling) compiles and runs correctly — confirms this is specific to the single-iteration unroll path, not a general `device=` attribute bug.
- **Fix identified and verified**: in `IRMutator::VisitExpr_(const CallPtr&)` (`src/ir/transforms/mutator.cpp`), add an `else if (k == kAttrDevice)` branch — alongside the existing whitelisted `Var` attrs — that re-visits the device `ExprPtr` through `ExprFunctor<ExprPtr>::VisitExpr(*dev)` and sets `attrs_changed = true` when it changes. After applying this fix and rebuilding, the P=1 repro passes (`max|O-golden| = 8.583e-06`), and generated `host_orch.py` correctly emits the literal `0` instead of `r__idx_v0`.

georgebisbas:
## Summary

Fixes a bug where `IRMutator::VisitExpr_(const CallPtr&)` drops the `kAttrDevice` remap during loop unrolling when `P == 1` in host-orchestration codegen.

### Root Cause

When a `host_orch` dispatches a chip kernel inside `for r in pl.range(P)` with `P == 1`, the `UnrollLoops` pass unrolls the single-iteration loop, substituting `r` → `0` via `DeepClone`. The `IRMutator` re-mutated a whitelist of `Var`-typed dependency attrs during this substitution but **omitted `kAttrDevice`** — the `ExprPtr`-valued `device=r` attribute. As a result, slice subscripts correctly folded to `0` but `device=` kept the now-undefined loop variable, causing `NameError: name 'r__idx_v0' is not defined` at dispatch.

### Fix

Add a `kAttrDevice` remap branch in `IRMutator::VisitExpr_(const CallPtr&)` that visits the device `ExprPtr` through `VisitExpr` alongside the call arguments, mirroring the existing `kAttrDevice` handling in the SSA pass. The branch only fires for calls carrying a `device=` attr (distributed host-orch dispatch), so non-distributed code is byte-identical.

### Changes

- `src/ir/transforms/mutator.cpp`: Add `else if (k == kAttrDevice)` branch to remap the device expression during mutation
- `tests/ut/ir/transforms/test_unroll_loops_pass.py`: Add `TestCallAttrSubstitution` regression test

### Testing

- **Regression test**: `TestCallAttrSubstitution` PASSES (verifies `device=r` → `device=0` after single-iteration unroll)
- **Full unit test suite**: 6424 passed, 2 skipped, 0 failures (simulation Docker, gcc-15)
- `P >= 2` (no unrolling) unchanged — no regression

### Verification

```
# P=1 (was broken, now fixed):
_submit_chip(orch, callables["chip_orch_stage2"], _ta_0, config, 0)  # ← literal 0, not r__idx_v0

# P=2 (unchanged):
_submit_chip(orch, callables["chip_orch_stage2"], _ta_0, config, r__idx_v0)  # ← real loop var, correctly bound
```

Fixes #1983

Co-authored-by: Vladimir Loncar <vloncar@users.noreply.github.com>

### DIFF
--- a/src/ir/transforms/mutator.cpp
+++ b/src/ir/transforms/mutator.cpp
@@ -385,6 +385,24 @@ ExprPtr IRMutator::VisitExpr_(const CallPtr& op) {
           continue;
         }
       }
+    } else if (k == kAttrDevice) {
+      // The distributed dispatch ``device=`` attr holds a single ExprPtr (a
+      // ConstInt rank or a Var loop index). It references a Var defined
+      // elsewhere (the host-orch ``for r in pl.range(P)`` loop var), so it must
+      // be remapped alongside the args — otherwise a loop-var substitution
+      // (e.g. the P==1 unroll folding ``r`` -> ``0``) rewrites the slice
+      // subscripts but leaves ``device=r`` dangling to the now-undefined loop
+      // var, and codegen emits an unbound index -> NameError. Mirrors the
+      // kAttrDevice handling in the SSA pass.
+      const auto* dev = std::any_cast<ExprPtr>(&v);
+      if (dev && *dev) {
+        auto new_dev = ExprFunctor<ExprPtr>::VisitExpr(*dev);
+        if (new_dev.get() != dev->get()) {
+          attrs_changed = true;
+          new_attrs.emplace_back(k, std::any(std::move(new_dev)));
+          continue;
+        }
+      }
     }
     new_attrs.emplace_back(k, v);
   }
--- a/src/ir/transforms/mutator.cpp
+++ b/src/ir/transforms/mutator.cpp
@@ -397,6 +397,7 @@ ExprPtr IRMutator::VisitExpr_(const CallPtr& op) {
       const auto* dev = std::any_cast<ExprPtr>(&v);
       if (dev && *dev) {
         auto new_dev = ExprFunctor<ExprPtr>::VisitExpr(*dev);
+        INTERNAL_CHECK_SPAN(new_dev, op->span_) << "Call device attribute mutated to null";
         if (new_dev.get() != dev->get()) {
           attrs_changed = true;
           new_attrs.emplace_back(k, std::any(std::move(new_dev)));
--- a/tests/ut/ir/transforms/test_unroll_loops_pass.py
+++ b/tests/ut/ir/transforms/test_unroll_loops_pass.py
@@ -275,6 +275,56 @@ class TestPrinterRoundTrip:
         assert "pl.unroll(" in printed
 
 
+class TestCallAttrSubstitution:
+    """Tests that Call attrs referencing the loop var are substituted on unroll."""
+
+    def test_device_attr_substituted_on_single_iteration_unroll(self):
+        """A ``device=`` attr (kAttrDevice) referencing the loop var must be
+        substituted with the literal when a single-iteration loop is unrolled.
+
+        Regression test: ``IRMutator::VisitExpr_(const CallPtr&)`` re-mutated a
+        whitelist of ``Var``-typed dependency attrs during loop-var substitution
+        but omitted ``kAttrDevice`` (an ``ExprPtr``-valued attr). At P==1 in a
+        host-orchestration loop, this left ``device=r`` dangling to the
+        now-unbound loop var after unrolling folded ``r`` -> ``0``, producing a
+        ``NameError`` in generated orchestration code.
+        """
+
+        SIZE = 8
+
+        @pl.program
+        class Before:
+            @pl.function(type=pl.FunctionType.Orchestration)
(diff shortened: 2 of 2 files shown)


YunjiQin (approved):
LGTM. Correct fix: remaps kAttrDevice through VisitExpr so P=1 unroll no longer leaves a dangling loop var in the device= attr. CI green, regression test covers the fold to device=0.

PR #2026, opened 2026-07-13
fix(jit): guard un-inferrable tuple/scalar return elements (#2016 follow-up)

lyfne123:
## Summary

Follow-up to #2016 (`b2f347fd`). That PR taught `_infer_return_type` to fill
un-inferrable tuple-return elements — a local with no `TensorMeta` (a TASK_ID
from `pl.spmd`, or `logits` unpacked from a dep call) — from the function's
**explicit** return annotation, copied **verbatim** (`ast.unparse`). Two cases
were unguarded and started failing at parse time (surfaced when kernels adopted
`jit_fn.compile(...)` for host wrappers, e.g. pypto-lib qwen3 decode):

- **Bare** element (`pl.Tensor`, no shape) → `ParserTypeError: Incomplete type annotation: pl.Tensor`.
- **Symbolic-shaped** element (`pl.Tensor[[BATCH, VOCAB], pl.FP32]`) → `NameError: name 'BATCH' is not defined`.

## Fix

- **Copy an explicit element only when it is a *complete* (subscripted) type**
  (`pl.Scalar[pl.TASK_ID]`, `pl.Tensor[[...], ...]`), folding module
  int/float/bool constants in its shape to literals (`[BATCH, VOCAB]` →
  `[16, 152064]`) so symbolic dims become concrete extents.
- **Drop the whole annotation only when it is dispensable** — a bare element is
  present AND no sibling element *requires* the annotation.

### Why dropping is safe (and when it is not)

A plain tensor return is re-inferable by the parser from its IR value once the
annotation is gone — the pre-#2016 behavior that worked for years. A **TASK_ID**,
by contrast, the parser **cannot** re-derive from a value; that is exactly why
#2016 added the copy. So the drop is gated:

| Return | Behavior |
| ------ | -------- |
| all-tensor bare tuple (qwen3 host wrapper) | annotation dropped → parser re-infers → compiles |
| Out-param + `pl.Scalar[pl.TASK_ID]` (#2016) | kept (TASK_ID preserved) |
| bare tensor-local + `TASK_ID` (the risky mix) | **not** dropped — TASK_ID kept; the bare element surfaces as a clear `Incomplete type annotation` the author must shape, instead of silently stripping the TASK_ID |

## Testing

- New unit tests in `tests/ut/jit/test_specializer.py`: bare element → dropped;
  bare element next to a TASK_ID → annotation kept; subscripted element with
  symbolic module constants → folded to literals; no `py_globals` → left symbolic.
- `test_specializer.py`: 104 passed. Full `tests/ut/jit/` passes (2 pre-existing
  `test_split_k` numerical failures reproduce on `main` without this change).
  ruff clean.

## Impact

Once merged, host-orchestrator kernels that return locals with (bare) tuple
annotations — e.g. pypto-lib qwen3 `qwen3_decode_host`, which blocks
pypto-serving #53 and pypto-lib #765 — compile again with **no kernel change**.

## Related

- Regression introduced by #2016 (`b2f347fd`, 2026-07-13).

### DIFF
--- a/python/pypto/jit/specializer.py
+++ b/python/pypto/jit/specializer.py
@@ -35,6 +35,7 @@ other_jit_func(a, b)              self.other_jit_func(a, b)  (multi-function onl
 from __future__ import annotations
 
 import ast
+import copy
 import functools
 import inspect
 import textwrap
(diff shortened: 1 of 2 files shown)


Hzfengsy (approved):
LGTM. Verified the un-inferrable-element handling: `annotation_required` correctly keeps a sibling `Scalar[TASK_ID]` (which the parser cannot re-derive from a value) while a lone bare element still drops the annotation back to pre-#2016 behavior. The `copy.deepcopy` in `_fold_const_names` correctly avoids mutating the `func_def` shared with the body specialization pass. Good test coverage on all four paths.

Re: the bot's `isinstance(value, (int, float, bool))` note — not an issue here: `DataType` is a nanobind-bound C++ class, not an int-backed enum (`isinstance(pl.FP32, (int, float, bool))` is `False`), and dtypes render as `ast.Attribute` (`pl.FP32`) which `visit_Name` never visits. A strict `type(value) in (...)` check would be marginally more defensive, but not blocking.

Issue #2027, opened 2026-07-13
[Feature] Support dynamic DistributedTensor context merges in PTO control flow


Hzfengsy:
### Summary

Support dynamically selecting a `DistributedTensor` across control-flow branches by carrying its base pointer and `CommContext` as synchronized SSA results. Until this is implemented, PTO codegen should continue rejecting merges whose branches yield different communication contexts.

### Motivation / Use Case

A direct alias of a distributed tensor can safely inherit the source tensor view, base pointer, and communication context. The same is true for loop carriers and `if` branches that refer to the same distributed allocation.

A genuinely dynamic selection is different:

```python
result = lhs
if cond:
    result = rhs
pld.tile.remote_load(result, peer=peer, offsets=[0, 0], shape=[16, 16])
```

When `lhs` and `rhs` belong to different communication domains, selecting only one static context can address the wrong remote window. Current PTO codegen therefore rejects this merge instead of silently miscompiling it.

Supporting the use case requires the selected data pointer and selected context to follow the same control-flow edge.

### Proposed API / Behavior

Keep the DSL unchanged. Lower a conflicting distributed-tensor merge conceptually as a paired control-flow result:

```mlir
%base, %ctx = scf.if %cond -> (!pto.ptr<i32>, !pto.ptr<i64>) {
  scf.yield %rhs_base, %rhs_ctx
} else {
  scf.yield %lhs_base, %lhs_ctx
}
%view = pto.make_tensor_view %base ...
```

The implementation should:

- Select the tensor base pointer and `CommContext` together.
- Reconstruct the statically typed tensor view after the merge.
- Preserve tensor offsets and view metadata for aliases of subviews.
- Extend the same paired-carry model to loops if loop iterations may switch allocations.
- Diagnose any path where only the tensor or only the context is merged.
- Retain the current same-context fast path and conservative rejection until the paired lowering is complete.

### Alternatives Considered

- Keep one branch context unconditionally: unsafe because it can route remote access through the wrong communication window.
- Select only the context: insufficient because the base pointer must follow the same branch.
- Yield the tensor view directly through `scf.if`: currently loses static dimensions needed by `pto.partition_view`; see #1533.
- Permanently prohibit cross-allocation merges: safe but prevents useful runtime selection patterns.

### Additional Context

Related: #1533, #1956

[...]

Issue #1956, opened 2026-07-06
[Feature] Explicit `alloc_tile` IR op so codegen is 1:1 (root-fix for PTOAS if/else-yield miscompile)


lyfne123:
> **Superseded by #2693.** The additive `alloc_tile`/MemRef design below has been replaced by the final Tile-to-Buffer architecture. Design discussion, M2–M5 implementation milestones, and final acceptance are tracked in #2693; M1 is in #2666. Closing this issue as `not planned` consolidates tracking and does not mean the migration is implemented.

---

### Summary

Promote the PTO **tile handle** (`pto.alloc_tile`) from a codegen-synthesized artifact to a **first-class IR op** (`alloc_tile`), so PTO codegen becomes a strict 1:1 emitter and tile-buffer *placement / aliasing* decisions move out of codegen into the IR/passes. This is the root fix for a confirmed `memory_planner=PTOAS` miscompile and removes a whole class of codegen-scope hacks.

### Motivation / Use Case

**Confirmed correctness bug** (found while extending `memory_planner=PTOAS`, added in #1934): `compile(memory_planner=MemoryPlanner.PTOAS)` **silently miscompiles any `if/else`-yield (phi) kernel**.

Minimal repro — `if c: r1 = mul(ta, ta); yield r1  else: r2 = add(ta, ta); yield r2`, then store the phi result:

- **PYPTO / level3 (correct):** `r1` and `res` share addr `16384`; the else branch inserts a `TMOV` copying `r2` into `res`'s addr → the store reads the right data.
- **PTOAS / level2 (WRONG):** `res` gets its own addr `32768`; the then-branch writes `r1 @16384`, the else writes `r2 @32768`, **no `TMOV`** → taking the `then` branch stores `res @32768`, which was **never written** → garbage. Confirmed by compiling the `.pto` with `ptoas` and reading the CCE (`TASSIGN` addresses + missing `TMOV`); at the MLIR level ptoas rejects it as `use of undeclared SSA value name` once handles are shared in-branch.

Loop-carried accumulators are already correct; the hole is must-alias that needs a codegen-emitted `tile.move` or a return_var handle-merge: **if/else-yield**, and **for-loops whose yield buffer diverges from init**.

**Root cause — two levels, but only one is on the IR:**

[...]

PR #2028, opened 2026-07-13
fix(codegen): Preserve distributed alias communication context

Hzfengsy:
## Summary

- Preserve each `DistributedTensor` communication-context binding through direct aliases, `if` merges that retain one allocation, and `for` / `while` carriers.
- Fail early with construct-specific internal diagnostics when distributed tensor metadata loses its `CommContext`.
- Reject `if` branches that merge different context SSAs, because choosing either static context can pair the other branch data pointer with the wrong remote-window table.
- Add regression coverage for direct alias propagation and conflicting-context rejection.
- Apply the configured brace style to existing single-line loops in the touched control-flow file so diff-scoped clang-tidy passes.

## Why

PTO codegen represents a distributed tensor with three coordinated pieces of state: its tensor view, backing base pointer, and a side-table mapping to the runtime `CommContext` pointer. Alias and structured-control-flow lowering already propagated the view and base pointer, but context propagation was optional on several paths. Losing that binding makes later remote or synchronization operations unable to resolve the correct communication domain.

A dynamic merge of tensors from different allocations cannot safely retain one branch context. Supporting that case requires selecting the base pointer and context together, then rebuilding the static tensor view. That larger design is tracked separately in #2027; this PR keeps the conservative rejection.

## User impact

Aliases and same-allocation control-flow carriers now remain valid inputs to distributed PTO operations. Invalid or unsupported metadata flows fail at the point where context is lost or conflicts, instead of silently selecting an incorrect communication window or failing later with a less actionable error.

## Testing

- `cmake --build build --parallel`
- `cmake --build build-pypto --parallel` using the existing Python 3.12 test environment
- `python -m pytest tests/ut/codegen/distributed/test_distributed_pto_codegen.py -n auto --maxprocesses 8 -q` — 28 passed
- `pre-commit run --files src/codegen/pto/pto_codegen.cpp src/codegen/pto/pto_control_flow_codegen.cpp tests/ut/codegen/distributed/test_distributed_pto_codegen.py`
- `python tests/lint/clang_tidy.py --diff-base HEAD`
- Full unit suite — 7,134 passed, 2 skipped, 1 unrelated environment failure in `test_no_dfx_without_output_prefix_is_ok`: the installed `_task_interface.CallConfig` lacks `enable_dump_args`

## Related issues

- #2027 tracks dynamic paired base-pointer / `CommContext` selection across control flow.
- #1533 and #1956 describe the static-view and PTOAS constraints behind the current conservative lowering.

### DIFF
--- a/src/codegen/pto/pto_codegen.cpp
+++ b/src/codegen/pto/pto_codegen.cpp
@@ -1521,7 +1521,13 @@ void PTOCodegen::VisitStmt_(const AssignStmtPtr& op) {
         BindTensorView(op->var_, view);
         BindVarToMlir(op->var_, view);  // view name == SSA name, as in ForStmt
         RegisterBasePtr(op->var_, GetTensorBasePtr(rhs_var));
-        RegisterCommCtxFor(op->var_, GetCommCtxSSAFor(rhs_var.get()));
+        const std::string comm_ctx = GetCommCtxSSAFor(rhs_var.get());
+        if (As<ir::DistributedTensorType>(op->var_->GetType())) {
+          INTERNAL_CHECK_SPAN(!comm_ctx.empty(), op->span_)
+              << "Internal error: DistributedTensor alias '" << op->var_->name_hint_ << "' from source '"
+              << rhs_var->name_hint_ << "' has no CommContext binding";
+        }
+        RegisterCommCtxFor(op->var_, comm_ctx);
         return;
       }
     } else if (!emit_tile_addr_ && As<TileType>(op->var_->GetType())) {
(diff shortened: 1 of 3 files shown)


Hzfengsy:
Closing in favor of #2029. The prototype established the intended conservative CommContext invariant, but the distributed-system run left one remote-store subview half unwritten. Although that test does not exercise the touched alias/control-flow paths and passed on main shortly before, distributed address-routing changes should be revalidated on hardware before reopening. Dynamic paired base/context selection remains tracked in #2027.

PR #2038, opened 2026-07-14
refactor(ir): Deduplicate window externalization helpers

lyfne123:
## Summary

- share call-name and Out-param return mapping helpers through the window externalization utility
- remove the stale duplicated helper block from OptimizeOrchTensors
- validate window operator identity through `IsOp` instead of raw name comparisons

## Testing

- `clang-format --dry-run --Werror` on all changed C++ files
- standalone C++17 syntax checks for both changed translation units
- `git diff --check`
- Full CMake build, unit tests, and clang-tidy were not run because the isolated worktree could not initialize the `msgpack-c` and runtime submodules; clang-tidy 18.1.8 is also older than the repository-required 21.1.0.

## Documentation

- Not required; this is an internal refactor with no user-facing API or intended behavior change.


Hzfengsy (approved):
Verified the refactor:

- **Dead-code removal is safe** — all 15 helpers dropped from `optimize_orch_tensors_pass.cpp` (`MakeUniqueFunctionName`, `VarRefCounter`, `CountVarRefsIn*`, `ExprReferencesOnlyVarsIn`, `CollectAllowedVars`, `BuildFunctionLookup`, `ScopedLoopIterInitSubst`, `IsAllZeroOffsets`, `IsTensorAllocationOp`, `IsOutputDirection`, `CollectLoopLocalTensorAllocs`, `CollectOutParamIndices`, ...) have zero remaining references in the file; they formed a self-contained cluster whose only callers were each other.
- **Sharing is wired correctly** — the header is already included, and every helper still living in `window_externalization.cpp`'s anonymous namespace retains a live caller, so no unused-function fallout.
- **`IsOp` conversions are behavior-preserving** — `IsOp` matches by canonical name and is null-safe, so a `GlobalVar` callee still yields `false` exactly as the raw `op_->name_` comparison did. All six literals (`tensor.create`, `tensor.full`, `tile.store`, `tensor.assemble`, `tile.load`, `tensor.slice`) are registered ops, so `GetOp`'s throw-on-typo guard cannot misfire at runtime.

This also brings `window_externalization.cpp` in line with the operator-identity-check convention. CI is fully green.

PR #2042, opened 2026-07-14
chore(ci): raise task-submit --timeout to 3600s

luohuan19:
## Summary
- Bump every `task-submit --timeout` from `1800s` to `3600s` across `.github/workflows/ci.yml` (12 invocations) and `.github/workflows/daily_ci.yml` (2 invocations).
- `--timeout` governs the device-acquisition **queue wait** — raising it gives device jobs more tolerance for runner contention before task-submit gives up waiting for a card.
- `--max-time` (the run wall-clock cap: 1800 / 900 / 600) is deliberately **left unchanged**, so this only grants more queue-wait headroom, not longer executions.

## Testing
- [ ] CI config change — validated by the CI runs on this PR.

## Notes
- The `--timeout 60` in the kill-orphaned-tasks path and `--task-queue-timeout=1800` in the batched system-tests step are different flags and are intentionally untouched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -314,7 +314,7 @@ jobs:
         # here collapses pytest's dir collection.)
         run: |
           source activate.sh
-          task-submit --device "$DEVICE_ID" --timeout 1800 --max-time 1800 \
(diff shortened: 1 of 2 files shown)


luohuan19:
Folding these timeout changes into #2039 instead of a separate PR.

Issue #1622, opened 2026-06-01
[Bug] tensor.gather on local TileType input can overwrite source row during lowering


zhaozhaozz:
### Component

IR / `ConvertTensorToTileOps` / `tensor.gather` lowering for `TileType` inputs

### Description

`tensor.gather` can mutate or alias its local TileType input when the input is a local tensor lowered to a tile.

The repro below creates a local `[16, 256]` FP32 tensor `local_scores`, reads it with a scalar loop, then calls:

```python
gathered_tile = pl.gather(local_scores, dim=-1, index=idx)
```

The gather output itself is correct. However, reading the same `local_scores` again with the same scalar loop after `pl.gather` returns wrong values: row 0 behaves as if it has been overwritten by row 15.

This violates the expected read-only semantics of `gather`: neither the source input nor the index input should be modified by the operation.

This was observed on PyPTO main after #1609, where `tensor.gather` now accepts `TileType` input/index and no longer fails at compile time.

### Minimal Reproducer

Save this as `repro_tensor_gather_local_alias.py` and run it with the normal PyPTO runtime environment configured.

```python
import argparse

import pypto.language as pl
import torch
from pypto.runtime.runner import RunConfig


ROWS = 16
COLS = 256
TOPK = 8


@pl.jit
def repro_kernel(
    scores: pl.Tensor[[ROWS, COLS], pl.FP32],
    idx: pl.Tensor[[ROWS, TOPK], pl.INT32],
    before: pl.Out[pl.Tensor[[ROWS, TOPK], pl.FP32]],
    gathered: pl.Out[pl.Tensor[[ROWS, TOPK], pl.FP32]],
    after: pl.Out[pl.Tensor[[ROWS, TOPK], pl.FP32]],
):
    for _ in pl.spmd(1, name_hint="gather_alias_repro"):
        local_scores = pl.create_tensor([ROWS, COLS], dtype=pl.FP32)
        local_scores[:, :] = scores[:, :]

        local_idx = pl.create_tensor([ROWS, TOPK], dtype=pl.INT32)
        local_idx[:, :] = idx[:, :]

        before_tile = pl.full([ROWS, TOPK], dtype=pl.FP32, value=0.0)
        for r in pl.range(ROWS):
            for k in pl.range(TOPK):
                col_i32 = pl.read(local_idx, [r, k])
                col = pl.cast(col_i32, pl.INDEX)
                val = pl.read(local_scores, [r, col])
[...]

PR #2044, opened 2026-07-14
WIP: changes to run DeepSeek V4 on A5

yanghaoran29:
## Summary

WIP stack for DeepSeek V4 (MXFP8/MXFP4) on Ascend950 (A5). Rebased onto current `main` (post-#2100).

### Included (unique to this PR)

1. **LegalizeTileCast** — expand unsupported `tile.cast` dtype pairs into native cast chains before AutoTileMatmulL0.
2. **A5 gather** — full-tile flat-index lowering for `tensor.gather`, plus materialize strided `tile.slice` sources before that path.
3. **AutoTile Vec→Mat staging** — when matmul is already L0-sized but LHS is still Vec-resident, insert `tile.move`→Mat (`BuildStageOnlyVecLhs`) so ExpandMixedKernel sees a proper V→C boundary (Hadamard / fused cast→matmul Left).

### Intentionally removed / not in this PR

- **#2100 already merged**: layout-aware same-addr `tile.move` elision + Vec-only ND↔NZ MemoryReuse gate. Do not re-land codegen elide or the wider `AreStorageLayoutsCompatible` gate (that caused Left overflow on `fits_l0c_split_k`).
- **VF-fusion / simpler fork pin**: outdated PR description; not part of the rebased commits (`runtime` follows main).

## Test plan

- [x] `pytest tests/ut/ir/transforms/test_legalize_tile_cast.py`
- [x] `pytest tests/ut/ir/transforms/test_auto_tile_matmul_l0.py`
- [x] `pytest tests/ut/ir/transforms/test_memory_reuse.py::TestStorageLayoutReuseGate` (still #2100 Vec-only)
- [x] gather-related UT filter (`-k gather ...`)
- [ ] CI full suite
- [ ] Optional board: `decode_sparse_attn.py` fused (no `--split-cast-matmul`) to confirm no #2100 regression

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,6 +193,7 @@ set(PYPTO_SOURCES
     src/ir/transforms/flatten_tile_nd_to_2d/rewrite_utils.cpp
     src/ir/transforms/flatten_tile_nd_to_2d/transpose.cpp
     src/ir/transforms/flatten_tile_nd_to_2d/verification.cpp
+    src/ir/transforms/legalize_tile_cast_pass.cpp
     src/ir/transforms/pass_context.cpp
     src/ir/transforms/passes.cpp
     src/ir/transforms/resolve_backend_op_layouts_pass.cpp
--- a/docs/en/dev/passes/14a-legalize_tile_cast.md
+++ b/docs/en/dev/passes/14a-legalize_tile_cast.md
@@ -1,0 +1,31 @@
+# LegalizeTileCast Pass
+
+Expands `tile.cast` `(src, dst)` pairs that the active `pto.tcvt` profile cannot emit as a single instruction into the shortest chain of native casts.
+
+## Overview
+
+For each `var = tile.cast(...)`:
+
+1. Look up the backend-specific adjacency table (from pto-isa `tcvt` Supported Conversions for A5 / A2A3).
+2. Already native: leave unchanged (including FIXPIPE-foldable `FP32→BF16/FP16` with `mode=rint`).
+3. Non-native: BFS for a shortest path; among equal-length paths prefer "same byte-width → float, then adjust width".
+
+Typical A5 results: `INT32→FP16` → `INT32→FP32→FP16`; `FP16→BF16` → `FP16→FP32→BF16`.
+
+Unreachable pairs hard-fail with src/dst/arch in the diagnostic.
+
+**Requires / Produces / Invalidates**: none (empty `PassProperties`).
+
+## When it runs
+
+Default pipeline:
+
+```text
+lower_composite_ops → flatten_tile_nd_to_2d → legalize_tile_cast → auto_tile_matmul_l0
+```
+
+## API
+
+| C++ | Python |
+| --- | ------ |
+| `pass::LegalizeTileCast()` | `passes.legalize_tile_cast()` |
--- a/docs/zh-cn/dev/passes/14a-legalize_tile_cast.md
+++ b/docs/zh-cn/dev/passes/14a-legalize_tile_cast.md
@@ -1,0 +1,38 @@
+# LegalizeTileCast Pass
+
+把目标 profile（A5 / A2A3）上 `pto.tcvt` **不支持** 的 `tile.cast` `(src, dst)` 对，展开为最短的原生 cast 链，避免静默发出非法 `tcvt`。
+
+## 概述
+
+`LegalizeTileCast` 是函数级 Pass。对每条 `var = tile.cast(...)`：
+
+1. 查当前 backend 的原生邻接表（来自 pto-isa `tcvt` Supported Conversions）。
+2. 已原生：原样保留（含可被 `AutoTileMatmulL0` FIXPIPE-fold 的 `FP32→BF16/FP16` + `rint`）。
+3. 非原生：在邻接图上 BFS 求最短路径；等长路径优先「同字节转浮点 → 再调宽度」。
+
+典型结果（A5）：
+
+| 用户 Cast | 分解 |
+|-----------|------|
+| INT32→FP16 | INT32→FP32 → FP32→FP16 |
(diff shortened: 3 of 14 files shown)


lwDavid:
Found and fixed a corruption bug in the A5 flat-index `tensor.gather` lowering added here (2718c5ab). Fix is on `lwDavid/pypto:fix/a5-gather-strided-flat-index` (commit [63d38fd6](https://github.com/lwDavid/pypto/commit/63d38fd6e98cc79b37a6b2d1d5d4b0cdf52080a3)) — feel free to cherry-pick it into this PR.

## Root cause

The A5 Case 1/2 full-tile path computes flat offsets

```
flat[i,j] = i * src_cols + index[i,j]
```

and emits one `pto.tgather`. This assumes the **source is stored row-major with stride == `src_cols`**. That holds for a `TensorType` source (materialized contiguous by `tile.load`) and for a freshly-created tile, but **not for a `tile.slice` view of a wider tile**: the view keeps the parent stride.

Concrete example (DeepSeek-V4 `sparse_attn` inverse-RoPE): `m_rope = n_full[:, NOPE_DIM:HEAD_DIM]` is logically `[H_TILE=16, ROPE_DIM=64]` but its rows sit at stride `HEAD_DIM`, not `ROPE_DIM`. So `flat[i,j] = i*64 + idx` misaddresses every row past the first. On A5 only **row 0** survived (ROPE rel-L2 ~0.42; pypto-lib worked around it by unrolling the rotation per head-row in [pypto-lib#801](https://github.com/hw-native-sys/pypto-lib/pull/801)).

The contiguous `[16,64]` case does **not** reproduce it — the discriminator is a strided source, not the row count.

## Fix

In the A5 flat-index path, when the gather source is a `TileType` (lowered to a view-only `tile.slice`), copy it into a fresh contiguous tile (`tile.create` + `tile.assemble`) before the flat-index `tile.gather`, so the flat offsets address a contiguous buffer. `TensorType` sources already materialize via `tile.load` and are not copied. A2A3 keeps the legacy per-row path and is untouched. (`src/ir/transforms/op_conversion_registry.cpp`, Case 1 + Case 2.)

## Validation (A5 / Ascend 950)

- **New ST `gather_rank2_swap_16row_strided`** — `[16,128]` tile, column-slice `[:,64:128]` (strided), `[16,64]` swap index: **960/1024 mismatched → PASS** after the fix.
- All **12 `TestGatherIndex`** ST cases pass on A5 — no regression; the #1622 tile-source-mutation regression is intact.
- **End-to-end**: `decode_sparse_attn_swa` with the **original `[16,64]` strided gather form** (no per-row workaround) now matches the workaround precision (~69% fail, same band) — the pypto-lib per-row workaround becomes unnecessary once this lands.
- **UT**: new `test_gather_conversion_strided_tile_source_materializes` locks in the materialization at the IR level; **136/136** conversion UTs pass.

## Drive-by

The commit also fixes a missing `from pypto.backend import BackendType` import in `tests/st/runtime/ops/test_gather.py` — without it the whole gather ST file fails to collect (`NameError`), which is why the >8-row / strided corruption shipped undetected. Added `[16,64]` contiguous + strided ST cases and the materialization UT.

Happy to split the test-import fix into a separate main-targeted PR if that's cleaner; the gather fix itself only applies on top of 2718c5ab.

lwDavid:
**dsv4 a5 status: 34/37 PASS** (pypto-lib commit [`00799b0`](https://github.com/lwDavid/pypto-lib/commit/00799b0))

With the gather fix (`63d38fd6`) + a pypto-lib PV-cube fusion workaround, **34 of 37** dsv4 scripts now PASS on A5 (was 20/37).

**Fixed:**
1. **a5 strided-source gather** (pypto `63d38fd6`, earlier comment).
2. **a5 PV-cube fusion bug** (pypto-lib `00799b0`): composed qk_pv mis-fuses `exp→cast(bf16)→cube`, losing ~15% precision. Workaround: GM-materialize `qk_exp_bf16` before the PV cube (4 files x3 lines). **+13 PASS** (whole attention family).
3. **hc_pre a5 guard** (`00799b0`): separate impl is core-agnostic. **+1 PASS**.

**Remaining 3 (hard):** `prefill_indexer` (6.14%, composed GM-to-cube drift, 7 fusion-break attempts failed); `prefill_layer`/`prefill_fwd` (507000/507901, multi-device EP2 deadlock/crash at 128 tokens; needs split-dispatch or simpler runtime fix).

---

**Warning: a5 fixes vs a3 performance**

These changes are in the **shared dsv4 operator source** (`models/deepseek/v4/`). These operators already pass on **a2a3 with tuned performance**. The a5 fixes add GM round-trip overhead that is correctness-safe for a2a3 but could regress a2a3 **performance**.

**Suggestion**: if dsv4 needs to run on **both** a2a3 (tuned) and a5 (workarounds), consider a **separate a5-specific directory** (e.g. `models/deepseek/v4_a5/`) to avoid a3 perf regression. Once the pypto/simpler root-cause fusion fix lands, the a5 workarounds can be removed and directories merged.

yanghaoran29:
## Update (2026-07-21)

Pushed `7aa5bca9` on `wip/legalize-tile-cast`:

**fix(ir): stage Vec LHS and gate ND/NZ reuse for A5 cast→matmul**

1. **AutoTileMatmulL0** — already-L0 matmuls with a Vec LHS get an explicit `tile.move`→Mat (same as K-tiled PV).
2. **MemoryReuse** — do not colocate tiles whose `blayout`/`slayout`/`fractal` differ (ND vs NZ).

### A5 board

`prefill_indexer` e2e (`score` / `topk_idxs` / `idx_kv_cache`) **PASS** with pto-isa Acc C2V `srcStride=Rows`, **without** `PREFILL_INDEXER_SPLIT_HAD_CAST_MATMUL=1` / without C2V inject.

Issue #2059, opened 2026-07-15
[Feature] Wire runtime dispatch-predicate into the pypto frontend (pl.spmd(predicate=...))


luohuan19:
### Summary

Expose the runtime's **dispatch-predicate** mechanism through the pypto frontend so a statically-materialized SPMD task can be conditionally retired (never dispatched to a core) based on a value only a prior task produces — evaluated by the scheduler at the dispatch point, not in orchestration.

The runtime side already shipped this in [hw-native-sys/simpler PR #1314](https://github.com/hw-native-sys/simpler/pull/1314) (currently pinned at `cddf2b6c`): `L0TaskPredicate` / `L0PredicateOperand` / `PredicateOp{NONE,EQ,NE,GT,LT,GE,LE}` and `Arg::set_predicate(...)`, evaluated in the scheduler via `DispatchPredicate::pass()`. **pypto codegen does not emit any of it yet** — that PR's description explicitly notes "Frontend (pypto `pl.spmd(predicate=...)`) lands separately." This issue tracks that frontend wiring.

### Motivation / Use Case

**DeepSeek-V4 expert-routed MoE — skip empty experts.** Each expert's per-expert `row_count` is runtime data only known after the router/gather stage, written into a tensor. pypto's AOT orchestration only has the tensor *address*, not the *value*, so "skip experts with `row_count == 0`" cannot be a constant known at submit:

- Reading the value in orchestration stalls on `wait_for_tensor_ready` (the producer must finish first) — serializing the pipeline.
- Submitting all experts and branching inside the kernel still pays the full launch + core-occupancy + data-movement cost per empty expert (in decode, most of 256 experts are empty).
- Making the expert count dynamic breaks the static task-graph model the whole runtime is built on.

A dispatch predicate resolves this: submit all expert tasks statically (no orchestration stall, graph stays static); each carries `predicate = row_count[e] > 0` and depends on the gather producer; the scheduler reads `row_count[e]` at the dispatch point (value already current, no wait) and dispatches only non-empty experts. A FALSE predicate retires the task inline via the dep-only path — it still settles fanin/fanout so the downstream combine unlocks (it does NOT vanish from the graph).

### Proposed API / Behavior

Constructor-first surface (comparison-expression sugar `row_count[e] > 0` is a follow-up):

```python
for e in pl.range(NUM_EXPERTS):          # static loop
    with pl.spmd(num_blocks,
                 deps=[gather_tid],                        # producer of row_count MUST be a dep
[...]

PR #2062, opened 2026-07-16
[WIP] Wire runtime dispatch-predicate into the pypto frontend (pl.spmd_submit(predicate=...))

luohuan19:
> **Status: WIP / Draft.** Core frontend is complete and validated end-to-end on onboard a2a3; see *Deferred* for what is intentionally out of scope for this PR.

Tracks [hw-native-sys/pypto#2059](https://github.com/hw-native-sys/pypto/issues/2059). Depends on the runtime `DispatchPredicate` struct + `Arg::set_predicate(...)` from [hw-native-sys/simpler#1314](https://github.com/hw-native-sys/simpler/pull/1314). That runtime is **already on `upstream/main`** (bumped to `a756969c`, a descendant of `cddf2b6c`, in [hw-native-sys/pypto#2061](https://github.com/hw-native-sys/pypto/pull/2061)), so **this PR no longer bumps the submodule** — it is pure frontend on top of the current runtime. That runtime PR explicitly deferred the frontend ("Frontend (pypto `pl.spmd(predicate=...)`) lands separately") — this is it.

## What this does

Exposes the runtime **dispatch predicate** through the DSL: a statically-materialized SPMD task can be **conditionally retired** (never dispatched to a core) based on a value only a *prior* task produces. The condition is evaluated by the scheduler **at the dispatch point** — after the task's dependencies are satisfied, so the value is current — **not** in orchestration (which would stall on `wait_for_tensor_ready`). A false predicate retires the task inline via the dep-only path, still settling fanin/fanout so downstream consumers unlock; it does **not** vanish from the graph.

### Motivating use case — MoE skip-empty-experts

DeepSeek-V4 expert-routed MoE computes each expert's `row_count` at runtime (written into a tensor). Orchestration only has the tensor *address*, not the *value*, so "skip experts with `row_count == 0`" cannot be a constant known at submit. Alternatives and why they fail:

| Approach | Problem |
| --- | --- |
| Read the count in orchestration, branch on submit | Stalls on `wait_for_tensor_ready` → serializes the pipeline |
| Submit all experts, branch inside the kernel | Still pays launch + core-occupancy + data-movement per empty expert |
| Make the expert count dynamic | Breaks the static task-graph / AOT-orchestration model |

The dispatch predicate resolves this: submit all experts statically (graph stays static, no orchestration stall); each carries `predicate = row_count[e] > 0` and depends on the gather producer; the scheduler reads `row_count[e]` at the dispatch point (no wait) and dispatches only non-empty experts.

## Frontend usage (the writing form)

Constructor-first surface — a new `pl.dispatch_pred(...)` passed as `predicate=` to `pl.spmd_submit` / `pl.submit`:

```python
out, tid = pl.spmd_submit(
    self.kernel, *args, core_num=N,
    deps=[producer_tid],                               # see Contract below
    predicate=pl.dispatch_pred(tensor, [i0, i1, ...], "op", target),
)
```

`pl.dispatch_pred(tensor, indices, op, target)` maps 1:1 onto the runtime `operand[indices] <op> target`:

| Arg | Meaning | Constraint |
| --- | --- | --- |
| `tensor` | operand tensor read at the dispatch point | must be a tensor |
| `[indices]` | element locator | **list literal**; each element a `ConstInt` or loop `Var`; non-empty |
| `"op"` | comparison spelling | one of `==` `!=` `>` `<` `>=` `<=` (string literal) |
| `target` | right-hand side | **integer literal** (may be negative) |

MoE example (the shape validated end-to-end):

```python
@pl.function(type=pl.FunctionType.Orchestration)
def main(self, x, out, rc):                            # rc = row_count tensor
    with pl.manual_scope():
        rc, g_tid = pl.spmd_submit(self.gate, rc, core_num=1)      # producer of rc
        out, _ = pl.spmd_submit(
            self.expert, x, out, core_num=1,
            deps=[g_tid],                                          # producer is a dep
            predicate=pl.dispatch_pred(rc, [0, 0], ">", 0),        # rc[0,0] > 0
        )
    return out
```

Generated orchestration C++ (mirrors the runtime's hand-written form):

```cpp
L0TaskPredicate params_t1_pred;
params_t1_pred.operand.tensor = &ext_rc;
params_t1_pred.operand.ndims = 2;
params_t1_pred.operand.indices[0] = 0;
params_t1_pred.operand.indices[1] = 0;
params_t1_pred.op = PredicateOp::GT;
params_t1_pred.target = 0;
params_t1.set_predicate(params_t1_pred);
```

Parse-time validation rejects malformed predicates (non-tensor operand, non-list indices, bad op spelling, non-literal target, a non-`dispatch_pred` value).

### Contract

**The predicate operand tensor's producer MUST be one of the submit's `deps=`** so the dispatch-point read observes the current value. In this WIP only the *structural* well-formedness is enforced (operand present & a tensor, indices non-empty, op in range) in the `Submit` constructor; the *dataflow* "producer ∈ deps" check is deferred (see below), so the author is currently responsible for `deps=`.

## Design approach

Mirrors the existing `allow_early_resolve` path (simpler#1065) end-to-end: `pl.spmd_submit → Submit → SubmitToCallView → orchestration codegen`.

- **IR (`Submit`)** — four first-class fields: `predicate_operand` (`optional<ExprPtr>`), `predicate_indices` (`vector<ExprPtr>`), `predicate_op` (`int64`, a `DispatchPredicateOp`), `predicate_target` (`int64`). The operand/indices are SSA-bearing IR-node fields; op/target are stored as `int64` **leaves on purpose** — the reflection system requires an explicit `VisitLeafField` overload per leaf type (structural-equal / hash / serializer / deserializer), so a dedicated enum leaf would cost 4+ touchpoints; `int64` + `optional<ExprPtr>` + `vector<ExprPtr>` are all already supported, keeping reflection zero-cost. A `DispatchPredicateOp` enum is used at the API boundary; the int is an internal encoding whose values match the runtime `PredicateOp`.
- **Pass safety** — the operand + indices are part of the use-def chain, so the C++ visitor/mutator walk them (like `deps_` / `core_num_`), and **all 13 `Submit` reconstruction sites** across the passes preserve the predicate via a new `Submit::GetPredicateInit()` helper. Verified: the predicate survives inline / SSA / outline / materialize and resolves to `&ext_<name>` at codegen.
- **Cross-layer** — nanobind `Submit` ctor gains the predicate args; `DispatchPredicateOp` enum is bound (the `None`/int-0 member is not exposed to Python — `None` is a keyword; "no predicate" = pass none); `ir.pyi` updated.
- **Codegen** — `EmitPredicateHint` emits the `L0TaskPredicate` + `set_predicate` block on the task's `L0TaskArgs` before `rt_submit_*`. `elem_size` is intentionally left for the runtime to derive from the tensor dtype.
- **Printer / deserializer** — round-trip the four fields (`predicate=pl.dispatch_pred(...)` prints and reparses).

### Why `pl.spmd_submit` (not the scope form) for this cut

`pl.spmd_submit` builds a `Submit` directly at parse time (before SSA), so the Submit-level pass-walk fully covers it with **zero scope-attr plumbing**. The `with pl.spmd(predicate=...)` scope form is outlined *after* SSA (pass 8-9), so its operand/index Vars would live on scope attrs during SSA and need extra `VisitScopeAttrs`/`MutateScopeAttrs` Expr-walking + scope serializer/printer handling — deferred to keep this PR tight.

## Scope / deferred

Out of scope for this WIP (each a clean follow-up; all share the same IR fields + codegen):

1. **Dataflow verifier** — "predicate operand producer must be a dep" (only the structural check is in).
2. **`with pl.spmd(predicate=...)` scope form** (needs scope-attr Expr-walking).
3. **Comparison-expression sugar** — `predicate=(row_count[e] > 0)` instead of the constructor.

Predicate expressiveness is deliberately limited to `tensor[indices] OP const` — this matches the runtime's single-comparison `DispatchPredicate`. Multiple conditions / boolean combination / arithmetic are not expressible; the pattern is to reduce them to one gate value in a prior kernel.

## Testing

- `tests/ut/language/parser/test_submit_predicate.py` — parse/IR field population, all 6 op spellings, negative-target, 5 negative cases, print→parse round-trip.
- `tests/ut/codegen/test_predicate_codegen.py` — full Default pipeline → asserts the `set_predicate` block (and its absence without a predicate).
- `tests/st/runtime/scheduling/test_predicated_dispatch.py` — end-to-end: `x_producer → clobber(predicate=gate[0,0]>0, deps=[xp]) → consumer(deps=[clobber_tid])`. gate=0 → clobber skipped → sentinel(42) survives; gate=1 → clobber dispatched → poison(999); consumer unlocks in **both** cases (proving inline-retire still settles fanout). **Passes on onboard a2a3.**
- Regression: 2808 transforms + codegen UTs green.


luohuan19:
Superseded by the phased tracking issue [hw-native-sys/pypto#2066](https://github.com/hw-native-sys/pypto/issues/2066) and a fresh Phase-1 PR (linked below). Same branch/commits — reopening as a non-WIP, tracker-scoped PR. No work is lost.

PR #2063, opened 2026-07-16
ci: split system-tests into two balanced jobs

lyfne123:
## Problem

The `system-tests` job's batched `Test system tests` step dominated its wall-clock: measured at ~7–15 min across recent `main` runs (60–75% of all test time) and unbounded on heavy runs. Because that one step was the long pole, the job could not be shortened by rearranging the lighter direct-device / DFX steps around it.

Per-step timings from three recent `main` runs (seconds):

| Step | run A | run B | run C |
| --- | --- | --- | --- |
| Setup NPU (build) | 133 | 206 | 142 |
| **Test system tests** (batched) | **426** | **455** | **932** |
| direct-device + paged + multi-orch + swimlane + dump-args | 276 | 424 | 283 |

## Change

Split the batched `device_batch` load **by directory** across two jobs that run in parallel:

- **`system-tests`** — batched shard 1: the heavy dirs `runtime/ops` + `runtime/cross_core` (~68% of the batched load).
- **`system-tests-direct`** — batched shard 2: `device_batch` on the remaining dirs, plus **all** `-m 'not device_batch'` direct-device tests and the DFX steps (paged-attention, multi-orch L2, swimlane, dump-args).

The ~68/32 batched split balances shard 1 against the extra direct + DFX load carried by the second job, so the critical path is roughly the heavier of the two rather than their sum. Heavy-run estimate: **~18 min → ~12.5 min**, with the two jobs finishing at about the same time.

## Coverage (unchanged)

- shard-1 dirs ∪ shard-2 (rest) exactly reconstruct the original batched scope; shard 2 `--ignore`s `ops`/`cross_core` so nothing runs twice.
- The direct-device step still runs `-m 'not device_batch'` over all of `tests/st`, so direct tests living under `ops`/`cross_core` are not dropped.
- `distributed` / `codegen` / `dump_tag` exclusions are preserved.
- The second job checks out under `./st-direct-checkout` to avoid colliding with `system-tests`' `./st-checkout` on the shared self-hosted host, and cleans up its own orphaned task-submit tasks.

## Notes

- The 68/32 partition is estimated from the count of `test_runner.run` files per directory; the directory allocation may need a small tweak once real per-job timings are observed (documented in the job comments).
- Both jobs run their own build (ccache-shared, usually a cache hit) — the fixed cost of parallelizing.
- The existing required check name `system-tests` is unchanged. To gate merges on the new job too, add `system-tests-direct` to the branch protection required status checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,6 +195,19 @@ jobs:
           --ignore=tests/st/codegen/torch/test_torch_codegen_paged_attention.py
 
   system-tests:
+    # Batched shard 1 (mechanism A): the HEAVY device_batch directories
+    # (runtime/ops + runtime/cross_core, ~68% of the batched load). The batched
+    # `Test system tests` step used to run ~7-15 min and dominated the whole job,
+    # so it is split by directory across two jobs that run in parallel:
+    #   - system-tests        (this job): device_batch on ops + cross_core
+    #   - system-tests-direct (sibling)  : device_batch on the REST + all the
+    #                                      direct-device (mechanism B) + DFX steps
+    # The two shards are balanced so the critical path ≈ the heavier of the two,
+    # not their sum. Coverage is preserved: shard-1 dirs ∪ shard-2 (rest) = the
+    # original batched scope, with no overlap (shard 2 --ignore's ops/cross_core).
+    # Both jobs do their own (ccache-shared) build; keep them in sync on setup and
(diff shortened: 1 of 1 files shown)


YunjiQin (approved):
CI-only change: splits the batched system-tests load across two parallel jobs (system-tests heavy dirs, system-tests-direct light ops/cross_core + direct + DFX). Coverage preserved via mutual --ignore; all checks green. LGTM.

Issue #1968, opened 2026-07-07
[RFC] Window-Buffer Allocation API Is Byte-Centric and Error-Prone


georgebisbas:
> **This is a request for comment (RFC), not a committed work plan.** The goal is
> to discuss whether the window-buffer allocation API should change and, if so,
> which direction to take. Feedback on the options below — especially the
> trade-offs and the open questions — is very welcome.

### Summary

`pld.alloc_window_buffer(size)` takes a **raw byte count**, so every caller manually
computes `product(shape) × bytes_per_element` and then restates the *same* shape and
dtype one line later in `pld.window(buf, shape, dtype=...)`. This leaks a hardware
detail (byte size) into the frontend, duplicates information across two calls, and is
a silent foot-gun when the multiplier is wrong. This RFC lays out the problem and
several possible API directions, from a tiny accessor to a shape+dtype allocation
API, and asks which (if any) we should pursue.

### Motivation

The current pattern:

```python
data_buf   = pld.alloc_window_buffer(nr * SIZE * 4)   # magic 4 = FP32 bytes
signal_buf = pld.alloc_window_buffer(nr * 4)          # magic 4 = INT32 bytes
...
data = pld.window(data_buf,   [nr, SIZE], dtype=pl.FP32)   # shape+dtype AGAIN
sig  = pld.window(signal_buf, [nr, 1],    dtype=pl.INT32)  # shape+dtype AGAIN
```

Why the current state is insufficient:

1. **Byte arithmetic leaks to the frontend.** The user must know FP32 = 4 bytes,
   BF16 = 2, and remember `nelements × bytes`. That is a runtime/HW concern, not a
   modeling concern.
2. **Duplication / two sources of truth.** Shape and dtype are stated at the alloc
   (implicitly, via the byte math) *and* again at `window()`. The two can silently
   drift out of sync.
3. **Silent corruption on a wrong multiplier.** A wrong constant (or forgetting a
   dimension in the product) under/over-allocates the buffer — no error, just memory
   corruption at runtime.
4. **Not self-documenting.** `alloc_window_buffer(nr * SIZE * 4)` needs a comment to
   explain what it is; the intent (`an nr×SIZE FP32 buffer`) is lost.
5. **Sub-byte bug surface.** Ad-hoc `bits / 8` truncates 4-bit types (INT4/UINT4/
   FP4/HF4) to 0 bytes; only `(bits + 7) / 8` is correct. Scattering this math across
   call sites invites the wrong variant.

[...]

PR #2065, opened 2026-07-16
feat(distributed): add shape+dtype overload to alloc_window_buffer

georgebisbas:
## Summary

Adds `pld.alloc_window_buffer(shape, *, dtype=...)` — a convenience overload
that computes the per-rank byte size from the shape and dtype, eliminating
manual `product(shape) * dtype.get_byte()` arithmetic. Normalizes to the
existing IR op — no C++ changes, no new IR surface.

Implements option 5 of RFC #1968.

## Before / After

```
# Canonical byte form (preserved)
data_buf   = pld.alloc_window_buffer(nr * SIZE * pl.FP32.get_byte())
signal_buf = pld.alloc_window_buffer(nr * pl.INT32.get_byte())

# Shape+dtype overload (new)
data_buf   = pld.alloc_window_buffer([nr, SIZE], dtype=pl.FP32)
signal_buf = pld.alloc_window_buffer([nr], dtype=pl.INT32)
```

The canonical form is unchanged. The two-phase `alloc -> window` design
remains first-class — a single buffer can still back multiple `window()`
views across loop iterations.

## Changes (2 files, DSL-only)

| File | Change |
|------|--------|
| `tensor_ops.py` | `_compute_window_buffer_bytes` helper, two `@overload` signatures, dispatch on first-arg type |
| `test_alloc_window_buffer.py` | 2 new tests, 2 error-match strings updated |

## Verification

- [x] `pytest tests/ut/ir/parser/test_alloc_window_buffer.py` — 12/12 pass (sim Docker)
- [ ] Distributed ST regression

### DIFF
--- a/python/pypto/language/distributed/op/tensor_ops.py
+++ b/python/pypto/language/distributed/op/tensor_ops.py
@@ -88,19 +88,129 @@ def _validate_pipeline(pipeline: bool, chunk_rows: int, chunk_cols: int, op_name
         )
 
 
-def alloc_window_buffer(size: IntLike, *, name: str = "") -> Ptr:
-    """Declare a per-rank HCCL window-buffer in a comm-domain scope slot of ``size`` bytes.
-
-    Mirrors ``tile.alloc(memory_space, size)``: pure allocation semantics, no
-    shape / dtype concept on the buffer itself. The result is the
-    allocation-identity token that ``pld.tensor.window`` consumes.
+def _validate_window_buffer_shape(shape: list[int | Expr]) -> None:
+    """Validate that a window-buffer shape is non-empty and static dims are positive.
 
     Args:
-        size: Per-rank allocation size in **bytes**. Accepts an ``int``
-            literal, a DSL ``Scalar``, or a raw :class:`ir.Expr`.
+        shape: Normalized shape — elements are ``int`` or :class:`ir.Expr`.
+
+    Raises:
+        ValueError: If ``shape`` is empty or any statically-known dimension
+            is non-positive.
+    """
+    if not shape:
+        raise ValueError("pld.tensor.alloc_window_buffer shape must be non-empty")
+    for i, d in enumerate(shape):
+        if isinstance(d, int):
+            if d <= 0:
+                raise ValueError(
+                    f"pld.tensor.alloc_window_buffer shape dimension[{i}] is {d}; "
+                    "all dimensions must be positive"
+                )
+        elif isinstance(d, _ir.ConstInt):
+            if d.value <= 0:
+                raise ValueError(
+                    f"pld.tensor.alloc_window_buffer shape dimension[{i}] is {d.value}; "
+                    "all dimensions must be positive"
+                )
+
+
+def _compute_window_buffer_bytes(shape: list[int | Expr], dtype: DataType) -> int | Expr:
+    """Compute byte size for a shape+dtype ``alloc_window_buffer`` call.
+
+    When all shape dimensions are statically known (Python ``int`` or
+    :class:`ir.ConstInt`) the product is computed eagerly as a Python
+    ``int`` (no IR nodes emitted).  Otherwise (any dimension is a dynamic
+    :class:`ir.Expr`) a chain of :class:`ir.Mul` nodes is emitted.
+
+    Args:
+        shape: Normalized shape — elements are ``int`` or :class:`ir.Expr`.
+        dtype: Element data type (byte size via ``dtype.get_byte()``).
+
+    Returns:
+        Byte size as ``int`` (static) or ``ir.Expr`` (dynamic).
+    """
+    _validate_window_buffer_shape(shape)
+
+    # Fold pure-static: Python ints AND ir.ConstInt — both are compile-time
+    # constants.  Per the parser, integer literals arrive as ConstInt(INDEX),
+    # not raw Python int, so we must check both.
+    static_total = 1
+    for d in shape:
+        if isinstance(d, int):
+            static_total *= d
+        elif isinstance(d, _ir.ConstInt):
+            static_total *= d.value
+        else:
+            break
+    else:
+        # All dims folded — return an eager Python int.
+        return static_total * dtype.get_byte()
+
+    # Dynamic path: at least one genuinely dynamic dim.
+    span = _ir.Span.unknown()
+    # Seed from the first dim instead of ConstInt(1) to avoid a redundant
+    # Mul(1, dim0) identity in the emitted IR.
+    first = shape[0]
+    if isinstance(first, int):
+        result: Expr = _ir.ConstInt(first, DataType.INT64, span)
+    elif isinstance(first, _ir.ConstInt):
+        result = _ir.ConstInt(first.value, DataType.INT64, span)
+    else:
+        result = first
+    for dim in shape[1:]:
+        if isinstance(dim, int):
+            dim_expr = _ir.ConstInt(dim, DataType.INT64, span)
+        elif isinstance(dim, _ir.ConstInt):
+            dim_expr = _ir.ConstInt(dim.value, DataType.INT64, span)
+        else:
+            dim_expr = dim
+        result = _ir.Mul(result, dim_expr, DataType.INT64, span)
+    byte_expr: Expr = _ir.ConstInt(dtype.get_byte(), DataType.INT64, span)
+    return _ir.Mul(result, byte_expr, DataType.INT64, span)
+
+
+@overload
+def alloc_window_buffer(size: IntLike, *, name: str = "") -> Ptr: ...
+
+
+@overload
+def alloc_window_buffer(shape: Sequence[IntLike], *, dtype: DataType, name: str = "") -> Ptr: ...
+
+
+def alloc_window_buffer(  # type: ignore[no-redef]
+    size: IntLike | Sequence[IntLike], *, dtype: DataType | None = None, name: str = ""
+) -> Ptr:
+    """Declare a per-rank HCCL window-buffer in a comm-domain scope slot.
+
+    Two forms:
+
+    * **Canonical byte form:**
+      ``alloc_window_buffer(size, *, name=...)`` — ``size`` is the per-rank
+      allocation size in **bytes**.  Accepts an ``int`` literal, a DSL
+      ``Scalar``, or a raw :class:`ir.Expr`.
+
+    * **Shape+dtype convenience overload:**
+      ``alloc_window_buffer(shape, *, dtype=..., name=...)`` — ``shape`` is a
+      list / tuple of per-rank dimensions.  The byte size is computed
+      automatically as ``product(shape) × dtype.get_byte()`` and the call
+      normalizes to the canonical byte form.  ``dtype`` is required when
+      ``shape`` is a sequence and rejected otherwise.
+
+    Both forms return the singleton :class:`ir.PtrType` allocation-identity
+    token that ``pld.tensor.window`` consumes.  The two-phase
+    ``alloc → window`` design is preserved — a single buffer can back multiple
+    ``window()`` views across loop iterations.
+
+    Args:
+        size: (Canonical form) Per-rank allocation size in **bytes**.
+        shape: (Convenience form) Per-rank shape as a sequence of per-dimension
+            sizes — ``int``, DSL ``Scalar``, or raw :class:`ir.Expr`.
+        dtype: (Convenience form, required) Element data type. Must be
+            ``None`` (default) in the canonical byte form.
         name: Unique buffer identifier. The parser injects this from the LHS
             of the surrounding assignment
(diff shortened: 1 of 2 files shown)


python/pypto/language/distributed/op/tensor_ops.py
```diff
@@ -88,19 +88,78 @@ def _validate_pipeline(pipeline: bool, chunk_rows: int, chunk_cols: int, op_name
...
+        nelements = math.prod(shape)  # type: ignore[arg-type]
+        return nelements * dtype.get_byte()
+
+    span = _ir.Span.unknown()
+    result: Expr = _ir.ConstInt(1, DataType.INT64, span)
+    for dim in shape:
+        dim_expr = dim if isinstance(dim, Expr) else _ir.ConstInt(dim, DataType.INT64, span)
+        result = _ir.Mul(result, dim_expr, DataType.INT64, span)
+    byte_expr: Expr = _ir.ConstInt(dtype.get_byte(), DataType.INT64, span)
+    return _ir.Mul(result, byte_expr, DataType.INT64, span)
```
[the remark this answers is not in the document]
georgebisbas: Intentional — the `ConstInt(1)` initializer simplifies the loop by avoiding a first-iteration special case (no `result is None` branching). The extra `Mul(ConstInt(1), d0)` is a single IR node at negligible cost. The 1-byte short-circuit for BOOL/INT8 is similarly a micro-optimization that adds branching for a rare case. Prefer simplicity here.

python/pypto/language/distributed/op/tensor_ops.py
```diff
@@ -112,16 +171,32 @@ def alloc_window_buffer(size: IntLike, *, name: str = "") -> Ptr:
...
+        ValueError: If ``shape`` is a sequence but ``dtype`` is not provided.
+        ValueError: If ``size`` is scalar but ``dtype`` is provided.
     """
     if not name:
         raise ValueError(
             "pld.tensor.alloc_window_buffer must appear as the RHS of a simple assignment "
             "(its result must be bound to a named variable)"
         )
+
     if isinstance(size, (list, tuple)):
```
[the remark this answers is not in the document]
georgebisbas: Not needed — `_normalize_intlike` and `_compute_window_buffer_bytes` both accept `Sequence` and iterate identically over list and tuple. No functional difference between the two.

python/pypto/language/distributed/op/tensor_ops.py
```diff
@@ -88,19 +88,78 @@ def _validate_pipeline(pipeline: bool, chunk_rows: int, chunk_cols: int, op_name
...
+    return _ir.Mul(result, byte_expr, DataType.INT64, span)
+
+
+@overload
+def alloc_window_buffer(size: IntLike, *, name: str = "") -> Ptr: ...
+
+
+@overload
+def alloc_window_buffer(shape: Sequence[IntLike], *, dtype: DataType, name: str = "") -> Ptr: ...
+
```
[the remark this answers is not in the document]
georgebisbas: `Sequence[IntLike]` is the established convention for shape parameters across the DSL (e.g. `window`, `put`, `get`). The `isinstance(size, (list, tuple))` guard is the standard dispatch pattern used throughout the codebase to discriminate scalar vs sequence forms. Tightening the overload to `list | tuple` would be inconsistent with the rest of the DSL.

python/pypto/language/distributed/op/tensor_ops.py
```diff
@@ -88,19 +88,78 @@ def _validate_pipeline(pipeline: bool, chunk_rows: int, chunk_cols: int, op_name
...
+
+def alloc_window_buffer(  # type: ignore[no-redef]
+    size: IntLike | Sequence[IntLike], *, dtype: DataType | None = None, name: str = ""
+) -> Ptr:
```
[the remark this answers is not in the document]
georgebisbas: Same as above — `Sequence[IntLike]` is the standard DSL convention for shape parameters. Tightening to `list | tuple` would be inconsistent with sibling ops.

tests/ut/ir/parser/test_alloc_window_buffer.py
```diff
@@ -220,5 +224,44 @@ def host_orch(self):
...
+            return buf
+
+    func = _get_host_orch(P)
+    stmt = _find_alloc_assignment(func)
+    assert isinstance(stmt.value, ir.Call)
+    call = stmt.value
+    assert call.op.name == "pld.tensor.alloc_window_buffer"
+    assert call.kwargs["name"] == "buf"
+    assert "dtype" not in call.kwargs
+    assert len(call.args) == 1
```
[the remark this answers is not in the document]
georgebisbas: A byte-size assertion was tried and removed during development. The parser inflates Python ints to `Scalar`/`Expr` objects before `invoke_dsl`, so `_compute_window_buffer_bytes` sees `[Expr, Expr]` and takes the dynamic `Mul`-chain path — the result is never a single `ConstInt`. The test correctly verifies structural IR properties (single arg, name kwarg, no dtype in kwargs) which is sufficient for correctness.

tests/ut/ir/parser/test_alloc_window_buffer.py
```diff
@@ -220,5 +224,44 @@ def host_orch(self):
...
+            buf = pld.tensor.alloc_window_buffer([32, 16], dtype=pl.INT32)
+            return buf
+
+    func = _get_host_orch(P)
+    stmt = _find_alloc_assignment(func)
+    assert isinstance(stmt.value, ir.Call)
+    call = stmt.value
+    assert call.op.name == "pld.tensor.alloc_window_buffer"
+    assert call.kwargs["name"] == "buf"
+    assert len(call.args) == 1
```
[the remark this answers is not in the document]
georgebisbas: Same rationale as the previous thread — the static shape dimensions are inflated to IR expressions by the parser, so the byte size is an IR `Mul` chain and can never be a raw `ConstInt` at this level. The structural checks (op name, kwargs, arg count) are sufficient.

YunjiQin (commented):
Code review of the shape+dtype overload. Four inline points below; the first and third are the important ones. Context: on this repo dynamic window-buffer sizes are a first-class, ST-exercised feature (e.g. `pld.alloc_window_buffer(pld.world_size() * pl.INT32.get_byte())`), and the size expr is lowered symbolically all the way to `orch.allocate_domain(window_size=...)` in `distributed_codegen.cpp`. So the dynamic (Mul-chain) path is exactly the one that matters and needs coverage.

python/pypto/language/distributed/op/tensor_ops.py
```diff
@@ -88,19 +89,76 @@ def _validate_pipeline(pipeline: bool, chunk_rows: int, chunk_cols: int, op_name
...
     Args:
-        size: Per-rank allocation size in **bytes**. Accepts an ``int``
-            literal, a DSL ``Scalar``, or a raw :class:`ir.Expr`.
+        shape: Normalized shape — elements are ``int`` or :class:`ir.Expr`.
+        dtype: Element data type (byte size via ``dtype.get_byte()``).
+
+    Returns:
+        Byte size as ``int`` (static) or ``ir.Expr`` (dynamic).
+    """
+    all_ints = all(isinstance(d, int) for d in shape)
```
YunjiQin: The `all_ints` eager-folding branch is effectively dead for parser-driven DSL calls. When a shape reaches this helper through the parser, integer literals arrive as `ir.ConstInt`, not Python `int`: `parse_constant` builds `ConstInt(value, INDEX)`, and `_dsl_invoker._wrap_arg` unwraps the `MakeTuple` to a list of the element *Exprs*. So `isinstance(d, int)` is `False` for every dim and even a fully static shape like `[64, 128]` falls through to the `Mul`-chain path — contradicting the docstring's "computed eagerly ... (no IR nodes emitted)". It also regresses IR cleanliness vs. the canonical byte form, whose Python-level product folds to a single `ConstInt`.

Suggestion: fold `ConstInt` too — treat a dim as static when it is a Python `int` *or* an `ir.ConstInt` (extract `.value`), so parser-supplied static shapes still collapse to one constant.

python/pypto/language/distributed/op/tensor_ops.py
```diff
@@ -88,19 +89,76 @@ def _validate_pipeline(pipeline: bool, chunk_rows: int, chunk_cols: int, op_name
...
+    if all_ints:
+        nelements = math.prod(shape)  # type: ignore[arg-type]
+        return nelements * dtype.get_byte()
+
+    span = _ir.Span.unknown()
+    result: Expr = _ir.ConstInt(1, DataType.INT64, span)
```
YunjiQin: Two minor issues in the dynamic path:
1. Seeding with `ConstInt(1)` leaves a redundant `Mul(1, dim0)` identity node in the emitted IR. Prefer seeding `result` from the first dim and iterating the rest.
2. `dim_expr` may be a parser-supplied `ConstInt` typed `INDEX`, but every `Mul` is forced to `INT64`, mixing `INDEX x INT64` operands and bypassing the DSL's normal type promotion. Worth confirming the verifier and `GetExprAsCode` (distributed_codegen.cpp) handle this cleanly for a genuinely dynamic dim.

python/pypto/language/distributed/op/tensor_ops.py
```diff
@@ -88,19 +89,76 @@ def _validate_pipeline(pipeline: bool, chunk_rows: int, chunk_cols: int, op_name
...
-            literal, a DSL ``Scalar``, or a raw :class:`ir.Expr`.
+        shape: Normalized shape — elements are ``int`` or :class:`ir.Expr`.
+        dtype: Element data type (byte size via ``dtype.get_byte()``).
+
+    Returns:
+        Byte size as ``int`` (static) or ``ir.Expr`` (dynamic).
+    """
+    all_ints = all(isinstance(d, int) for d in shape)
+    if all_ints:
+        nelements = math.prod(shape)  # type: ignore[arg-type]
```
YunjiQin: No validation on the shape here. `alloc_window_buffer([], dtype=...)` gives `math.prod([]) == 1` and silently allocates one element's worth of bytes; a zero or negative static dim yields `0`/negative bytes. Consider rejecting an empty shape and non-positive static dims with a clear `ValueError`.

tests/ut/ir/parser/test_alloc_window_buffer.py
YunjiQin, line 246: Both new tests only cover static int-literal shapes and assert `len(call.args) == 1`, which holds whether the byte size folded to a constant or expanded into a `Mul` chain — so they cannot catch the dead-eager-path issue noted in the wrapper. Please add (UT only, no ST needed):
1. An assertion that a static shape folds to a single `ConstInt` arg (guards the fold).
2. A dynamic-dim test (e.g. a shape dim from `pld.world_size()` or a scalar) exercising the `Mul`-chain path, checking it parses and type-deduces. This is the primary real-world usage of window-buffer sizing and is currently untested for the new form.

python/pypto/language/distributed/op/tensor_ops.py
georgebisbas: Fixed in `db680113` — `_compute_window_buffer_bytes` now checks both `int` and `ir.ConstInt` for the static fold, and the eager path extracts `.value` from ConstInt dims. Parser-supplied `[64, 128]` now collapses to a single `ConstInt(byte_size, INT64)` instead of a Mul chain.

python/pypto/language/distributed/op/tensor_ops.py
georgebisbas: Fixed in `db680113` — (1) The dynamic path now seeds from the first dim instead of `ConstInt(1)`, avoiding the redundant identity Mul. (2) Static dims in the dynamic path (int or ConstInt) are now re-wrapped as `ConstInt(value, INT64)` so all Mul operands are consistently INT64. The existing codegen/distributed_codegen.cpp handles both Mul and INT64 types without issue.

python/pypto/language/distributed/op/tensor_ops.py
georgebisbas: Fixed in `db680113` — added `_validate_window_buffer_shape` that rejects empty shapes and non-positive static dims (both int and ConstInt) with clear error messages. Two new UT tests cover both edge cases.

tests/ut/ir/parser/test_alloc_window_buffer.py
georgebisbas: Fixed in `db680113` — (1) Both static-shape tests now assert the arg is a `ConstInt` with the exact byte-size value and INT64 dtype, guarding the fold. (2) Added `test_alloc_window_buffer_shaped_dynamic` exercising `[64, pld.world_size()]` which verifies the dynamic path produces a Mul chain with INT64 ScalarType.

YunjiQin (approved):

PR #2105, opened 2026-07-22
feat(ir): Add bitcast op for zero-copy dtype reinterpretation

lyfne123:
## Summary

Adds `tensor.bitcast` / `tile.bitcast` — a zero-copy element-type reinterpretation over the same bytes, mapping to PTOAS `pto.bitcast` semantics. Shape, `valid_shape`, layout, `fractal` and `pad` all carry through unchanged; only `dtype` differs.

Until now the only way to reinterpret bits was `tile.gather_mask(..., output_dtype=...)`, which forces an otherwise unnecessary mask gather. `tile.cast` is a *value* conversion and `reshape` only changes shape, so neither covers e.g. extracting `sort32` index bits from an FP32 tile or doing integer bit manipulation on float data.

```python
# extract sort32 index bits
idx = pl.tile.bitcast(sorted_tile, pl.UINT32)

# sign-bit masking through integer ops (pl.bitcast dispatches Tensor vs Tile)
bits   = pl.bitcast(x, pl.INT32)
masked = pl.tile.and_(bits, mask)
y      = pl.bitcast(masked, pl.FP32)
```

## Width rules

`strict` (default `True`) controls how strictly the widths must match:

| | `strict=True` (default) | `strict=False` |
| --- | --- | --- |
| Equal width (`f32`↔`i32/u32`, `bf16`↔`i16`, …) | accepted | accepted |
| Narrowing (`f32`→`f16`) | rejected, error points at `strict=False` / `cast` / `reshape` | accepted |
| Widening (`f16`→`f32`) | rejected | rejected |
| Same dtype | rejected | rejected |

Narrowing keeps the source shape, so the result covers only the leading bytes of the buffer. That is well defined only for a `row_major` / `none_box` tile (ND tensor) — under a fractal or boxed layout the element-to-byte mapping itself depends on the element width (`fractal` is a *byte* count), so narrowing is rejected there. Equal-width bitcasts are layout-agnostic and unaffected.

`tensor.bitcast` is **in-core only**: `ConvertTensorToTileOps` lowers it 1:1 to `tile.bitcast`, and orchestration codegen already rejects unregistered tensor ops with a clear "Misplaced tensor op" message.

## Lowering

Two cases, mirroring `tile.reshape` / `tile.transpose_view`:

- **PyPTO planner** — `set_output_memory_inherit_input()` makes InitMemRef share the source MemRef, so the result's own `pto.alloc_tile` at the shared address already carries the target dtype. Those two decls *are* the bitcast; **no instruction is emitted**.

  ```mlir
  %loaded = pto.alloc_tile addr = %c0_i64 valid_row = %c16 valid_col = %c128 : ...dtype=f32...
  %bits   = pto.alloc_tile addr = %c0_i64 valid_row = %c16 valid_col = %c128 : ...dtype=i32...
  ```

- **PTOAS planner** (`memory_planner=PTOAS`) — addr-less aliased vars collapse onto one `tile_buf` handle, so that second `alloc_tile` never appears and the bitcast lowers to `pto.treshape` + `pto.bitcast`. The `treshape` is required: `pto.bitcast` takes no `valid_row`/`valid_col` operands, so ptoas builds the destination tile from the result type alone, and a `v_row=?` type lowers to a default-constructed `Tile` whose `RowMaskInternal` / `ColMaskInternal` are left **uninitialized** (pto-isa `pto_tile.hpp`). Pinning the valid extent first keeps the view's extent well defined.

`tile.bitcast` is also added to `IsLegalTileViewOp` so the MemoryReuse load-derived hazard closure propagates through the view.

## Testing

- **Unit tests** (17): type deduction and every rejection path (equal-width pairs, narrowing, widening, same-dtype, non-`row_major` narrowing), the tensor→tile conversion, and **both** codegen lowerings — including the PTOAS-planner path that emits `pto.bitcast` with static valid dims.
- **System test** (`tests/st/runtime/ops/test_bitcast.py`, 5 cases) — validated on a2a3 hardware against `torch.Tensor.view(dtype)`, the exact bit-level reference. A value conversion or a lost alias would both fail it. Includes a BF16 → INT16 → `not_` → BF16 round trip that only reproduces the reference bits if both directions alias correctly.
- Generated `.pto` assembles with ptoas **0.45 / 0.48 / 0.50**.
- Full `tests/ut` suite green (7463 passed); `pre-commit` and `clang-tidy` clean.

## Docs

Developer reference (`docs/{en,zh-cn}/dev/ir/05-operators.md`) and user reference (`docs/{en,zh-cn}/user/02-operation_reference.md`) updated in both languages.


Hzfengsy:
Superseded by #2113, which landed `pl.reinterpret_view` as the dtype alias operator. Closing in favor of that implementation.

PR #2162, opened 2026-07-27
[RFC/NFM] docs: distributed programming guides — collectives, benchmarking, and API reference

georgebisbas:
## ⚠️ Superseded — split into four smaller PRs

This PR grew too large to review comfortably (36 files, ~3000 lines) and has
been split into four small, independently-reviewable PRs carrying the exact
same content:

1. **#2218** — Distributed op docstring & type-stub fixes (pure Python, ~184 lines)
2. **#2219** — Distributed Programming chapter (`docs/en/user/distributed/` + zh mirror)
3. **#2220** — Performance & Benchmarking chapter (`docs/en/user/performance/` + zh mirror)
4. **#2221** — Wire both chapters into the landing pages and operation reference
   (draft — depends on #2219 and #2220 merging first)

All four were cut from this branch's final commit and independently verified
(markdownlint, EN/ZH parity, nav coverage, `pre-commit run --all-files`,
`mkdocs build --strict`); a full end-to-end integration of all four together
was also verified to build cleanly with zero broken links and no lost or
duplicated content versus this branch.

Closing this PR without merging — please review and merge #2218–#2221 instead.

---

## Summary

Adds the **Distributed Programming** and **Performance & Benchmarking**
chapters to the PyPTO user manual, covering PyPTO's L3 distributed execution
model and its benchmarking/tuning workflow end to end. Structure and scope
follow the site's documentation plan and slot directly into the existing
`mkdocs.yml` nav (EN + ZH).

### Distributed Programming (`docs/en/user/distributed/`)

- **`index.md`** — PyPTO's distributed model in one page: symmetric memory
  + signals, comm domains, and how PyPTO's simplified L2 (single-chip) /
  L3 (cross-rank) vocabulary relates to the `simpler` runtime's own finer
  L0–L6 hierarchy.
- **`00-model.md`** — A minimal end-to-end `HelloAllReduce` walkthrough:
  host orchestrator, per-chip orchestration function, InCore kernel, window
  buffers, and how to launch it (a distributed program launches as a plain
  `python script.py` — PyPTO forks the per-chip workers internally, there's
  no external multi-process launcher).
- **`01-collectives.md`** — Every collective (`allreduce`, `barrier`,
  `broadcast`, `allgather`, `reduce_scatter`, `all_to_all`), the three
  execution paths (InCore hand-rolled, InCore composite, HOST builtin) and
  when to reach for each, supported `ReduceOp`/dtype per operation, and
  worked examples for each collective.
- **`02-primitives.md`** — The lower-level building blocks for anyone
  writing a custom protocol: `notify`/`wait` signal handshakes, window
  buffer allocation, tile-level RMA (`remote_load`/`remote_store`),
  one-sided `put`/`get` with their chunking/pipelining constraints, and how
  to compose primitives into a hand-rolled collective.
- **`03-execution.md`** — `DistributedWorker` lifecycle, `DistributedConfig`,
  and how a compiled distributed program is dispatched.
- **`04-debugging.md`** — Common failure modes (hangs, dispatch-time type
  errors, misconfigured `device_ids`) with symptoms and fixes.

### Performance & Benchmarking (`docs/en/user/performance/`)

- **`index.md` / `00-methodology.md`** — A decision tree for where to look
  first (compiler hints, span tree, op-simulator, memory map, dependency
  graph) and the programmatic `benchmark()` API: signature, `BenchmarkStats`
  fields, aggregates, and span-tree rendering.
- **`01-single-node.md`** — Single-node tuning: partitioning/parallelism
  (`pl.split`, `pl.spmd`, `pl.cluster`), pipelining/unrolling, the matmul
  path, memory placement (`target_memory`, `MemorySpace`), and scheduling
  controls (`predicate=`, `no_dep`, `manual_scope`).
- **`02-distributed.md`** — L3-specific performance: ring sizing and
  `RunConfig` prewarming, and a bus-bandwidth estimation formula.
- **`03-cases.md`** — End-to-end worked cases (status: additional cases
  planned).

Every page is mirrored in full under `docs/zh/user/...`.

### Operation Reference

`docs/en/user/02-operation_reference.md` gains a **Distributed / Collective
Operations** section: a capability matrix covering every `pld.*` op (API,
modes, `ReduceOp`, atomic support, supported dtypes per execution path, and
notes), cross-linked to the new chapters above.

### Site wiring

`mkdocs.yml`: `Distributed:` and `Performance:` nav sections under
`User Manual:`, with `nav_translations` entries for the ZH sidebar.

### Supporting docstring updates

A handful of docstrings in `tensor_ops.py`, `system_ops.py`, `tile_ops.py`,
`dsl_api.py`, `distributed_compiled_program.py`, `bench.py`, and the
`ir.pyi` type stub were tightened to match the semantics described in the
new guides (per-operation `ReduceOp` support, per-operation signal-shape
rank acceptance, and similar precision fixes) so the in-editor docs and the
manual agree.

## Verification

- `npx markdownlint-cli2` (project config)
- `python3 tests/lint/check_docs_en_zh_parity.py`
- `python3 tests/lint/check_docs_nav.py`
- `pre-commit run --all-files`
- `mkdocs build --strict`
- Every embedded Python code example checked for valid syntax
- Full CI (unit/system/dist-system tests, codegen tests, clang-tidy)

## What this does NOT cover

- Any runtime or compiler code changes — docs, docstrings, and one type
  stub only
- Full EN/ZH structural parity for every subsection of
  `performance/01-single-node.md` (the ZH mirror is intentionally more
  condensed in a few places; tracked as a follow-up)
- Automated CI verification of embedded doc code blocks (checked manually
  for this PR; a good candidate for a follow-up lint rule)

### DIFF
--- a/docs/en/user/02-operation_reference.md
+++ b/docs/en/user/02-operation_reference.md
@@ -277,3 +277,26 @@ scratch tile to materialize numeric results on A2/A3.
 | `create_tensor` | `(shape: Sequence[IntLike], dtype: DataType, layout: TensorLayout = None, init_value: int \| float \| None = None) -> Tensor` | Create tensor (promoted from `pl.tensor`; `init_value` AICPU-pre-fills the buffer — `0` zeroes any dtype, non-zero needs an int / 32-bit-or-wider float dtype) |
 | `max` | `(lhs: Scalar \| int \| Expr, rhs: Scalar \| int \| Expr) -> Scalar` | Max of two **scalars** (not a tile reduction — use `pl.tile.row_max` / `pl.tile.col_max`) |
 | `min` | `(lhs: Scalar \| int \| Expr, rhs: Scalar \| int \| Expr) -> Scalar` | Min of two **scalars** (not a tile reduction — use `pl.tile.row_min` / `pl.tile.col_min`) |
+
+## Distributed / Collective Operations (`pld.*`)
+
+Distributed and collective ops are in the `pld` namespace (`import pypto.language.distributed as pld`).
+Full reference at [distributed/index.md](distributed/index.md);
+tutorial at [distributed/00-model.md](distributed/00-model.md).
+
+### Capability Matrix
+
+| Operation | API | Modes | ReduceOp | Atomic | Supported dtypes | Notes |
+| --------- | --- | ----- | -------- | ------ | ---------------- | ----- |
+| AllReduce | `pld.tensor.allreduce` | `mesh` (InCore + HOST), `ring` (InCore only) | `Sum`, `Max`, `Min`, `Prod` | — | FP16, FP32 — a hard compile-time check (`allreduce.cpp`), both InCore and HOST | Mesh: O(N) remote traffic. Ring: O(N/P) traffic, 2(P-1) steps; currently InCore-only. |
+| AllGather | `pld.tensor.allgather` | — | — | — | FP32 only (HOST builtin); any GM dtype (InCore) | Push-based. Input and target must be different buffers. |
+| ReduceScatter | `pld.tensor.reduce_scatter` | — | `Sum` only | — | FP32 only (HOST builtin); any GM dtype (InCore) | Every rank stages all NR chunks before the call. |
+| Broadcast | `pld.tensor.broadcast` | — | — | — | FP32 only (HOST builtin); any GM dtype (InCore) | Root stages data before the call. |
+| All-to-All | `pld.tensor.all_to_all` | — | — | — | FP32 only (HOST builtin); any GM dtype (InCore) | Personalized exchange. Input and target must be different buffers. |
+| Barrier | `pld.tensor.barrier` | — | — | — | — | Signal is INT32, single-shot per call. |
+| Put | `pld.tensor.put` | — | — | `None_` / `Add` | All GM dtypes | `dst` must be window-bound. Supports chunked + pipelined staging. |
+| Get | `pld.tensor.get` | — | — | — | All GM dtypes | `src` must be window-bound. Supports chunked + pipelined staging. |
+| Notify | `pld.system.notify` | `AtomicAdd` / `Set` | — | — | — | Side-effect-only signal deposit. |
+| Wait | `pld.system.wait` | `Eq` / `Ge` | — | — | — | Side-effect-only signal block. |
+| Remote Load | `pld.tile.remote_load` | — | — | — | Any (tile) | Tile-level cross-rank load. |
+| Remote Store | `pld.tile.remote_store` | — | — | — | Any (tile) | Tile-level cross-rank store. |
--- a/docs/en/user/distributed/03-execution.md
+++ b/docs/en/user/distributed/03-execution.md
@@ -1,0 +1,166 @@
+# Execution
+
+A one-off compile-and-dispatch call is enough for a quick test. Production
+code instead amortizes setup — forking chip processes, assembling kernels —
+across many dispatches on a reusable `DistributedWorker`.
+
+## DistributedWorker
+
+Obtained via `compiled.prepare()`. Setup (fork, comm bootstrap, kernel assembly)
+happens once; dispatch happens many times.
+
+```python
+from pypto.runtime import DistributedWorker
+
+with DistributedWorker(compiled) as rt:
+    rt(host_x, host_out)
+    # ... more dispatches ...
+# rt.close() runs on exit — releases buffers and shuts down workers.
+```
+
+### Methods
+
+| Method | Description |
+| ------ | ----------- |
+| `compiled.prepare(config=None, callbacks=None)` | Create worker, fork chip processes, return `DistributedWorker`. Use as context manager. |
+| `rt(x, y, z)` | Single dispatch — coerces args, calls host_orch. |
+| `rt.run(compiled, x, y, z)` | Multi-program dispatch — selects the target program. |
+| `rt.alloc_tensor(shape, dtype, *, init=None)` | Allocate a worker-resident `DeviceTensor`. `init` copies from host (one-time H2D). |
+| `rt.free_tensor(tensor)` | Release a `DeviceTensor`. |
+| `rt.alloc_stacked_tensor(host_w)` | Shard host_w along dim 0 — shard `i` uploaded to card `i`. Returns `StackedDeviceTensor`. |
+| `rt.free_stacked_tensor(stacked)` | Release all shards of a `StackedDeviceTensor`. |
+| `rt.copy_stacked_from(stacked, host_out)` | D2H read-back of every shard into `host_out` (shared-memory, allocated before `prepare()`). |
+| `rt.release_inherited_host_tensor_refs()` | Drop runtime-held host references in the parent process after fork. |
+| `rt.close()` | Release buffers, shut down chip workers. Called automatically as context manager. |
+
+## DeviceTensor
+
+A worker-resident buffer that lives on the device across dispatches.
+When a `DeviceTensor` is passed as an argument to a compiled program,
+the runtime skips H2D/D2H copies — the device already has the data.
+
+```python
+import torch
+from pypto.runtime import DeviceTensor
+
+with compiled.prepare() as rt:
+    weight = rt.alloc_tensor((1024, 4096), torch.float16, init=host_weight)
+    rt(x, weight, out)   # dispatch via worker — no H2D/D2H
+```
+
+## StackedDeviceTensor
+
+Sharded across devices — obtained via `rt.alloc_stacked_tensor()`:
+
+```python
+# Host tensor sharded along dim 0 — shard[i] lives on card i.
+host_weights = torch.randn(4, 1024, 4096).share_memory_()  # 4 shards
+with compiled.prepare() as rt:
+    stacked = rt.alloc_stacked_tensor(host_weights)
+    rt(x, stacked, out)
+```
+
+> **Fatal pitfall:** `host_weights` must call `.share_memory_()` *before*
+> `prepare()`. The upload runs inside the already-forked chip worker, which
+> can only read host memory it inherited at fork — a plain `torch.Tensor`
+> raises `ValueError` at `alloc_stacked_tensor()`.
+
+## One-Shot vs Persistent Worker
+
+### One-Shot
+
+```python
+import torch
+from pypto.ir.distributed_compiled_program import DistributedConfig
+from pypto import ir
+
+dc = DistributedConfig(device_ids=[0, 1, 2, 3])
+compiled = ir.compile(HelloAllReduce, platform="a2a3", distributed_config=dc)
+
+inputs = torch.randn(4, 1, 256)
+outputs = torch.zeros_like(inputs)
+compiled(inputs, outputs)   # blocks until all ranks finish
+```
+
+### Persistent Worker (Repeated Dispatch)
+
+```python
+host_x = torch.zeros((4, 1, 256), dtype=torch.float32).share_memory_()
+host_out = torch.zeros_like(host_x).share_memory_()
+
+with DistributedWorker(compiled) as rt:
+    for step in steps:
+        host_x.copy_(next_input(step))
+        rt(host_x, host_out)
+        consume(host_out)
+```
+
+> **Fatal pitfall:** IO buffers passed to `DistributedWorker` must call
+> `.share_memory_()` before `prepare()`. If you forget, the runtime rejects
+> the buffer at dispatch time — the child processes cannot access the
+> parent's private memory.
+
+## Several Programs on One Worker
+
+A single `DistributedWorker` can dispatch multiple compiled programs:
+
+```python
+compiled_a = ir.compile(ProgramA, platform="a2a3", distributed_config=dc)
+compiled_b = ir.compile(ProgramB, platform="a2a3", distributed_config=dc)
+
+with compiled_a.prepare(extra_compiled=[compiled_b]) as rt:
+    rt.run(compiled_a, host_x, host_out)  # dispatch ProgramA
+    rt.run(compiled_b, host_x, host_out)  # dispatch ProgramB
+```
+
+The worker reuses its chip processes and comm setup — no fork penalty.
+`compiled_b` must be passed via `extra_compiled=` for `rt.run(compiled_b, ...)`
+to find it; passing an unregistered program raises `ValueError`. Preparing
+more than one program also puts the worker in multi-program mode, where the
+`rt(*args)` shortcut is ambiguous and raises `TypeError` — dispatch every
+program explicitly through `rt.run(...)`, including the primary one.
+
+## CLI Launch
+
+A distributed program launches the same way as a single-device one — plain
+`python script.py`. `DistributedConfig(device_ids=[...])` picks the rank
+count and devices; the runtime forks one worker process per device from this
+one Python process, so there is no separate multi-process launcher to invoke.
+
+```bash
+python script.py
+```
+
+## Environment Variables
+
+### Compile-Time Macros
+
+These are C preprocessor `#define` macros in `profiling_config.h`, **not environment variables**.
+They default to `1` (enabled) and are set at build time via CMake flags. Setting them as shell
+env vars has no effect.
+
+| Macro | Default | Effect |
+| ----- | ------- | ------ |
+| `SIMPLER_HOST_STRACE` | `1` (on) | Required at build time for `benchmark()` timing markers. Without it, `benchmark()` raises `RuntimeError`. |
+| `SIMPLER_DFX` | `1` (on) | Umbrella gate for device-side profiling (orchestrator/scheduler metrics, PMU counters, scope stats, swimlane trace). Sub-tier flags require this to be `1`. |
+
+### Runtime Environment Variable
+
+| Variable | Default | Effect |
+| -------- | ------- | ------ |
+| `SIMPLER_DEVICE_STRACE_ENABLE` | on (unset or non-`"0"`) | Runtime toggle for device-domain `[STRACE]` markers. Set to `0` to suppress device markers while keeping host markers. |
+
+### Benchmark Env Vars
+
+The `pypto-lib` golden benchmark harness reads `PYPTO_BENCH` /
+`PYPTO_BENCH_ROUNDS` / `PYPTO_BENCH_WARMUP` / `PYPTO_BENCH_RAW` — these are
+not defined or consumed anywhere in this repository. See `pypto-lib`'s own
+documentation for current defaults. `pypto.runtime.benchmark()` (this
+repo's own harness) is documented in [Performance](../performance/index.md).
+
+## See Also
+
+- [00-model](00-model.md) — Quickstart and model vocabulary
+- [04-debugging](04-debugging.md) — Common failure patterns
+- [Performance](../performance/index.md) — Benchmarking and tuning
+- [Getting Started](../00-getting_started.md) — Runtime setup
--- a/docs/zh/user/00-getting_started.md
+++ b/docs/zh/user/00-getting_started.md
@@ -130,6 +130,9 @@ PR #1177，在 `SIMPLER_DFX` 下默认开启）；用 simpler 的 `strace_timing
 
 ### 性能基准（`benchmark`）
 
+完整的基准测试指南见 [performance/00-methodology.md](performance/00-methodology.md)——
+涵盖程序化 `benchmark()` API、L3 分布式计时、bus-bandwidth 公式以及常见注意事项。
+
 对于 register-once + 多轮（rounds）模式，`pypto.runtime.benchmark` 封装了循环
 与聚合：它注册 *compiled* 一次并发起 `rounds` 次廉价 launch（不再每轮重付
 register/load），读取每次 launch 的 `[STRACE]` 标记并返回 `BenchmarkStats`：
@@ -203,6 +206,60 @@ dispatch 的量。当某 rank 每轮恰好只有 1 次 dispatch 时，求和即�
 确定，则 `stats.fallback_flattened` 被置位，per-rank / `union` 视图为空。
 
 ### 分布式（L3+）程序
+
+完整的分布式编程模型——从 `alloc_window_buffer` 到 `allreduce`、`barrier`、
+`broadcast` 等集合通信——见 [分布式编程](distributed/00-model.md)。下面展示
+InCore kernel（执行平面）的 mesh allreduce Hello World 示例；完整的可运行程序
+还需包含 host 编排器、`ir.compile` 及分布式 worker 设置，详见上述指南：
+
+```python
+import pypto.language as pl
+import pypto.language.distributed as pld
+
+NR = pl.dynamic("NR")
+
+@pl.program
+class HelloAllReduce:
+    @pl.function(type=pl.FunctionType.InCore)
+    def reduce_step(
+        self,
+        inp: pl.Tensor[[1, 256], pl.FP32],
+        out: pl.Out[pl.Tensor[[1, 256], pl.FP32]],
+        data: pl.InOut[pld.DistributedTensor[[1, 256], pl.FP32]],
+        signal: pl.InOut[pld.DistributedTensor[[NR, 1], pl.INT32]],
+    ) -> pl.Tensor[[1, 256], pl.FP32]:
+        ctx = pld.get_comm_ctx(data)
+        my_rank = pld.rank(ctx)
+        nranks = pld.nranks(ctx)
+
+        # 1. Stage-in：将本地输入复制到本 rank 的 window 分片。
+        data = pl.store(pl.load(inp, [0, 0], [1, 256]), [0, 0], data)
+
+        # 2. Barrier：通知每个对端，然后等待每个对端。
+        for peer in pl.range(nranks):
+            if peer != my_rank:
+                pld.system.notify(signal, peer=peer, offsets=[my_rank, 0],
+                                  value=1, op=pld.NotifyOp.AtomicAdd)
+        for src in pl.range(nranks):
+            if src != my_rank:
+                pld.system.wait(signal, offsets=[src, 0],
+                                expected=1, cmp=pld.WaitCmp.Ge)
+
+        # 3. 计算：加载自身分片，remote-load 每个对端，累加。
+        acc = pl.load(data, [0, 0], [1, 256])
+        for peer in pl.range(nranks):
+            if peer != my_rank:
+                peer_tile = pld.tile.remote_load(
+                    data, peer=peer, offsets=[0, 0], shape=[1, 256])
+                acc = pl.add(acc, peer_tile)
+
+        # 4. Stage-out：将累加器写入本地输出。
+        out = pl.store(acc, [0, 0], out)
+        return out
+```
+
+指南包含**逐行解读、ring allreduce 权衡、notify/wait 握手模式以及调试表格**。
+完整章节见 [distributed/index.md](distributed/index.md)。
 
 `ir.compile` 对 L3+ 分布式程序返回的 `DistributedCompiledProgram` 与 `CompiledProgram`
 一样接受 `DeviceTensor` 入参：用 worker 常驻 buffer 替代 `torch.Tensor`，runtime 即对该
--- a/python/pypto/language/distributed/op/tensor_ops.py
+++ b/python/pypto/language/distributed/op/tensor_ops.py
@@ -193,7 +193,7 @@ def alloc_window_buffer(  # type: ignore[no-redef]
     * **Shape+dtype convenience overload:**
       ``alloc_window_buffer(shape, *, dtype=..., name=...)`` — ``shape`` is a
       list / tuple of per-rank dimensions.  The byte size is computed
-      automatically as ``product(shape) × dtype.get_byte()`` and the call
+      automatically as ``product(shape) x dtype.get_byte()`` and the call
       normalizes to the canonical byte form.  ``dtype`` is required when
       ``shape`` is a sequence and rejected otherwise.
 
@@ -219,6 +219,17 @@ def alloc_window_buffer(  # type: ignore[no-redef]
         binds it to the LHS as a plain :class:`ir.Var`; passing that Var
         through :func:`window` materialises a :class:`DistributedTensor`
         view.
+
+    .. note::
+
+       This function is callable only inside HOST-level orchestration
+       (``level=pl.Level.HOST, role=pl.Role.Orchestrator``). Calling it
+       inside InCore (``type=pl.FunctionType.InCore``) raises a parser error.
+
+    .. seealso::
+
+       :func:`window` for creating typed DistributedTensor views over an
+       allocated buffer.
 
     Raises:
         ValueError: If ``name`` is empty (the parser must have injected it).
@@ -274,6 +285,11 @@ def window(
 
     Returns:
         A :class:`DistributedTensor` view of the given shape and dtype.
+
+    .. seealso::
+
+       :func:`alloc_window_buffer` for the two-phase ``alloc → window`` pattern.
+
     """
     buf_expr = _unwrap(buf)
     if not isinstance(buf_expr, Expr):
@@ -519,10 +535,33 @@ def allreduce(
         pub = pld.tensor.allreduce(pub, sig, op=pld.ReduceOp.Sum)
         pub = pld.tensor.allreduce(pub, sig, op=pld.ReduceOp.Sum, mode="ring")
 
+        # Mesh mode on the host orchestrator can omit the signal argument:
+        pub = pld.tensor.allreduce(pub, op=pld.ReduceOp.Sum)
+
+    **Signal shape:** host builtins accept rank-1 ``[world_size]`` or rank-2
+    ``[world_size, 1]`` (the compiler-synthesized signal is rank-2). InCore
+    composites take rank-2 ``[nranks, 1]`` for mesh -- the rank count may be
+    dynamic -- and ``[2*(NR-1), NR]`` for ring, where ``NR`` must be a
+    compile-time constant. Both are single-shot per call.
+
+    **Do not reuse the same signal buffer for a back-to-back allreduce**
+    — allocate a fresh signal buffer (``alloc_window_buffer`` + ``window``)
+    for each allreduce call. All allreduce calls in ``for`` and ``while``
+    loops are rejected because the current signal protocol cannot provide a
+    fresh signal for every dynamic iteration. A self-resetting variant is
+    blocked on a runtime fix — PTOAS issue #797.
+
+    .. seealso::
+
+        :func:`alloc_window_buffer` and :func:`window` for buffer allocation
+        and view creation.
+
+    .. rubric:: Implementation Notes
+
     LowerCompositeOps expands the explicit-signal InCore form into either:
     (a) a notify/wait ready barrier followed by UB-sized
     remote_load+accumulate/store chunks for ``mode="mesh"`` (default); or
-    (b) the NCCL-style 2(P−1)-step
+    (b) the NCCL-style 2(P-1)-step
     chunked reduce-scatter + allgather ring schedule for ``mode="ring"``.
     In both modes the kernel sees only the lowered primitives.
     Host-orchestrator code can omit ``signal``
@@ -531,26 +570,11 @@ def allreduce(
     (mesh mode only — ring mode on the HOST rail is delivered by a
     subsequent host builtin).
 
-    Mesh signal shape is ``[NR, 1]``; ring signal shape is
-    ``[2 * (NR − 1), NR]`` (one row per ring round). Both are single-shot
-    per call.
-
     Fully-valid packed mesh targets are viewed as one logical 1D stream and
     processed in chunks of at most 16 KiB. For statically known smaller targets,
     the physical chunk width shrinks to the smallest 32-byte-aligned width that
     covers the target. The final chunk uses ``valid_shape`` so arbitrary element
-    counts do not read or store past the end. Mesh lowering
-    also preserves a packed ND target ``TensorView.valid_shape`` when
-    its valid box can be represented by collapsing leading dimensions to one
-    2D rectangle and fits within one 16-KiB chunk, and reduces only that
-    rectangle with the established single-rectangle path. Oversized partial
-    rectangles, strided targets, DN partial views, and
-    non-representable partial boxes are rejected explicitly.
-    Any symbolic target or partial-valid extent that survives lowering must be
-    runtime-bound by a kernel scalar, loop variable, or physical tensor-shape
-    parameter; a symbol that appears only in type metadata is rejected during
-    PTO codegen. A fully dynamic physical target dimension is bound from that
-    tensor parameter.
+    counts do not read or store past the end.
 
     **Mesh barrier protocol:** ``AtomicAdd(1) → WaitGe(1)`` is the ready wave.
     Every reduced chunk then performs ``AtomicAdd(1)`` and waits for the
@@ -633,13 +657,16 @@ def barrier(
     """Cross-rank barrier synchronisation.
 
     Blocks until all ranks in the comm group have reached the barrier.
-    Uses a window-bound INT32 ``signal`` matrix for cross-rank
-    synchronisation (one slot per rank).  LowerCompositeOps expands this
+    Uses a window-bound INT32 ``signal`` tensor for cross-rank
+    synchronisation.  LowerCompositeOps expands this
     into a notify-all / wait-all sequence.
 
     .. code-block:: python
 
         sig = pld.tensor.barrier(sig)
+
+    **Signal shape:** host builtins require rank-1 ``[world_size]``. InCore
+    composites take rank-2 ``[nranks, 1]`` -- the rank count may be dynamic.
 
     **Signal buffer is single-shot per call.**  The lowering uses
     ``Set(1)`` + ``Ge(1)`` — cells go from 0 to 1.  Do not reuse the
@@ -668,8 +695,11 @@ def broadcast(
     """Broadcast root rank's data to all ranks.
 
     After this call returns, every rank's slice of ``target`` holds
-    root's data.  Uses a window-bound INT32 ``signal`` matrix for the
+    root's data.  Uses a window-bound INT32 ``signal`` tensor for the
     cross-rank barrier.
+
+    **Signal shape:** host builtins require rank-1 ``[world_size]``. InCore
+    composites take rank-2 ``[nranks, 1]`` -- the rank count may be dynamic.
 
     .. code-block:: python
 
@@ -716,6 +746,10 @@ def allgather(
     plain :class:`pl.Tensor` ``[1, SIZE]`` — both are accepted.  HOST vs InCore
     is a function-context property resolved by the lowering passes.
 
+    **Signal shape:** host builtins accept rank-1 ``[world_size]`` or rank-2
+    ``[world_size, 1]``. InCore composites take rank-2 ``[nranks, 1]`` -- the
+    rank count may be dynamic.
+
     Args:
         local_data: This rank's single chunk — ``[1, SIZE]`` :class:`pl.Tensor`
             (InCore) or ``[1, SIZE]`` :class:`pld.DistributedTensor` staging
@@ -751,6 +785,9 @@ def reduce_scatter(
             data = pl.store(chunk_j, [j, 0], data)
         data = pld.tensor.reduce_scatter(data, sig, op=pld.ReduceOp.Sum)
         # data[my_rank, 0:SIZE] now holds this rank's reduced chunk.
+
+    **Signal shape:** host builtins require rank-1 ``[world_size]``. InCore
+    composites take rank-2 ``[nranks, 1]`` -- the rank count may be dynamic.
 
     Args:
         target: Window-bound :class:`pld.DistributedTensor` of shape
@@ -790,6 +827,10 @@ def all_to_all(
     by an earlier InCore step) rather than the InCore composite's plain
     :class:`pl.Tensor` — both are accepted.
 
+    **Signal shape:** host builtins accept rank-1 ``[world_size]`` or rank-2
+    ``[world_size, 1]``. InCore composites take rank-2 ``[nranks, 1]`` -- the
+    rank count may be dynamic.
+
     Args:
         input: [NR, SIZE] Tensor or DistributedTensor with per-destination
             chunks, distinct from ``target``.  ``input[dest, :]`` is the
@@ -797,7 +838,9 @@ def all_to_all(
         target: :class:`pld.DistributedTensor` [NR, SIZE] window that receives
             the result in-place.  After the call,
             ``target[src, :]`` holds the chunk received from rank ``src``.
-        signal: :class:`pld.DistributedTensor` [NR, 1] INT32 barrier.
+        signal: Window-bound INT32 :class:`pld.DistributedTensor` barrier
+            tensor.  Rank-1 ``[world_size]`` or rank-2 ``[world_size, 1]``
+            for host builtins; rank-2 ``[nranks, 1]`` for InCore composites.
 
     Returns:
         The ``target`` :class:`pld.DistributedTensor` (window-as-result).
--- a/python/pypto/language/dsl_api.py
+++ b/python/pypto/language/dsl_api.py
@@ -712,6 +712,19 @@ def spmd(
     ``range(n)``. Loop start is fixed at 0 and step at 1; each block gets an
     index ``i`` in ``[0, core_num)``.
 
+    .. rubric:: Three usage forms at a glance
+
+    ==============================  ===================================================
+    Form                             Description
+    ==============================  ===================================================
+    ``with pl.spmd(n):``             Dispatch or inline block (no captured TaskId).
+    ``for i in pl.spmd(n):``         Loop-style; ``i`` = per-block index, body
+                                    is auto-outlined to InCore.
+    ``with pl.spmd(n) as tid:``      Same body shapes as form 1, plus captured
+                                    producer TaskId in ``tid``.  Optionally pass
+                                    ``deps=[...]`` after ``n``.
+    ==============================  ===================================================
+
     Usage forms:
 
     1. ``with pl.spmd(n):`` — body is either a *dispatch* body calling a
--- a/python/pypto/runtime/bench.py
+++ b/python/pypto/runtime/bench.py
@@ -303,6 +303,23 @@ _METRIC_ATTR = {"device": "device_wall_us", "host": "host_wall_us", "effective":
 class BenchmarkStats:
     """Aggregated per-launch timing from :func:`benchmark`.
 
+    .. rubric:: Quick-reference
+
+    ============================== =========================================================
+    Accessor                       Description
+    ============================== =========================================================
+    ``stats.device_us_median``     Median device wall (µs).
+    ``stats.device_us_min``        Minimum device wall (µs).
+    ``stats.device_us_max``        Maximum device wall (µs).
+    ``stats.device_us_mean``       Arithmetic mean device wall (µs).
+    ``stats.device_us_stdev``      Std-dev of device wall (µs).
+    ``stats.all_zero_device``      True if every sample is 0 (e.g. sim builds).
+    ``stats.samples``              Alias for ``device_wall_us`` (the raw list).
+    ``stats.per_round("device")``  List[float]: device wall per round.
+    ``stats.per_rank("device")``   Dict[int, List[float]]: per-rank breakdown (L3).
+    ``stats.print_tree()``         Render per-dispatch span tree to stdout.
+    ============================== =========================================================
+
     The min / median / mean / max / stdev helpers operate on
     ``device_wall_us`` — the on-NPU metric. ``host_wall_us`` samples are kept
     for context, but they include per-launch arg coercion + H2D and so are not
--- a/docs/en/user/00-getting_started.md
+++ b/docs/en/user/00-getting_started.md
@@ -145,6 +145,10 @@ and read `l2_swimlane_records.json`.
 
 ### Benchmarking (`benchmark`)
 
+The full benchmarking guide is at [performance/00-methodology.md](performance/00-methodology.md)
+— it covers the programmatic `benchmark()` API, L3 distributed timing,
+bus-bandwidth formulas, and common caveats.
+
 For the register-once + rounds pattern, `pypto.runtime.benchmark` owns the loop
 and aggregation: it registers *compiled* once and dispatches `rounds` cheap
 launches (no per-round register/load), reads each launch's `[STRACE]` markers,
@@ -228,6 +232,61 @@ dispatch shape is non-deterministic, `stats.fallback_flattened` is set and the
 per-rank / `union` views are empty.
 
 ### Distributed (L3+) programs
+
+The complete distributed programming model — from `alloc_window_buffer` through
+collectives like `allreduce`, `barrier`, and `broadcast` — is covered in the
+[Distributed Programming](distributed/00-model.md). Here's the minimal mesh allreduce
+Hello World:
+
+```python
+import pypto.language as pl
+import pypto.language.distributed as pld
+
+NR = pl.dynamic("NR")
+
+@pl.program
+class HelloAllReduce:
+    @pl.function(type=pl.FunctionType.InCore)
+    def reduce_step(
+        self,
+        inp: pl.Tensor[[1, 256], pl.FP32],
+        out: pl.Out[pl.Tensor[[1, 256], pl.FP32]],
+        data: pl.InOut[pld.DistributedTensor[[1, 256], pl.FP32]],
+        signal: pl.InOut[pld.DistributedTensor[[NR, 1], pl.INT32]],
+    ) -> pl.Tensor[[1, 256], pl.FP32]:
+        ctx = pld.get_comm_ctx(data)
+        my_rank = pld.rank(ctx)
+        nranks = pld.nranks(ctx)
+
+        # 1. Stage-in: copy local input into this rank's window slice.
+        data = pl.store(pl.load(inp, [0, 0], [1, 256]), [0, 0], data)
+
+        # 2. Barrier: notify every peer, then wait on every peer.
+        for peer in pl.range(nranks):
+            if peer != my_rank:
+                pld.system.notify(signal, peer=peer, offsets=[my_rank, 0],
+                                  value=1, op=pld.NotifyOp.AtomicAdd)
+        for src in pl.range(nranks):
+            if src != my_rank:
+                pld.system.wait(signal, offsets=[src, 0],
+                                expected=1, cmp=pld.WaitCmp.Ge)
+
+        # 3. Compute: load own slice, remote-load every peer, accumulate.
+        acc = pl.load(data, [0, 0], [1, 256])
+        for peer in pl.range(nranks):
+            if peer != my_rank:
+                peer_tile = pld.tile.remote_load(
+                    data, peer=peer, offsets=[0, 0], shape=[1, 256])
+                acc = pl.add(acc, peer_tile)
+
+        # 4. Stage-out: store accumulator to output.
+        out = pl.store(acc, [0, 0], out)
+        return out
+```
+
+The guide includes a **line-by-line walkthrough, ring allreduce trade-offs,
+notify/wait handshake patterns, and a debugging table**. The full chapter is
+at [distributed/index.md](distributed/index.md).
 
 L3+ distributed programs returned by `ir.compile` (a `DistributedCompiledProgram`)
 accept `DeviceTensor` arguments the same way as `CompiledProgram`: pass a
--- a/docs/en/user/distributed/00-model.md
+++ b/docs/en/user/distributed/00-model.md
@@ -1,0 +1,193 @@
+# Distributed Programming Model
+
+> **Prerequisites:** [Getting Started](../00-getting_started.md) — basic PyPTO
+> tensor/tile model. This guide uses the `pld` namespace
+> (`import pypto.language.distributed as pld`).
+
+## Quickstart: 2-Rank AllReduce
+
+The simplest distributed program — two ranks sum their data, both see the same result.
+
+```python
+import pypto.language as pl
+import pypto.language.distributed as pld
+
+NR = pl.dynamic("NR")
+SIZE = 256
+
+@pl.program
+class HelloAllReduce:
+    @pl.function(type=pl.FunctionType.InCore)
+    def reduce_step(
+        self,
+        inp: pl.Tensor[[1, SIZE], pl.FP32],
+        out: pl.Out[pl.Tensor[[1, SIZE], pl.FP32]],
+        data: pl.InOut[pld.DistributedTensor[[1, SIZE], pl.FP32]],
+        signal: pl.InOut[pld.DistributedTensor[[NR, 1], pl.INT32]],
+    ) -> pl.Tensor[[1, SIZE], pl.FP32]:
+        ctx = pld.get_comm_ctx(data)
+        my_rank = pld.rank(ctx)
+        nranks = pld.nranks(ctx)
+
+        # 1. Stage-in: copy local input into this rank's window slice.
+        local = pl.load(inp, [0, 0], [1, SIZE])
+        data = pl.store(local, [0, 0], data)
+
+        # 2. Barrier: notify every peer, then wait on every peer.
+        for peer in pl.range(nranks):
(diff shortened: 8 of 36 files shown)


georgebisbas:
## AI Review Comment Fixes — All Addressed

All 17 review comments from CodeRabbit and chatgpt-codex-connector have been addressed in the latest commit. Here's a summary:

### chatgpt-codex-connector (6 comments — already fixed in previous commits, verified clean):

1. **Import `import pypto.language.pld`** → Changed to `import pypto.language.distributed as pld` in both EN and zh-cn.
2. **Dtype matrix overstates HOST-mode support** → Now distinguishes FP32 (HOST builtin) from FP16/BF16 (InCore hand-rolled).
3. **`prepare()` incorrectly lists `inherited_host_tensors=`** → Removed; it only exists on `DistributedWorker()`.
4. **L3 benchmark claims silent corruption on non-shared tensors** → Now correctly documents that `DistributedWorker.__call__` raises `TypeError`.
5. **DFX flags documented as "not plumbed through L3"** → Now correctly documents that `enable_l2_swimlane` IS plumbed.
6. **`ring_heap=128` is not a valid power-of-two** → Changed to `ring_heap=1024`.

### CodeRabbit (11 comments — all fixed in this commit):

1. **Put's `AtomicType.None_`/`Add` in ReduceOps column** → Added a dedicated "Atomic" column to the capability matrix.
2. **`all_zero_device` mentions "non-STRACE builds"** → Removed; non-STRACE raises `RuntimeError`, zero-valued stats occur only on `*sim` platforms.
3. **DeviceTensor example not executable** → Added `import torch`, wrapped in `compiled.prepare()` context, dispatch via `rt(...)`.
4. **`prepare()` API contract** → Already fixed (verified clean).
5. **`release_inherited_host_tensor_refs()` over-promises destruction** → Now says "drop runtime-held references", not "free the host copy".
6. **`SIZE` undefined in HelloAllReduce** → Added `SIZE = 256` in both EN and zh-cn.
7. **Handshake example missing tag >= 1 precondition** → Added callout in both languages.
8. **STRACE behavior inconsistency** → `all_zero_device` description aligned with the caveat (non-STRACE raises, `*sim` returns zero).
9. **zh-cn "可直接复制运行" misleading** → Clarified this is an InCore kernel sample requiring additional setup.
10. **Unicode math symbols (×, −)** → Replaced with ASCII `x` and `-` in `tensor_ops.py`.
11. **spmd table shows deps as required** → Updated to show `with pl.spmd(n) as tid:` with optional `deps=[...]`.

### Nit:

- **Markdownlint bare code blocks** → Added `text`/`python` language tags to bare fences in EN benchmarking guide.

docs/en/user/04-distributed-api-reference.md
```diff
@@ -0,0 +1,252 @@
+# Distributed API Reference
+
+All distributed operations are accessed via `import pypto.language.pld`. System-level ops
```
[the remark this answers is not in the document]
georgebisbas: Fixed: changed `import pypto.language.pld` to `import pypto.language.distributed as pld` in both EN and zh-cn.

docs/en/user/02-operation_reference.md
```diff
@@ -277,3 +277,26 @@ scratch tile to materialize numeric results on A2/A3.
...
+
+### Capability Matrix
+
+| Operation | API | Modes | ReduceOps | Supported dtypes | Notes |
+|-----------|-----|-------|-----------|-----------------|-------|
+| AllReduce | `pld.tensor.allreduce` | `mesh` (InCore + HOST), `ring` (InCore only) | `Sum` only | FP16, FP32, BF16 | Mesh: O(N) remote traffic. Ring: O(N/P) traffic, 2(P−1) steps. HOST ring builtin on a separate branch — not yet in this checkout. |
+| AllGather | `pld.tensor.allgather` | — | — | FP16, FP32 | Push-based. Input and target must be different buffers. |
+| ReduceScatter | `pld.tensor.reduce_scatter` | — | `Sum` only | FP16, FP32 | Every rank stages all NR chunks before the call. |
+| Broadcast | `pld.tensor.broadcast` | — | — | FP16, FP32 | Root stages data before the call. |
+| All-to-All | `pld.tensor.all_to_all` | — | — | FP16, FP32 | Personalized exchange. Input and target must be different buffers. |
```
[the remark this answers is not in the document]
georgebisbas: Fixed: the capability matrix now distinguishes FP32 (HOST builtin) from FP16/FP32/BF16 (InCore hand-rolled) for each collective.

docs/en/user/04-distributed-api-reference.md
```diff
@@ -0,0 +1,252 @@
...
+```
+
+### DistributedWorker
+
+Obtained via `compiled.prepare()`. Setup (fork, comm bootstrap, kernel assembly)
+happens once; dispatch happens many times.
+
+| Method | Description |
+|--------|-------------|
+| `compiled.prepare(config=None, callbacks=None, inherited_host_tensors=None)` | Create worker, fork chip processes, return `DistributedWorker`. Use as context manager. |
```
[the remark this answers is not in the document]
georgebisbas: Fixed: removed `inherited_host_tensors` from `compiled.prepare()` signature in both EN and zh-cn API references. It only exists on `DistributedWorker()`.

docs/en/user/06-benchmarking-guide.md
```diff
@@ -0,0 +1,386 @@
...
+# Shared-memory host tensors — MUST call .share_memory_() before benchmark().
+host_x = torch.zeros((4, 1, 256), dtype=torch.float32).share_memory_()
+host_out = torch.zeros_like(host_x).share_memory_()
+
+stats = benchmark(compiled, (host_x, host_out), rounds=100, warmup=3)
+```
+
+> **Fatal pitfall:** If you forget `.share_memory_()`, the benchmark passes
+> silently but measures stale data — the forked chip workers write to private
+> copies, not the parent's tensor. There is no error or warning.
```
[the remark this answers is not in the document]
georgebisbas: Fixed: the warning now correctly states that `DistributedWorker.__call__` raises `TypeError` for non-shared-memory tensors.

docs/en/user/06-benchmarking-guide.md
```diff
@@ -0,0 +1,386 @@
...
+device = stats.per_round("device")    # list[float], length = rounds
+union  = stats.per_round("union")     # list[float], L3 only
+```
+
+### DFX Flag Limitation
+
+The runtime's DFX flags (`RunConfig(enable_l2_swimlane=True)`) enable per-task
+timing inside the worker, but the swimlane is **not plumbed through L3** in the
+current runtime. For L3 per-dispatch profiling, use the span-tree `invocations`
+data (caller-visible per-dispatch markers) rather than runtime-side DFX.
```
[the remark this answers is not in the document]
georgebisbas: Fixed: the DFX section now correctly documents that `enable_l2_swimlane` IS plumbed through L3. Also cleaned up stale leftover text in both languages.

docs/en/user/06-benchmarking-guide.md
```diff
@@ -0,0 +1,386 @@
...
+For L3 ring allreduce, you can tune the ring-task window and heap sizes
+at dispatch time via `RunConfig`. These are passed at `prepare()` time, not
+compile time:
+
+```python
+from pypto.runtime import RunConfig
+from pypto import ir
+
+compiled = ir.compile(MyRingProgram, platform="a2a3", distributed_config=dc)
+worker = compiled.prepare(config=RunConfig(ring_task_window=256, ring_heap=128))
```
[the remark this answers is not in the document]
georgebisbas: Fixed: `ring_heap=128` changed to `ring_heap=1024` (valid power-of-two >= 1024).

docs/en/user/02-operation_reference.md
```diff
@@ -277,3 +277,26 @@ scratch tile to materialize numeric results on A2/A3.
...
+| Barrier | `pld.tensor.barrier` | — | — | — | Signal is INT32, single-shot per call. |
+| Put | `pld.tensor.put` | — | `None_` / `Add` | All GM dtypes | `dst` must be window-bound. Supports chunked + pipelined staging. |
+| Get | `pld.tensor.get` | — | — | All GM dtypes | `src` must be window-bound. Supports chunked + pipelined staging. |
```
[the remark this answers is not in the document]
georgebisbas: Fixed: added a dedicated "Atomic" column to separate Put's `AtomicType.None_`/`Add` from the `ReduceOp` column in both EN and zh-cn capability matrices.

docs/en/user/04-distributed-api-reference.md
```diff
@@ -0,0 +1,252 @@
...
+### simpler Compile-Time Macros
+
+These are C preprocessor `#define` macros in `profiling_config.h`, **not environment variables**.
+They default to `1` (enabled) and are set at build time via CMake flags
+(e.g. `-DSIMPLER_HOST_STRACE=1`). Setting them as shell env vars has no effect.
+
+| Macro | Default | Effect |
+|-------|---------|--------|
+| `SIMPLER_HOST_STRACE` | `1` (on) | Required at build time for `benchmark()` timing markers. When enabled, the runtime emits `[STRACE]` spans to stderr at `LOG_INFO_V9` tier. Without it, `benchmark()` raises `RuntimeError`. |
+| `SIMPLER_DFX` | `1` (on) | Umbrella gate for device-side profiling (orchestrator/scheduler metrics, PMU counters, scope stats, swimlane trace). With `SIMPLER_DFX=0`, all profiling code is compiled out. Sub-tier flags (`SIMPLER_ORCH_PROFILING`, `SIMPLER_SCHED_PROFILING`) require this to be `1`. |
```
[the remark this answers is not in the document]
georgebisbas: Fixed: removed "or non-STRACE builds" from `all_zero_device` description. Non-STRACE builds raise `RuntimeError`, so zero-valued stats only occur on `*sim` platforms.

docs/en/user/04-distributed-api-reference.md
```diff
@@ -0,0 +1,252 @@
...
+the runtime skips H2D/D2H copies — the device already has the data.
+
+```python
+from pypto.runtime import DeviceTensor
+
+# After prepare(), build a DeviceTensor referencing device memory.
+weight = rt.alloc_tensor((1024, 4096), torch.float16, init=host_weight)
+# Pass it like a torch.Tensor — no H2D/D2H.
+compiled(x, weight, out)
+```
```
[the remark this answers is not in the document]
georgebisbas: Fixed: DeviceTensor example now imports `torch`, wraps allocation inside `compiled.prepare()` context, and dispatches via `rt(...)`. The zh-cn version also had a malformed closing tag that was fixed.

docs/en/user/04-distributed-api-reference.md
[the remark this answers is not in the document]
georgebisbas: Fixed (verified): `inherited_host_tensors` was already removed from `prepare()` in a previous commit. Current docs show `prepare(config=None, callbacks=None)`.

docs/en/user/04-distributed-api-reference.md
```diff
@@ -0,0 +1,252 @@
...
+|--------|-------------|
+| `compiled.prepare(config=None, callbacks=None, inherited_host_tensors=None)` | Create worker, fork chip processes, return `DistributedWorker`. Use as context manager. |
+| `rt(x, y, z)` | Single dispatch — coerces args, calls host_orch. |
+| `rt.run(compiled, x, y, z)` | Multi-program dispatch — selects the target program. |
+| `rt.alloc_tensor(shape, dtype, *, init=None)` | Allocate a worker-resident `DeviceTensor`. `init` copies from host (one-time H2D). |
+| `rt.free_tensor(tensor)` | Release a `DeviceTensor`. |
+| `rt.alloc_stacked_tensor(host_w)` | Shard host_w along dim 0 — shard `i` uploaded to card `i`. Returns `StackedDeviceTensor`. |
+| `rt.free_stacked_tensor(stacked)` | Release all shards of a `StackedDeviceTensor`. |
+| `rt.copy_stacked_from(stacked, host_out)` | D2H read-back of every shard into `host_out` (shared-memory, allocated before `prepare()`). |
+| `rt.release_inherited_host_tensor_refs()` | Drop parent-process host references after fork (free the host copy). |
```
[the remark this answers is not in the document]
georgebisbas: Fixed: `release_inherited_host_tensor_refs()` now says "drop runtime-held host references" instead of promising unconditional host-tensor destruction. Updated in both languages.

docs/en/user/05-distributed-guide.md
```diff
@@ -0,0 +1,576 @@
...
+
+@pl.program
+class HelloAllReduce:
+    @pl.function(type=pl.FunctionType.InCore)
+    def reduce_step(
+        self,
+        inp: pl.Tensor[[1, SIZE], pl.FP32],
+        out: pl.Out[pl.Tensor[[1, SIZE], pl.FP32]],
+        data: pl.InOut[pld.DistributedTensor[[1, SIZE], pl.FP32]],
+        signal: pl.InOut[pld.DistributedTensor[[NR, 1], pl.INT32]],
```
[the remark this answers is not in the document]
georgebisbas: Fixed: added `SIZE = 256` definition before HelloAllReduce in both EN and zh-cn distributed guides.

docs/en/user/05-distributed-guide.md
```diff
@@ -0,0 +1,576 @@
...
+        # 1. Write our tag into the peer's signal cell.
+        pld.system.notify(
+            signal, peer=peer, offsets=[0, 0],
+            value=tag, op=pld.NotifyOp.Set,
+        )
+
+        # 2. Wait until our own cell has been written.
+        pld.system.wait(
+            signal=signal, offsets=[0, 0],
+            expected=1, cmp=pld.WaitCmp.Ge,
```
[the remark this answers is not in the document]
georgebisbas: Fixed: added a callout documenting the tag>=1 precondition: the wait uses `Ge` with `expected=1`, so passing `tag=0` causes permanent hang.

docs/en/user/06-benchmarking-guide.md
```diff
@@ -0,0 +1,386 @@
...
+| Field | Type | Description |
+|-------|------|-------------|
+| `device_wall_us` | `list[float]` | Per-round on-NPU device wall (µs). L2: one per launch (the `device_wall` span). L3: per-round max across ranks. |
+| `host_wall_us` | `list[float]` | Per-round host wall (µs). Includes arg coercion and H2D overhead — the device metric is the primary figure. |
+| `rounds` | `int` | Number of measured launches (warmup excluded). |
+| `warmup` | `int` | Number of leading launches discarded. |
+| `invocations` | `list[TraceInvocation]` | Full per-launch span trees (one per launch for L2; every rank's per-dispatch for L3). |
+| `rounds_dispatches` | `list` | L3 only — round → rank → [dispatch] grid. Empty for L2. |
+| `fallback_flattened` | `bool` | L3 only — `True` when non-deterministic dispatch shapes prevented per-round segmentation. |
+| `all_zero_device` | `bool` | `True` when every `device_wall_us` sample is `0` — typical on `*sim` platforms or non-STRACE builds. Check this before interpreting device metrics. |
```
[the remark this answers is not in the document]
georgebisbas: Fixed: `all_zero_device` description aligned with the caveat -- non-STRACE raises, only `*sim` returns zero-valued stats.

docs/zh/user/00-getting_started.md
```diff
@@ -400,6 +404,59 @@ dispatch 的量。当某 rank 每轮恰好只有 1 次 dispatch 时，求和即
...
+                acc = pl.add(acc, peer_tile)
+
+        # 4. Stage-out：将累加器写入本地输出。
+        out = pl.store(acc, [0, 0], out)
+        return out
+```
+
+指南包含**逐行解读、ring allreduce 权衡、notify/wait 握手模式以及调试表格**。
+API 参考见 [04-distributed-api-reference.md](04-distributed-api-reference.md)。
+
```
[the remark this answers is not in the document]
georgebisbas: Fixed: changed wording from "directly copy-paste run" to clarify this is an InCore kernel sample needing additional host/runtime setup.

python/pypto/language/distributed/op/tensor_ops.py
```diff
@@ -519,6 +535,28 @@ def allreduce(
         pub = pld.tensor.allreduce(pub, sig, op=pld.ReduceOp.Sum)
         pub = pld.tensor.allreduce(pub, sig, op=pld.ReduceOp.Sum, mode="ring")
 
+        # Mesh mode on the host orchestrator can omit the signal argument:
+        pub = pld.tensor.allreduce(pub, op=pld.ReduceOp.Sum)
+
+    **Signal shapes:**
+    Mesh: ``[NR, 1]`` — one global barrier covers all peer reads.
+    Ring: ``[2 × (NR − 1), NR]`` — one fresh row per ring step.
+    Both are single-shot per call. Host-level builtins use 1-D ``[world_size]``.
```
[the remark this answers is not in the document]
georgebisbas: Fixed: replaced Unicode × and − with ASCII x and - in `tensor_ops.py` docstrings.

python/pypto/language/dsl_api.py
```diff
@@ -712,6 +712,18 @@ def spmd(
...
+    .. rubric:: Three usage forms at a glance
+
+    ==============================  ===================================================
+    Form                             Description
+    ==============================  ===================================================
+    ``with pl.spmd(n):``             Dispatch or inline block (no captured TaskId).
+    ``for i in pl.spmd(n):``         Loop-style; ``i`` = per-block index, body
+                                    is auto-outlined to InCore.
+    ``with pl.spmd(n, deps=[...])``  Same body shapes as form 1, plus captured
+      ``as tid:``                    producer TaskId in ``tid``.
```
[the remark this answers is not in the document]
georgebisbas: Fixed: updated spmd table to show `with pl.spmd(n) as tid:` with optional `deps=[...]`.

docs/en/user/04-distributed-api-reference.md
```diff
@@ -0,0 +1,252 @@
...
+
+| Name | Signature | Description |
+| ---- | --------- | ----------- |
+| `world_size` | `() -> Scalar` | **Host-only.** Number of ranks in the distributed execution. Returns `INT64`. |
+| `get_comm_ctx` | `(dist_tensor: DT) -> Ctx` | Lift a `DistributedTensor` to its `CommCtx` handle. The verifier rejects plain `pl.Tensor`. |
+| `rank` | `(ctx: Ctx) -> Scalar` | Local rank index (`INT32`). Lowers to a load of `CommContext::rankId`. |
+| `nranks` | `(ctx: Ctx) -> Scalar` | Number of ranks in this comm group (`INT32`). Lowers to a load of `CommContext::rankNum`. |
+| `window` | `(buf: Ptr, shape: Sequence[IntLike], *, dtype: DataType) -> DT` | Materialise a window-buffer `Ptr` as a `DistributedTensor` view. `buf` comes from `alloc_window_buffer`. |
+| `alloc_window_buffer` | `(size: IntLike, *, name: str = "") -> Ptr` | Allocate a per-rank HCCL window buffer. **Size is in bytes.** The `name` kwarg is injected by the parser from the LHS assignment — never pass it explicitly. |
+| `alloc_window_buffer` | `(shape: Sequence[IntLike], *, dtype: DataType, name: str = "") -> Ptr` | Convenience overload. `size = prod(shape) × dtype.get_byte()` computed automatically. |
```
YunjiQin: **Wrong namespace.** `window` and `alloc_window_buffer` live in `pld.tensor.*`, not `pld.system.*`:

```python
# python/pypto/language/distributed/op/system_ops.py:170
__all__ = ["get_comm_ctx", "notify", "nranks", "rank", "wait", "world_size"]

# python/pypto/language/distributed/op/tensor_ops.py:806
__all__ = ["all_to_all", "alloc_window_buffer", "allgather", "allreduce",
           "barrier", "broadcast", "get", "put", "reduce_scatter", "window"]
```

This also contradicts this page's own short-form table further down, which correctly maps `pld.alloc_window_buffer(...)` -> `pld.tensor.alloc_window_buffer(...)`. Please move these three rows into the "Tensor Collectives (`pld.tensor.*`)" section (or give them their own `pld.tensor.*` window-management subsection).

docs/en/user/04-distributed-api-reference.md
```diff
@@ -0,0 +1,252 @@
...
+| `NotifyOp` | `AtomicAdd`, `Set` | Signal deposit mode. `AtomicAdd`: atomically increment the peer's signal slot (use for multi-rank barriers). `Set`: overwrite the peer's signal slot (use for 1:1 handshakes). |
+| `WaitCmp` | `Eq`, `Ge` | Wait predicate. `Eq`: block until signal slot equals expected value. `Ge`: block until signal slot >= expected value. |
+| `ReduceOp` | `Sum`, `Max`, `Min`, `Prod` | Reduction operator for collective operations. First-version lowering supports `Sum` only; `Max`/`Min`/`Prod` are reserved and currently rejected. |
+| `AtomicType` | `None_`, `Add` | Remote-store combine mode. `None_` (alias for `pld.AtomicType.None_`): plain store. `Add`: atomically accumulate into peer's destination. |
+| `DistributedTensor` | — | A tensor view bound to a comm-domain window buffer. Every collective and RMA op requires this type on the window side. |
+| `CommCtx` | — | Communication context handle. Produced by `get_comm_ctx()`; consumed by `rank()` and `nranks()`. |
+| `ScopeMode` | `AUTO`, `MANUAL` | Dependency-tracking mode for `pl.scope()`. `AUTO`: OverlapMap tracks deps. `MANUAL`: user declares every edge with `pl.submit(..., deps=...)`. |
```
YunjiQin: `ScopeMode` is `pl.ScopeMode`, not `pld.ScopeMode` — it is exported from `python/pypto/language/__init__.py:228` (`from .scope import ScopeMode, manual_scope, scope, spmd_submit, submit`) and is absent from `pypto/language/distributed/__init__.py`'s `__all__`.

It is already listed correctly in the `pl.*` table below, so this row can just be dropped.

docs/en/user/04-distributed-api-reference.md
```diff
@@ -0,0 +1,252 @@
...
+> for full safety rules and debugging advice.
+
+## Types and Enums (`pld.*`)
+
+| Name | Values | Description |
+| ---- | ------ | ----------- |
+| `NotifyOp` | `AtomicAdd`, `Set` | Signal deposit mode. `AtomicAdd`: atomically increment the peer's signal slot (use for multi-rank barriers). `Set`: overwrite the peer's signal slot (use for 1:1 handshakes). |
+| `WaitCmp` | `Eq`, `Ge` | Wait predicate. `Eq`: block until signal slot equals expected value. `Ge`: block until signal slot >= expected value. |
+| `ReduceOp` | `Sum`, `Max`, `Min`, `Prod` | Reduction operator for collective operations. First-version lowering supports `Sum` only; `Max`/`Min`/`Prod` are reserved and currently rejected. |
+| `AtomicType` | `None_`, `Add` | Remote-store combine mode. `None_` (alias for `pld.AtomicType.None_`): plain store. `Add`: atomically accumulate into peer's destination. |
```
YunjiQin: Minor: `` `None_` (alias for `pld.AtomicType.None_`) `` is self-referential — it defines the symbol in terms of itself. Presumably the intent was to note that `None_` is spelled with a trailing underscore because `None` is a Python keyword.

docs/en/user/04-distributed-api-reference.md
```diff
@@ -0,0 +1,252 @@
...
+| `get` | `(dst: DT \| Tensor, peer: IntLike, src: DT, dst_offsets=None, src_offsets=None, shape=None, *, chunk_rows=0, chunk_cols=0, pipeline=False) -> Call` | `dst: Out`, `src: In` | Read peer rank's `src` into local `dst`. `src` **must** be window-bound; `dst` may be plain `Tensor`. With no offsets/shape, reads the full peer slice. |
+| `allreduce` | `(target: DT, *, op: ReduceOp = Sum, mode: str = "mesh") -> DT` | `target: InOut` | In-place cross-rank allreduce. `mode="mesh"` (default): all-to-all exchange — O(N) remote traffic per step, one global barrier. `mode="ring"`: 2(P−1)-step reduce-scatter + allgather — O(N/P) remote traffic per step, more barriers. See the [Distributed Guide](05-distributed-guide.md) for when to use each. Host-orchestrator code may omit `signal` outside loops; the compiler synthesizes one of shape `[world_size(), 1]` (mesh only). |
+| `allreduce` | `(target: DT, signal: DT, *, op: ReduceOp = Sum, mode: str = "mesh") -> DT` | `target: InOut`, `signal: InOut` | Explicit-signal form. Mesh signal shape: `[NR, 1]`. Ring signal shape: `[2 × (NR − 1), NR]`. Signal is single-shot — allocate a fresh buffer per call. |
```
YunjiQin: **The host-side signal rank is not 1-D only.** `LowerHostTensorCollectives` accepts *both* rank-1 and rank-2 signals:

```cpp
// src/ir/transforms/lower_host_tensor_collectives_pass.cpp:101-110
CHECK_SPAN(signal_type->shape_.size() == 1 || signal_type->shape_.size() == 2, call->span_)
    << "LowerHostTensorCollectives: collective signal must be rank-1 [world_size] "
       "or rank-2 [world_size, 1]";
```

and the compiler-synthesized signal on the host rail is itself rank-2 (`synthesize_allreduce_signals_pass.cpp:243-247` builds `MakeTuple{world_size_var, 1}`) — consistent with the "synthesizes one of shape `[world_size(), 1]`" sentence in the row above, but inconsistent with the "host builtins use 1-D" claim repeated in six `tensor_ops.py` docstrings.

Suggest re-framing by **mode / layer** rather than host-vs-InCore:

- host builtins: rank-1 `[world_size]` **or** rank-2 `[world_size, 1]` (synthesized form is rank-2)
- InCore mesh composite: `[nranks, 1]` — rank count may be dynamic
- InCore ring composite: `[2*(NR-1), NR]` — `NR` must be a compile-time constant

See also the thread on `tensor_ops.py`.

docs/en/user/04-distributed-api-reference.md
```diff
@@ -0,0 +1,252 @@
...
+| `cluster` | `(*, name_hint: str = "") -> ClusterContext` | Group co-scheduled AIC (Cube) and AIV (Vector) kernels sharing physical cluster resources. |
+| `submit` | `(*args: Any, deps=None, dumps=None, allow_early_resolve=False, predicate=None) -> Any` | Submit a kernel and capture its producer TaskId. Parser construct — only valid inside `@pl.function`. Returns a 2-tuple `(result, task_id)`. |
+| `spmd_submit` | `(*args: Any, core_num: int, sync_start=False, deps=None, allow_early_resolve=False, predicate=None) -> Any` | SPMD variant of `submit`. Launches a kernel across `core_num` blocks. Returns `(result, task_id)`. |
+| `scope` | `(mode: ScopeMode = AUTO) -> scope` | Runtime scope context manager. `AUTO`: OverlapMap tracks deps (default). `MANUAL`: user must declare every edge. |
+| `manual_scope` | `() -> manual_scope` | Alias for `pl.scope(mode=pl.ScopeMode.MANUAL)`. Turns off auto dep-tracking for the region. |
+| `ScopeMode` | enum | `AUTO`, `MANUAL` |
```
YunjiQin: **This whole section looks out of scope for a distributed API reference.** `spmd` / `cluster` / `submit` / `spmd_submit` / `scope` / `manual_scope` / `ScopeMode` are L2 orchestration-layer task-graph constructs; none of them are rank-aware or related to cross-rank communication. They already belong to `01-language_guide.md` / `02-operation_reference.md`.

Duplicating them here creates a second source of truth that will drift — the `ScopeMode`-under-`pld.*` error above is exactly that drift already happening. Suggest deleting the section and cross-referencing the language guide instead.

docs/en/user/04-distributed-api-reference.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Intra-Device Cross-Core Ops (`pl.system.*`)
```
YunjiQin: **Also out of scope.** `tpush_to_aiv` / `tpop_from_aic` / `sync_set` / `syncall` / `reserve_buffer` / `import_peer_buffer` are single-device AIC<->AIV cross-core pipeline primitives — intra-chip, not cross-rank.

The section opens with a disclaimer explaining that these are *not* distributed ops and are "distinct from `pld.system.*`". Needing a disclaimer to justify the placement is a good sign the content belongs elsewhere.

There are also accuracy gaps that will have to be maintained here for no benefit — e.g. `syncall` actually takes `scratch` / `scratch_l1` (`python/pypto/language/op/system_ops.py:111-120`), and every op takes `span`. Suggest deleting the section.

docs/en/user/04-distributed-api-reference.md
```diff
@@ -0,0 +1,252 @@
...
+| `SIMPLER_HOST_STRACE` | `1` (on) | Required at build time for `benchmark()` timing markers. When enabled, the runtime emits `[STRACE]` spans to stderr at `LOG_INFO_V9` tier. Without it, `benchmark()` raises `RuntimeError`. |
+| `SIMPLER_DFX` | `1` (on) | Umbrella gate for device-side profiling (orchestrator/scheduler metrics, PMU counters, scope stats, swimlane trace). With `SIMPLER_DFX=0`, all profiling code is compiled out. Sub-tier flags (`SIMPLER_ORCH_PROFILING`, `SIMPLER_SCHED_PROFILING`) require this to be `1`. |
+
+### simpler Runtime Env Var
+
+| Variable | Default | Effect |
+| -------- | ------- | ------ |
+| `SIMPLER_DEVICE_STRACE_ENABLE` | on (unset or non-`"0"`) | Runtime toggle for device-domain `[STRACE]` markers (those with `clk=dev`). Set to `0` to suppress device markers while keeping host markers. Host markers are controlled by the `SIMPLER_HOST_STRACE` compile-time macro. |
+
+### pypto-lib Benchmark Env Vars
```
YunjiQin: **These belong to the upper repo, not this one.** `PYPTO_BENCH` / `PYPTO_BENCH_ROUNDS` / `PYPTO_BENCH_WARMUP` / `PYPTO_BENCH_RAW` have zero definitions anywhere in pypto:

```
$ grep -rn "PYPTO_BENCH_ROUNDS\|PYPTO_BENCH_WARMUP\|PYPTO_BENCH_RAW" --include=*.py .
(no matches)
```

They are defined and consumed by the pypto-lib golden harness and do not affect any behaviour in this repository. Documenting them here (in a *distributed API reference*, no less) means the defaults and output format will silently drift from their real definitions.

Suggest dropping this table entirely from this page. In `06-benchmarking-guide.md` it can stay if it is explicitly labelled as pypto-lib-owned and links out rather than transcribing the values.

docs/en/user/04-distributed-api-reference.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Environment Variables
+
+Env vars that affect distributed execution behaviour.
+
+### simpler Compile-Time Macros
+
+These are C preprocessor `#define` macros in `profiling_config.h`, **not environment variables**.
+They default to `1` (enabled) and are set at build time via CMake flags
+(e.g. `-DSIMPLER_HOST_STRACE=1`). Setting them as shell env vars has no effect.
```
YunjiQin: `cmake -DSIMPLER_HOST_STRACE=1` does not do anything today — no `CMakeLists.txt` or `*.cmake` in this repo (including `runtime/`) reads either name:

```
$ grep -rn "SIMPLER_HOST_STRACE\|SIMPLER_DFX" --include=CMakeLists.txt --include=*.cmake .
(no matches)
```

The macros are `#ifndef`-guarded defines in `runtime/src/common/task_interface/profiling_config.h:15-24`, so they have to arrive as compile definitions (`add_compile_definitions` / `-D` on the compiler command line). A bare `cmake -D...` only sets a cache variable and is silently ignored.

The "these are compile-time macros, not env vars" point is correct and valuable — it is just the suggested invocation that is wrong. Same issue at `05-distributed-guide.md:553-559`.

docs/en/user/05-distributed-guide.md
```diff
@@ -0,0 +1,581 @@
...
+            pld.world_size() * pl.INT32.get_byte(),
+        )
+        signal = pld.window(sig_buf, [NR, 1], dtype=pl.INT32)
+
+        for r in pl.range(pld.world_size()):
+            inp_r = pl.slice(inputs, [1, 1, SIZE], [r, 0, 0])
+            out_r = pl.slice(outputs, [1, 1, SIZE], [r, 0, 0])
+            out_r = self.reduce_step(inp_r, out_r, data, signal, device=r)
+            pl.store(out_r, [r, 0, 0], outputs)
+        return outputs
```
YunjiQin: **This orchestrator will not compile as written.** Four issues, all diverging from the reference implementation in `tests/st/distributed/collectives/test_l3_allreduce.py:127-150`:

1. **Missing the chip-orchestration layer.** A HOST orchestrator dispatches a `FunctionType.Orchestration` function with `device=r`, which in turn calls the InCore kernel — it does not dispatch InCore directly. Every reference test follows this (`test_l3_allreduce.py:118-125` `chip_orch`, `test_l3_host_tensor_allreduce.py:49/66`, `test_l3_host_tensor_barrier.py:48/67`).

2. **Rank mismatch on the slice.** `pl.slice(inputs, [1, 1, SIZE], [r, 0, 0])` produces `[1, 1, SIZE]`, but `reduce_step` takes `[1, SIZE]`. Needs `drop_dims=[0]`, or just the idiomatic `inputs[r]` the tests use.

3. **`pl.store(out_r, [r, 0, 0], outputs)`** stores a *tensor* where a tile is expected, and `out_r` is already the write target — the reference passes `outputs[r]` straight into the dispatch and returns `outputs`.

4. **`pld.window(sig_buf, [NR, 1], ...)`** uses the `pl.dynamic("NR")` symbol as a runtime shape value. The tests use `pld.window(signal_buf, [pld.world_size(), 1], dtype=pl.INT32)`.

Given this is the very first end-to-end example a reader meets, I would suggest lifting `host_orch` verbatim from `test_l3_allreduce.py` rather than writing a fresh variant — that also keeps it under test.

docs/en/user/05-distributed-guide.md
```diff
@@ -0,0 +1,581 @@
...
+
+PyPTO has two ways to do the same collective, and choosing the right one depends
+on where you want to write the logic:
+
+| Aspect | InCore Hand-Rolled | Host-Level Builtin |
+| ------ | ------------------ | ------------------ |
+| **Where** | `@pl.function(type=pl.FunctionType.InCore)` | `@pl.function(level=pl.Level.HOST, role=pl.Role.Orchestrator)` |
+| **How** | You write notify/wait + `remote_load` loops | You call `pld.tensor.allreduce(data, sig, ...)` |
+| **Signal shape** | 2-D: `[NR, 1]` (mesh) or `[2(NR−1), NR]` (ring) | 1-D: `[world_size]` (host builtins) |
+| **NR** | `pl.dynamic("NR")` (mesh only) | Implicit from `pld.world_size()` |
```
YunjiQin: **The InCore-vs-Host dichotomy this section is built on does not hold.** InCore is not limited to hand-rolled notify/wait — the `pld.tensor.*` composite intrinsics are callable directly from `FunctionType.InCore` functions and are covered by a dedicated ST suite:

| test | InCore-level call |
| --- | --- |
| `collectives/test_l3_tensor_allreduce_intrinsic.py:94` | `pld.tensor.allreduce(staged_data, signal, op=Sum)` |
| `collectives/test_l3_tensor_allreduce_ring_intrinsic.py:89` | `pld.tensor.allreduce(data, signal, op=Sum, mode="ring")` |
| `collectives/test_l3_tensor_barrier_intrinsic.py:74` | `pld.tensor.barrier(signal)` |
| `collectives/test_l3_tensor_allgather_intrinsic.py:72` | `pld.tensor.allgather(inp, data, signal)` |

(all under `@pl.function(type=pl.FunctionType.InCore)`, lowered by `LowerCompositeOps`).

Consequences for the table:

- **How** — "InCore: you write notify/wait + `remote_load` loops" is wrong; InCore can be a one-liner.
- **NR** — `pl.dynamic("NR")` availability is a **mesh vs ring** property, not a hand-rolled vs builtin one. Mesh never inspects the signal shape and loops on the runtime `pld.nranks(ctx)` (`lower_composite_ops_pass.cpp:222`), so dynamic NR is fine for the InCore mesh composite too. Ring needs a static `NR` (`lower_composite_ops_pass.cpp:948-966` folds `SIZE // NR` precisely to avoid a dynamic `FloorDiv` that `InitMemRef` cannot handle).
- **Signal shape** — see the thread on `04`; the host rail accepts rank-1 *and* rank-2.
- **When to use** — "use InCore only for teaching / custom protocols" steers readers away from the InCore composite, which is the **only** path that supports `mode="ring"` (as `02-operation_reference.md:291` in this same PR correctly states: `ring` (InCore only)).

The real axis is *which layer the composite is called from* — InCore composite (`LowerCompositeOps`) vs HOST (`LowerHostTensorCollectives` -> `builtin.tensor.*`) — with hand-rolled notify/wait as a third, lower-level option.

docs/en/user/05-distributed-guide.md
```diff
@@ -0,0 +1,581 @@
...
+
+```bash
+python script.py                                          # single-node
+torchrun --nproc_per_node=4 script.py                     # multi-process
+```
+
+> **Important:** IO buffers passed to `DistributedWorker` must call
+> `.share_memory_()` before `prepare()`. If you forget, the output is stale —
+> the child writes to a private copy, not the parent's tensor. This passes
+> silently with no error.
```
YunjiQin: **Contradicts `06-benchmarking-guide.md:185-188`**, which says a missing `.share_memory_()` "raises a `TypeError`". The runtime is explicit, so this note is the incorrect one:

```python
# python/pypto/runtime/distributed_runner.py:1734-1740
elif isinstance(arg, torch.Tensor):
    if not arg.is_shared():
        raise TypeError(
            f"Parameter {info.name!r}: a host torch.Tensor passed to a DistributedWorker "
            "must be shared memory allocated BEFORE prepare() (call .share_memory_() and "
            "reuse the same buffer across dispatches), so the forked chip worker can see it."
        )
```

It fails loudly at the Python level; there is no silent-stale-output mode. Worth fixing since "this passes silently with no error" would send readers hunting for a bug that the runtime already reports.

docs/en/user/05-distributed-guide.md
```diff
@@ -0,0 +1,581 @@
...
+(`#define` in `profiling_config.h`), not environment variables. They default to
+`1` (enabled) and are set at build time via CMake flags:
+
+```bash
+# Enable STRACE markers at build time (default on):
+cmake ... -DSIMPLER_HOST_STRACE=1
+
+# Disable device profiling at build time:
+cmake ... -DSIMPLER_DFX=0
+```
```
YunjiQin: Same as the `04` thread: no CMake code in this repo reads `SIMPLER_HOST_STRACE` / `SIMPLER_DFX`, so these two invocations are no-ops. The macros are `#ifndef` defines in `runtime/src/common/task_interface/profiling_config.h:15-24` and need to reach the compiler as definitions.

docs/en/user/05-distributed-guide.md
```diff
@@ -0,0 +1,581 @@
+# Distributed Programming Guide
```
YunjiQin: Length: this file is 581 lines (zh-cn 555), over the 500-line cap in `.claude/rules/documentation-length.md` ("condense at 500-700, split above 700").

The reworked "InCore vs Host" section will shrink it somewhat. Beyond that, "Hello World" and "Write Your Own Primitive: From Notify/Wait to Collective" walk through nearly the same notify/wait + `remote_load` code twice — merging them, or moving the ring algorithm walkthrough to its own page, should bring it under the cap.

docs/en/user/06-benchmarking-guide.md
```diff
@@ -0,0 +1,385 @@
...
+from pypto.runtime import RunConfig
+from pypto import ir
+
+compiled = ir.compile(MyRingProgram, platform="a2a3", distributed_config=dc)
+worker = compiled.prepare(config=RunConfig(ring_task_window=256, ring_heap=1024))
+```
+
+The `prepare(ring_task_window=...)` call prewarms the ring scheduler with the
+specified window size, so the first dispatch avoids cold-setup overhead.
+Subsequent dispatches reuse the prewarmed state.
```
YunjiQin: **`prepare()` has no `ring_task_window` kwarg, and there is no prewarm.**

```python
# python/pypto/ir/distributed_compiled_program.py:366-374
def prepare(self, config=None, *, extra_compiled=(), persistent=False,
            callbacks=None, sub_worker_overrides=None) -> "DistributedWorker":
```

`ring_task_window` / `ring_heap` are `RunConfig` fields forwarded per invocation into `CallConfig.runtime_env`:

```python
# python/pypto/runtime/runner.py:776-779
if run_config.ring_task_window is not None:
    call_config.runtime_env.ring_task_window = run_config.ring_task_window
if run_config.ring_heap is not None:
    call_config.runtime_env.ring_heap = run_config.ring_heap
```

The code block at 345-351 is correct; the paragraph after it describes behaviour that does not exist. Suggest replacing it with the actual constraints, which are enforced and worth documenting (`runner.py:307-312`): `ring_task_window` must be a power of 2 >= 4, `ring_heap` a power of 2 >= 1024 bytes; both also accept a per-ring list.

The section title "Ring Sizing and **Prewarm**" would need updating too.

docs/en/user/06-benchmarking-guide.md
```diff
@@ -0,0 +1,385 @@
...
+PYPTO_BENCH=1 python my_kernel.py
+```
+
+### Environment Variables
+
+| Variable | Default | Description |
+| -------- | ------- | ----------- |
+| `PYPTO_BENCH` | — | Set to `1` to enable the benchmark harness. |
+| `PYPTO_BENCH_ROUNDS` | `100` | Number of measured launches. |
+| `PYPTO_BENCH_WARMUP` | `3` | Leading launches discarded for cache/page warm-up. |
```
YunjiQin: These four variables are defined and consumed by the pypto-lib golden harness, not by anything in this repository:

```
$ grep -rn "PYPTO_BENCH_ROUNDS\|PYPTO_BENCH_WARMUP\|PYPTO_BENCH_RAW" --include=*.py .
(no matches)
```

Transcribing another repo's defaults and expected stdout here guarantees drift. Suggest replacing the table with a one-line pointer ("provided by the pypto-lib benchmark harness — see <link>") and keeping this guide focused on `pypto.runtime.benchmark()`, which *is* this repo's contract. The same table is duplicated in `04-distributed-api-reference.md:168-176` and should be dropped there outright.

docs/en/user/02-operation_reference.md
```diff
@@ -277,3 +277,26 @@ scratch tile to materialize numeric results on A2/A3.
...
+
+Distributed and collective ops are in the `pld` namespace (`import pypto.language.distributed as pld`).
+Full reference at [04-distributed-api-reference.md](04-distributed-api-reference.md);
+tutorial at [05-distributed-guide.md](05-distributed-guide.md).
+
+### Capability Matrix
+
+| Operation | API | Modes | ReduceOp | Atomic | Supported dtypes | Notes |
+| --------- | --- | ----- | -------- | ------ | ---------------- | ----- |
+| AllReduce | `pld.tensor.allreduce` | `mesh` (InCore + HOST), `ring` (InCore only) | `Sum` only | — | FP32 (HOST builtin). FP16, FP32, BF16 (InCore hand-rolled) | Mesh: O(N) remote traffic. Ring: O(N/P) traffic, 2(P−1) steps. HOST ring builtin on a separate branch — not yet in this checkout. |
```
YunjiQin: Two things in this row:

1. **"HOST ring builtin on a separate branch — not yet in this checkout"** — branch/worktree state does not belong in canonical user documentation; it is stale the moment the branch lands or is abandoned. If the limitation needs stating, phrase it as a capability ("`ring` is currently InCore-only") without referencing a branch.

2. **The dtype column** — I could not find backing for the `FP16, FP32, BF16 (InCore hand-rolled)` restriction. `LowerCompositeOps` handles the mesh target generically by dtype width (`lower_composite_ops_pass.cpp:699-705`: rejects only dtypes with no storage width, wider than the chunk budget, or not dividing the tile alignment) — no dtype allowlist. Either point at the check that enforces this list, or soften it to "not restricted by the lowering; validated per-dtype in <test>".

Note the "Modes" column here (`ring` (InCore only)) is correct and contradicts the InCore-vs-Host framing in `05-distributed-guide.md:361-371` — see that thread.

python/pypto/language/distributed/op/tensor_ops.py
```diff
@@ -626,14 +635,17 @@ def barrier(
...
-    synchronisation (one slot per rank).  LowerCompositeOps expands this
+    Uses a window-bound INT32 ``signal`` tensor for cross-rank
+    synchronisation.  LowerCompositeOps expands this
     into a notify-all / wait-all sequence.
 
     .. code-block:: python
 
         sig = pld.tensor.barrier(sig)
 
+    **Signal shape:** 1-D ``[world_size]`` for host-level builtins;
```
YunjiQin: This "1-D for host builtins / 2-D for InCore composites" sentence is added in six places in this file (`barrier` 646, `broadcast` 679, `allgather` 727, `reduce_scatter` 766, `all_to_all` 807 + its `Args: signal` at 816, plus the `allreduce` variant at 541-544), and mirrored into `04`/`05`. The host half is not accurate:

```cpp
// src/ir/transforms/lower_host_tensor_collectives_pass.cpp:101-110
CHECK_SPAN(signal_type->shape_.size() == 1 || signal_type->shape_.size() == 2, call->span_)
    << "LowerHostTensorCollectives: collective signal must be rank-1 [world_size] "
       "or rank-2 [world_size, 1]";
```

Both ranks are accepted, and the signal the compiler synthesizes for a host-orchestrator `allreduce` is itself rank-2 (`synthesize_allreduce_signals_pass.cpp:243-247`). Written as an either/or rule, this reads as "passing a 2-D signal on the host rail is illegal", which it is not.

The InCore half is also finer-grained than `[NR, 1]` suggests — mesh never reads the signal shape and loops on the runtime `pld.nranks(ctx)`, so the rank count may be dynamic; ring requires a compile-time-constant `NR` (`lower_composite_ops_pass.cpp:919`, `948-966`).

Suggested replacement (applies to all six sites):

```
**Signal shape:** host builtins accept rank-1 ``[world_size]`` or rank-2
``[world_size, 1]`` (the compiler-synthesized signal is rank-2). InCore
composites take rank-2 ``[nranks, 1]`` for mesh -- the rank count may be
dynamic -- and ``[2*(NR-1), NR]`` for ring, where ``NR`` must be a
compile-time constant.
```

Side note, not for this PR: ring's static-`NR` requirement has no `CHECK_SPAN` behind it. A symbolic `NR` falls into the `MakeFloorDiv` branch at `lower_composite_ops_pass.cpp:966` and then fails inside `InitMemRef` rather than with a message pointing at the ring limitation.

YunjiQin (commented):
Thanks for putting these together — the distributed and benchmarking surface has needed user-facing documentation for a while, and the structure here is good. I checked the prose against the implementation rather than reading for style; most of it holds up.

**Verified correct** (spot-checking the claims that are easiest to get wrong):

- C1 `pl.Level.HOST` + `pl.Role.Orchestrator` — matches `decorator.py:776` and `ir.pyi:837/896`.
- C2 `SIMPLER_HOST_STRACE` / `SIMPLER_DFX` as compile-time macros defaulting to `1` — `profiling_config.h:15-24`; `SIMPLER_DEVICE_STRACE_ENABLE` really is a `getenv` (`c_api_shared.cpp:474`). Good catch, this one was genuinely misdocumented before.
- `benchmark()` signature — byte-for-byte with `bench.py:840-849`; every `BenchmarkStats` field and property in the guide exists.
- Ring signal `[2(NR-1), NR]` + the factory-function pattern — matches `test_l3_allreduce_ring.py:71-84,205`.
- allreduce-inside-a-loop rejection — `synthesize_allreduce_signals_pass.cpp:190`.
- `pld.*` short-form / no-short-form table — exactly matches `distributed/__init__.py`'s `__all__`.
- `RunConfig(ring_task_window=256, ring_heap=1024)` — both satisfy the power-of-2 constraints in `runner.py:307-312`.

**Blocking, in rough priority order** (details inline):

1. `05` — the "InCore Hand-Rolled vs Host-Level Builtin" section is built on a dichotomy that does not exist; InCore composites are a first-class path with their own ST suite.
2. `05` — the Hello World HOST orchestrator will not compile (missing chip-orch layer, rank mismatch, tensor-into-`pl.store`, `pl.dynamic` used as a runtime shape).
3. `05` vs `06` — contradictory claims about a missing `.share_memory_()`; the runtime raises `TypeError`.
4. `tensor_ops.py` x6 + `04`/`05` — the "1-D host / 2-D InCore" signal-shape rule; the host rail accepts both.
5. `04` — `window` / `alloc_window_buffer` filed under `pld.system.*`; they are `pld.tensor.*`.
6. `06` — `prepare(ring_task_window=...)` and the "prewarm" paragraph describe an API that does not exist.
7. `04` / `05` — `cmake -DSIMPLER_HOST_STRACE=1` is a no-op; nothing in the build reads it.

**Scope / hygiene:**

8. `04` — the `pl.*` DSL-constructs and `pl.system.*` cross-core sections are L2 / intra-chip material and do not belong in a distributed API reference.
9. `04` (and to a lesser degree `06`) — the `PYPTO_BENCH*` table documents pypto-lib-owned variables that have no effect in this repo.
10. `04` — `ScopeMode` is `pl.ScopeMode`; `bench.py` — the new RST table does not parse; `05` is 581 lines against the 500-line cap in `.claude/rules/documentation-length.md`.

All of the above apply equally to `docs/zh-cn/user/`, which mirrors the English text faithfully — including the errors. Worth fixing both sides in the same pass.

python/pypto/runtime/bench.py
```diff
@@ -303,6 +303,23 @@ def _span_names() -> dict[str, str]:
...
+    Accessor                    Description
+    ========================== =========================================================
+    ``stats.device_us_median``  Median device wall (µs).
+    ``stats.device_us_min``     Minimum device wall (µs).
+    ``stats.device_us_max``     Maximum device wall (µs).
+    ``stats.device_us_mean``    Arithmetic mean device wall (µs).
+    ``stats.device_us_stdev``   Std-dev of device wall (µs).
+    ``stats.all_zero_device``   True if every sample is 0 (e.g. sim builds).
+    ``stats.samples``           Alias for ``device_wall_us`` (the raw list).
+    ``stats.per_round("device")``  List[float]: device wall per round.
```
YunjiQin: This RST simple table is malformed — the column-1 marker is 26 characters, but two cells overrun it:

```
========================== =========================================================
                                                     <- col 2 starts at column 28
``stats.per_round("device")``   <- 29 chars
``stats.per_rank("device")``    <- 28 chars
```

docutils reports "text in column margin" for cells that cross the boundary, so the table will not render. Widening the first column marker to 30 `=` (and re-aligning the rows) fixes it.

The `pl.spmd` table added in `dsl_api.py` is fine — its 30-char first column clears the longest cell.

docs/en/user/04-distributed-api-reference.md
georgebisbas: Fixed in 60c1b490 - moved `window`/`alloc_window_buffer` out of the "System Substrate (`pld.system.*`)" table into a new "Window Buffer Management (`pld.tensor.*`)" subsection in `distributed/02-primitives.md` (mirrored in `docs/zh/`).

docs/en/user/04-distributed-api-reference.md
georgebisbas: Already resolved — the restructure commit (f3705b32) dropped the `pld.*` types table that had this row. `ScopeMode` isn't referenced under `pld.*` anywhere in the current `distributed/` chapter (only correctly as `pl.ScopeMode` in `performance/01-single-node.md`).

docs/en/user/04-distributed-api-reference.md
georgebisbas: Already resolved — the restructure commit dropped the self-referential wording. Current text in `distributed/02-primitives.md` just says "`None_`: plain store."

docs/en/user/04-distributed-api-reference.md
georgebisbas: Fixed in 60c1b490 - `distributed/01-collectives.md`'s InCore-vs-Host table and `tensor_ops.py`'s signal-shape docstrings now state host builtins accept rank-1 `[world_size]` or rank-2 `[world_size, 1]` (the compiler-synthesized signal is rank-2), the InCore mesh composite is `[nranks, 1]` (rank count may be dynamic), and the InCore ring composite is `[2*(NR-1), NR]` (`NR` must be a compile-time constant).

docs/en/user/04-distributed-api-reference.md
georgebisbas: Already resolved — the restructure commit dropped the "Distributed DSL Constructs (`pl.*`)" section entirely. `pl.spmd`/`pl.submit`/`pl.scope` etc. aren't documented in the `distributed/` chapter.

docs/en/user/04-distributed-api-reference.md
georgebisbas: Already resolved — the restructure commit dropped the "Intra-Device Cross-Core Ops (`pl.system.*`)" section entirely.

docs/en/user/04-distributed-api-reference.md
georgebisbas: Fixed in 60c1b490 - dropped the `PYPTO_BENCH*` table from `distributed/03-execution.md`, replaced with a one-line pointer noting these vars belong to the `pypto-lib` benchmark harness, not this repo (confirmed via repo-wide grep, including the `runtime/` submodule — no definitions here).

docs/en/user/04-distributed-api-reference.md
georgebisbas: Fixed in 60c1b490 - removed the non-functional `cmake -D...` example from `distributed/04-debugging.md` (confirmed no `CMakeLists.txt`/`*.cmake` anywhere in the repo reads either macro name — the real mechanism is a `CXX="g++ -D..."` compiler-flag override into `simpler_setup/build_runtimes.py`, per `runtime/.github/workflows/ci.yml`). Replaced with a note that this is `simpler`-runtime build config, linked out rather than restated.

docs/en/user/05-distributed-guide.md
georgebisbas: Fixed in 60c1b490 - `distributed/00-model.md`'s Quickstart example now matches the working reference `tests/st/distributed/collectives/test_l3_allreduce.py::AllReduceMesh`: adds the missing `Orchestration`-type `chip_orch` dispatch layer between HOST and the InCore kernel, indexes `inputs[r]`/`outputs[r]` instead of the rank-mismatched `pl.slice`, drops the redundant post-dispatch `pl.store`, and uses `pld.world_size()` for the signal window shape instead of the symbolic `NR`.

docs/en/user/05-distributed-guide.md
georgebisbas: Fixed in 60c1b490 - replaced the InCore-hand-rolled-vs-HOST-builtin dichotomy in `distributed/01-collectives.md` with the real three-way split (hand-rolled / InCore composite via `LowerCompositeOps` / HOST builtin via `LowerHostTensorCollectives`), since the InCore composite supports both `mesh` and `ring` while the HOST builtin path only lowers `mesh`.

docs/en/user/05-distributed-guide.md
georgebisbas: Already resolved — the current `distributed/03-execution.md` and `04-debugging.md` correctly say a missing `.share_memory_()` raises a `TypeError` at dispatch time, not silent stale output. The wrong "passes silently" claim only existed in the now-deleted `05-distributed-guide.md`.

docs/en/user/05-distributed-guide.md
georgebisbas: Fixed in 60c1b490 - same fix as the `04-distributed-api-reference.md` thread; the non-functional `cmake -D...` example is removed from `distributed/04-debugging.md`.

docs/en/user/05-distributed-guide.md
georgebisbas: Resolved by the restructure commit (f3705b32) — the content is now split across `distributed/00-model.md` through `04-debugging.md`, each well under the 500-line cap (largest is 184 lines).

docs/en/user/06-benchmarking-guide.md
georgebisbas: Fixed in 60c1b490 - `performance/02-distributed.md`'s Ring Sizing section now shows `RunConfig` passed to both `prepare()` and every dispatch call (per `distributed_compiled_program.py`'s docstring, `prepare(config=...)` only prewarms the runtime-arena cache — it isn't stored on the worker), and states the real power-of-2 constraints (`ring_task_window >= 4`, `ring_heap >= 1024` bytes) per `runner.py`.

docs/en/user/06-benchmarking-guide.md
georgebisbas: Fixed in 60c1b490 - dropped the `PYPTO_BENCH*` table from `performance/00-methodology.md`'s Quick Start section, replaced with a pointer to `pypto-lib`'s own docs; kept `pypto.runtime.benchmark()` as this repo's documented benchmarking contract.

docs/en/user/02-operation_reference.md
georgebisbas: Fixed in 60c1b490 - removed the branch reference from the AllReduce capability-matrix row in `02-operation_reference.md`; softened the dtype claim since `LowerCompositeOps` doesn't enforce a dtype allowlist (only storage-width/alignment checks at `lower_composite_ops_pass.cpp:755-760`).

python/pypto/language/distributed/op/tensor_ops.py
georgebisbas: Fixed in 60c1b490 - applied the corrected signal-shape wording at all 6 sites in `tensor_ops.py` (`allreduce`, `barrier`, `broadcast`, `allgather`, `reduce_scatter`, `all_to_all` + its `Args: signal` entry). Note: only `allreduce` actually has a `mode` parameter, so the other 5 sites state the mesh-only rank-2 `[nranks, 1]` shape without the ring formula, which doesn't apply to them.

python/pypto/runtime/bench.py
georgebisbas: Fixed in 60c1b490 - widened the `BenchmarkStats` quick-reference table's first column marker from 26 to 30 characters. Verified with docutils that the table now parses as a single `table` node with no errors.

georgebisbas:
## Update: Rebased onto main, wired into the site, and addressed all review comments

This PR is now rebased onto current `main` and all outstanding review feedback has been resolved (commit `60c1b490`).

### Rebase onto `main`

The branch had drifted ~2 days behind `main`, which in the meantime gained the MkDocs site scaffold (#2193), including the `docs/zh-cn` → `docs/zh` locale rename. Rebased and moved this PR's zh content to the new `docs/zh/` path accordingly.

### Wired into the live site

The `distributed/` and `performance/` chapters didn't exist in `mkdocs.yml`'s nav or the top-level user-manual `index.md` when this branch forked. Added:
- Both chapters to `mkdocs.yml`'s `User Manual` nav (with zh nav translations).
- Links to both chapters from `docs/en/user/index.md` / `docs/zh/user/index.md`, replacing the old "not here yet" placeholder rows.
- Retargeted 5 dangling cross-references (in `00-getting_started.md` / `02-operation_reference.md`, en + zh) that still pointed at the deleted flat `04-/05-/06-*.md` files.
- Fixed a handful of broken relative links (wrong path depth, links to not-yet-existing `language/`/`tutorials/`/`execution/` chapters) that `mkdocs build --strict` caught.

### CI fixes

The restructure commit introduced 66 markdownlint errors (MD060 table-pipe spacing, MD028 blank lines inside blockquotes, MD040 missing fence languages) that were never fixed. All resolved — `pre-commit run --all-files` and `mkdocs build --strict` are both clean.

### Review comments — all 18 addressed

13 needed real fixes, all applied and replied to inline; 5 turned out to already be resolved by the restructure commit:

- **`distributed/00-model.md`** — the Quickstart `HelloAllReduce` example had 4 real bugs (missing `Orchestration` dispatch layer, a rank mismatch from `pl.slice`, a redundant post-dispatch `pl.store`, wrong window-buffer shape). Rewritten to match the working `tests/st/distributed/collectives/test_l3_allreduce.py::AllReduceMesh` reference.
- **`distributed/02-primitives.md`** — `window`/`alloc_window_buffer` moved out of the `pld.system.*` table into their actual `pld.tensor.*` namespace.
- **`distributed/01-collectives.md`** + **`tensor_ops.py`** (6 docstrings) — corrected the host-builtin signal shape (rank-1 *or* rank-2, not only 1-D) and replaced the InCore-hand-rolled-vs-HOST-builtin dichotomy with the real three-way split (hand-rolled / InCore composite / HOST builtin), since `ring` mode is InCore-composite-only.
- **`distributed/03-execution.md`**, **`performance/00-methodology.md`** — dropped fabricated `PYPTO_BENCH*` defaults that belong to `pypto-lib`, not this repo.
- **`distributed/04-debugging.md`** — removed a non-functional `cmake -D...` example for compile-time macros that no `CMakeLists.txt` in the repo reads.
- **`performance/02-distributed.md`** — corrected Ring Sizing to show `RunConfig` passed to every dispatch (not just `prepare()`), with the real power-of-2 size constraints.
- **`02-operation_reference.md`** — dropped a stale branch reference and an unverified dtype restriction from the AllReduce capability matrix.
- **`bench.py`** — fixed a malformed RST table (column marker narrower than its widest cell).

All fixes mirrored in `docs/zh/user/`. CI is green (13/13 checks).

georgebisbas:
converting to draft, no need for review yet, needs some work and polishing

georgebisbas:
This PR will be broken down to smaller parts as discussed offline with @lyfne123. Making it draft, and will step by step break it down and to smaller ones.

docs/en/user/distributed/03-execution.md
```diff
@@ -0,0 +1,156 @@
...
+
+## StackedDeviceTensor
+
+Sharded across devices — obtained via `rt.alloc_stacked_tensor()`:
+
+```python
+# Host tensor sharded along dim 0 — shard[i] lives on card i.
+host_weights = torch.randn(4, 1024, 4096)  # 4 shards
+with compiled.prepare() as rt:
+    stacked = rt.alloc_stacked_tensor(host_weights)
```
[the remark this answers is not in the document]
georgebisbas: Fixed in 0cebf07c — added `.share_memory_()` to `host_weights` plus a fatal-pitfall callout explaining why: the upload runs inside the already-forked chip worker, which can only read memory it inherited at fork.

docs/en/user/distributed/03-execution.md
```diff
@@ -0,0 +1,156 @@
...
+
+A single `DistributedWorker` can dispatch multiple compiled programs:
+
+```python
+compiled_a = ir.compile(ProgramA, platform="a2a3", distributed_config=dc)
+compiled_b = ir.compile(ProgramB, platform="a2a3", distributed_config=dc)
+
+with compiled_a.prepare() as rt:
+    rt(host_x, host_out)            # dispatch ProgramA
+    rt.run(compiled_b, host_x, host_out)  # dispatch ProgramB
```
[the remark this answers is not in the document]
georgebisbas: Fixed in 0cebf07c — added `extra_compiled=[compiled_b]` to `prepare()` so `compiled_b` is actually registered, switched the primary dispatch to `rt.run(compiled_a, ...)` too (multi-program mode disables the `rt(*args)` shortcut entirely, raising `TypeError`), and fixed the same pattern in the performance page's mirror of this example (both EN and ZH).

docs/zh/user/00-getting_started.md
```diff
@@ -41,7 +41,7 @@ def vector_add(
 
 ```python
 print(vector_add.as_python())
-```
+```python
```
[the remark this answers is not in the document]
georgebisbas: This is stale — the diff hunk here predates main's #2198 doc restructure. docs/zh/user/00-getting_started.md no longer differs from main at all in this PR (the vector_add.as_python() content moved out entirely), so there's no fence to fix here anymore. Resolving.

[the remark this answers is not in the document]
georgebisbas: Correction to my last reply: I was wrong. I had mis-diagnosed this as stale due to a local branch mix-up on my end (briefly compared against the wrong checkout). Re-verified against the actual PR branch: this bug was real, and worse than reported — every closing ``` fence in this file (8 of them, not just this one) had been rewritten to ```python. Fixed for real in c7157024.

georgebisbas:
Closing in favor of the split: #2218, #2219, #2220, #2221. See the updated description above for the mapping.

Issue #2131, opened 2026-07-24
[Feature] Explicit buffer-slot binding for fine-grained multi-level double buffering


lyfne123:
## Background

In pypto-lib, users can create logical tiles in Mat/Left/Right/Acc, but cannot
explicitly bind those tiles to named physical buffer slots. There is no way to
write the CCE-style pattern:

```text
slot = iteration & 1
GM -> l1_buffer[slot]        # 512-token cadence
L1 -> l0b_buffer[sub & 1]    # 128-token cadence
MAD -> l0c_buffer[sub & 1]
TSTORE(l0c_buffer[sub & 1])
```

`pl.pipeline` and the allocator own buffer identity, count, reuse, and lifetime.
Consequently a nested `stage=2` pipeline happens to allocate two 512-wide L1
buffers and two 128-wide L0B buffers, but only one Acc/L0C buffer. Generated
code drains each result with TSTORE before the next TMATMUL can reuse L0C:

```text
TEXTRACT(B0), TEXTRACT(B1),
TMATMUL(C0, B0), TSTORE(C0),
TMATMUL(C0, B1), TSTORE(C0)
```

Reproduce with:

```bash
PYPTO_BENCH=1 python models/qwen3/14b/repro_two_level_l0c.py -p a2a3 -d 3
```

The repro case is not committed. Save the inlined script below as
`models/qwen3/14b/repro_two_level_l0c.py` in pypto-lib.
The only repo-local dependency is the public pypto-lib `golden` package at the
commit recorded below.

Reproduction environment:

| Component | Version |
|---|---|
| pypto-lib | `72ee6a1e2e34ac2f9cad9e89fb6a70f6ee5302eb` (branch: `main`) |
| pypto | `c9b2674c8fe32c738b97d8130a2b47d09daa000f` (branch: `main`) |
| simpler | `8cdb306cb9a81ad1a0561325021105c676a69c1e` (detached, matches pin) |
| ptoas | `0.48` (matches pin) |
| pto-isa | `83d01313d9bfc247c4b7c8bcf969d1019f0d106f` (detached, matches pin) |
| CANN | `9.0.0` (`npu-smi 26.0.rc1`) |

Diagnosis: **pypto** — the frontend/IR exposes memory spaces but not explicit
buffer identity. `pl.create_tile(..., target_memory=...)` and
`pl.tile.extract(..., target_memory=...)` create logical SSA tiles; users cannot
[...]

PR #2210, opened 2026-07-30
feat(ir): Give a declared MemRef N slots, selected by subscript

lyfne123:
## What

`pl.MemRef("l0c")` declares one allocation. Double-buffering wants two that share it — [issue 2131](https://github.com/hw-native-sys/pypto/issues/2131)'s shape, where an Acc buffer alternates between the accumulating half and the half being drained. Writing that today means two declarations, which the compiler is free to place apart.

```python
l0c = pl.MemRef(slots=2)

for i in pl.range(N):
    # slot alternates per iteration; both slots live in one allocation
    c: pl.Tile[[M, N], pl.FP32, l0c[i % 2], pl.Mem.Acc] = pl.tile.matmul(a_t, b_t)
    ...
```

`slots=N` reserves N equally-sized slots in **one** allocation; a subscript picks one. The index is an ordinary index expression, so it may be a runtime value.

Nothing about this is tied to `pl.pipeline` — the compiler only sees an index expression in the annotation, so a plain `pl.range`, a `pl.pipeline` sub-index, a function parameter, or a constant all work the same. A bare `pl.range(4)` with `l0c[i % 2]` lowers to one `pl.tile.alloc(pl.Mem.Vec, 32768, pinned=True)` and a per-iteration offset of `i % 2 * 16384`.

## Why a runtime index is safe

PTOAS takes `alloc_tile addr` **by value** — it lowers to `TASSIGN(tile, uint64_t)` — so a per-iteration address is a legal target. Verified against ptoas 0.54 with `(i % 2) * 8192` inside an `scf.for`; the emitted C++ is a per-iteration `TASSIGN`.

## How

| Layer | Change |
| ----- | ------ |
| `MemRef` | New `slot_count_` (uint64) and `slot_index_` (`optional<ExprPtr>`); both meaningful only while `is_pinned_` |
| `InitMemRef` | Sizes a slot to the largest tile bound to **any** slot, the allocation to `slot_count * slot`, and turns the index into `byte_offset_ = index * slot` |
| `AllocateMemoryAddr` | Adds the base address to a dynamic offset instead of dropping it; the address check skips high-water tracking for a non-constant offset |
| PTO codegen | Emits the widened offset expression as the tile's address |
| `MemoryReuse` | Co-liveness is now per **slot**, not per allocation |
| Parser / printer | `pl.MemRef(slots=N)` and `decl[k]`, round-tripping through the Python printer |

A constant index folds to a `ConstInt` in `InitMemRef`, so constant slots take the ordinary static-address path completely unchanged.

## The one invariant worth calling out

Because the slot index may name SSA values, it is the **only** MemRef field that substitution has to rewrite. `CloneTypeWithMemRefAndRemapExprs` does so **only for a pinned MemRef** — which confines every MemRef rebuild to before `InitMemRef`, strictly earlier than every pass that keys on MemRef *pointer* identity (`AllocateMemoryAddr` matches old→new by raw pointer). Rebuilding one of those later would silently split a shared MemRef.

## Tests

`tests/ut/ir/transforms/test_declared_memref.py` — 38 pass. New `TestSlots` (6) and `TestSlotRejects` (3) cover:

- Two constant slots → **one** allocation, offsets `0` / `16384`
- A runtime index → `pl.tile.alloc(pl.Mem.Acc, 16384, pinned=True)` with the address computed per iteration
- Different slots co-live is accepted; the same slot co-live is still rejected
- Round-trip of both `l0c[0]` and `l0c[i % 2]` via `assert_structural_equal`
- Out-of-range index, `slots=0`, and subscripting a non-declaration are rejected

Full `tests/ut/`: **8184 passed, 1 failed** — the failure is `tests/ut/runtime/test_benchmark.py::test_benchmark_l3_ignores_prepare_setup_groups`, pre-existing on `main` (a test double's `prepare()` signature doesn't match `bench.py`'s call; neither file is touched here).

## Docs

`docs/{en,zh}/dev/language/00-python_syntax.md` — new Slots subsection.

## Note

`i & 1` is not currently writable in the DSL (no `BitAnd`); `i % 2` is the idiom for now.


lyfne123:
Addressed the review in `7ae95acb`.

**Fixed**

| Finding | Fix |
| ------- | --- |
| P1 — slot MemRef carried its slot's offset but the whole allocation's size | Each slot MemRef is now sized to one slot; the allocation keeps the full size via a new `alloc_size` override on `CreateAllocStatement`. Confirmed real: slot 1 of two spanned `[S, 3S)`, so the address verifier demanded `3S` for something that fits in `2S`, and `MayAlias` reported the ping-pong's two halves as aliasing — the exact pair the feature exists to keep apart. |
| P1 — runtime slot index printed without the parent's rename maps | Sub-expressions nested in a type annotation now print through a new `PrintSubExpr` that inherits the naming state. Also covers the MemRef byte offset, which had the same latent gap for dynamic (tile.slice) offsets. |
| P2 — `pl.MemRef` docstring still said runtime indices were rejected and co-liveness was per allocation | Rewritten. |
| `slots >= 1` unchecked in the raw declaration constructors | `CheckSlotCount` in both `__init__` overloads + a test. |
| clang-tidy: missing `<optional>` in `src/ir/memref.cpp` | Added. |

The printer finding is worth calling out because it was **silent**: two sibling loops both written `i` print as `i` and `i_1`, but the slot index emitted the bare `name_hint_`, so the dump reparsed with the address bound to the *other* loop's variable and read a different slot. Verified the new test fails without the fix:

```
t: pl.Tile[..., pl.MemRef("l0c", slots=2)[i % 2], ...]   # inside `for i_1 in pl.range(2)`
assert 'l0c", slots=2)[i_1 % 2]' in ...  -> AssertionError
```

**Not changed**

- `github-code-quality` "Statement has no effect" on `python/pypto/language/typing/memref.py:115,117` — those are `@overload` stubs with `...` bodies. The bot flagged only the two lines this PR touched; the six adjacent pre-existing overloads use the identical idiom. False positive.

**System tests** — `tests/st/runtime/ops/test_memref_slots.py`, one case per memory class (`pl.Mem.Acc` / `pl.Mem.Vec`), since the address reaches a different ISA operand in each. Both hold the two slots **co-live** (each iteration writes `[i % 2]` and `[(i + 1) % 2]` and only then consumes them), so a collapsed address corrupts the golden rather than passing quietly. Both compile; the emitted address is per-iteration as intended:

```mlir
%0 = arith.remsi %i__idx_v0, %c2_index : index
%1 = arith.muli  %0, %c16384_index : index
%2 = arith.addi  %c16384_index, %1 : index
%3 = arith.index_cast %2 : index to i64
%lo__ssa_v0 = pto.alloc_tile addr = %3 ...
```

I could not execute them on hardware locally: every on-device ST test in my environment — including untouched ones like `tests/st/runtime/ops/test_abs.py` — fails at `worker.init` with `TypeError: init(): incompatible function arguments ... CallConfig, bool`, an installed `_task_interface` / simpler-source mismatch. CI's `system-tests` job is the real check.

**Unit tests** — `1 failed, 8186 passed`. The failure is the pre-existing `tests/ut/runtime/test_benchmark.py::test_benchmark_l3_ignores_prepare_setup_groups` (a test double's `prepare()` signature does not match `bench.py`'s call; neither file is touched here).

lyfne123:
CI is green — all 15 checks pass.

`97dd1aab` fixes a regression the earlier commits introduced, which only surfaced once `clang-tidy` stopped gating the heavy jobs.

**What broke.** Emitting `base_addr + offset` for *every* non-constant byte offset broke programs with a dynamic-offset `tile.slice` view — that expression is not renderable at the `pto.alloc_tile addr` position, so PTOAS rejected the module:

```
qk_pv.pto:127:19: error: expected SSA operand
merge_norm.pto:119:19: error: expected SSA operand
```

My comment in that commit claimed such a view was "unaffected either way". That was wrong. A slice view *does* re-derive its offset from its own operands via `pto.subview`, so this address is unused — but it is still **emitted**, and an unrenderable expression there is fatal rather than ignored.

**The fix.** A dynamic address is only meaningful for an author-declared allocation, where it carries the runtime slot index. Everything else keeps the original behaviour of falling back to the bare base. `InitMemRef` deliberately drops `is_pinned_`/`slot_index_` from the MemRefs it resolves, so the surviving record of which allocations were declared is the `pinned=True` kwarg on their alloc statements; `PinnedAllocBaseCollector` gathers those in one O(N) walk.

**Verified both directions** on the model CI failed on (`decode_attention_swa`), locally:

| | Result |
| --- | --- |
| With the guard | compiles and **matches golden** (`kv_cache` PASS, `x_out` PASS) |
| Guard removed | reproduces CI's error exactly — same file, same line, same message |

**On-device.** `system-tests-direct` ran both new system tests on hardware:

```
tests/st/runtime/ops/test_memref_slots.py::TestMemRefSlots::test_acc_slots_rotate_at_runtime PASSED
tests/st/runtime/ops/test_memref_slots.py::TestMemRefSlots::test_vec_slots_rotate_at_runtime PASSED
```

That is the validation I could not get locally (my environment fails every on-device ST test at `worker.init` on an installed-`_task_interface` mismatch), so the co-live ping-pong is now confirmed correct on real silicon in both memory classes.

lyfne123:
Resolved the six addressed threads. `ffa1eb59` fixes the one that was still open — CodeRabbit's 🔴 Critical, which was real.

**Sentinel collision (`init_memref.cpp`)** — `DeclaredAlloc::slot_count` defaulted to 1 and `Record` used that same 1 as its "not yet recorded" sentinel. An ordinary unsubscripted declaration carries a count of 1, so recording one never consumed the sentinel; a later, genuinely different count on the same name overwrote it instead of tripping the mismatch check. Fixed by using 0, which `Record`'s own `slot_count_ >= 1` guard makes unreachable for real data.

Worth recording why the obvious repro does *not* show it — I nearly filed this as already-caught. The collector visits a Var at **every occurrence**, so a 1-slot tile still live across the 2-slot declaration gets re-checked and trips the mismatch regardless of the sentinel. Both of these error either way:

```python
t0: pl.Tile[..., pl.MemRef("buf"), ...] = pl.load(a, [0, 0], [64, 64])
t1: pl.Tile[..., pl.MemRef("buf", slots=2)[1], ...] = pl.exp(t0)   # t0 still live -> caught
```

The hole needs each tile fully consumed before the next declaration appears, so nothing re-checks the first:

```python
t0: pl.Tile[..., pl.MemRef("buf"), ...] = pl.load(a, [0, 0], [64, 64])
r0 = pl.store(t0, [0, 0], o1)                                       # t0 dead here
t1: pl.Tile[..., pl.MemRef("buf", slots=2)[1], ...] = pl.load(a, [0, 0], [64, 64])
r1 = pl.store(t1, [0, 0], o2)
```

Old sentinel: accepted silently, `buf: pl.Ptr = pl.tile.alloc(pl.Mem.Vec, 32768, pinned=True)` — sized for 2 slots from a declaration the author contradicted. New sentinel: rejected. That is the shape the regression test uses, and it fails with the old sentinel.

**The other new thread** (`python_printer.cpp:2866`, 🟠 Major, `temp_printer` dropping the rename map) is the same finding as Codex's P1 and was already fixed in `7ae95acb` — the suggested diff patches a `temp_printer` block that no longer exists; it now goes through `PrintSubExpr`, which inherits `var_rename_map_` / `dyn_var_rename_map_` / `free_body_vars_`. Resolved as already-addressed.

**Left open deliberately:** the two `github-code-quality` "Statement has no effect" comments on `python/pypto/language/typing/memref.py:126,128`. Those are `@overload` stubs with `...` bodies — the bot flagged only the two lines this PR touched, while the six adjacent pre-existing overloads use the identical idiom. False positive; say the word if you'd rather I resolve them too.

Unit tests: `1 failed, 8187 passed` — the failure is the pre-existing `test_benchmark_l3_ignores_prepare_setup_groups`, untouched here.

lyfne123:
Both points are correct. Fixed in `20b1de2a`.

**The overlap is real, and it is corruption rather than wasted space.** Reproduced exactly:

```
aaa: pl.Ptr = pl.tile.alloc(pl.Mem.Vec, 32768, pinned=True)
m0  base=aaa        [0, 16384)
m1  base=aaa        [16384, 32768)
x   base=mem_vec_3  [16384, 32768)   <- co-live with m1, different buffer
s1  base=mem_vec_3  [16384, 32768)
```

Root cause is as diagnosed. `7ae95acb` made each slot MemRef report its own slot — needed so `MayAlias` does not call the two halves of a ping-pong aliased, and so the address verifier does not bounds-check `[S, 3S)` — but `AllocateMemoryAddresses` sizes a base group's buffer from `max(member.size_)`. It had been relying on the invariant its own comment states, "the root MemRef (byte_offset == 0) is sized to the full alloc", which per-slot sizing broke.

A declared allocation is the one base whose buffer is deliberately larger than any single MemRef in it, so the size cannot be inferred from the members. It now comes from the `tile.alloc` statement, which `InitMemRef` already sized to `slots x slot_size` — `PinnedAllocCollector` carries it alongside the base it was already collecting for the dynamic-address check.

Worth noting **why the existing tests missed it**: addresses are handed out in `name_hint_` order, and `multi`/`l0c`/`ub` all sort *after* the compiler's `mem_vec_*` bases, so the short reservation was always last and nothing followed it to collide. The regression test names the declaration `aaa` so it sorts first, then asserts no two ranges on different bases overlap. It fails without the fix with exactly the reported overlap.

**Interface sync** — also correct. The binding's int-offset constructor grew `is_pinned` / `slots` / `slot`; `ir.pyi`'s overload had not. Synced. (The `ExprPtr`-offset overload genuinely does not take them — `nb::init<VarPtr, ExprPtr, uint64_t, Span>` — so the stub now matches the binding rather than the C++ constructor.)

Verification, all after the fix:

| Check | Result |
| --- | --- |
| `tests/ut/` | 1 failed, 8188 passed — the failure is the pre-existing `test_benchmark_l3_ignores_prepare_setup_groups` |
| Slots system tests (codegen) | 2 passed |
| pypto-lib `decode_attention_swa` | PASS against golden |
| Regression test without the fix | fails with `m1 [16384, 32768) on 'aaa' overlaps x [16384, 32768) on 'mem_vec_3'` |

Thanks — this one would have been very unpleasant to hit in the field.

lyfne123:
Re-reviewed the latest head (`20b1de2a`). The previously reported allocation-overlap bug and the `ir.pyi` constructor mismatch are both fixed, and the new overlap regression test covers the original failure well. I found two remaining blockers:

### 1. Embedded MemRef serialization drops the slot metadata

`src/ir/serialization/serializer.cpp::SerializeMemRef` still writes only `base`, `byte_offset`, and `size`. The updated deserializer can read `is_pinned`, `slot_count`, and `slot_index`, but those fields are never emitted for a MemRef embedded in a `TileType`/`TensorType`.

Minimal result from the current head:

```text
before serialize: is_pinned=True, slot_count=2, slot_index=1
after deserialize: is_pinned=False, slot_count=1, slot_index=None
```

So `ir.serialize`/`ir.deserialize` silently turns `pl.MemRef("buf", slots=2)[1]` into an ordinary one-slot MemRef and loses the selected slot. Please serialize all three fields and add a round-trip test for a MemRef embedded in a shaped type. The existing standalone-MemRef serialization tests use a different serialization path and therefore do not catch this.

### 2. The final MemRef sort is no longer a strict weak ordering

`AllocateMemoryAddresses` now legitimately leaves runtime-slot addresses as expressions, so `memref_pairs` may contain both constant and dynamic offsets. The comparator at `allocate_memory_addr_pass.cpp:341` compares by offset when *both* values are constant, but falls back to name when either is dynamic.

That relation is non-transitive. For example:

```text
A = (name="z", offset=0)
B = (name="a", offset=1)
C = (name="m", offset=dynamic)

A < B  # offset comparison
B < C  # name comparison
A < C is false  # name comparison
```

Passing such a comparator to `std::sort` is undefined behavior. Please use one consistent total key—for example, constant/dynamic category first, then offset/name with a stable tie-breaker—or remove the sort if its ordering is not used by the subsequent pointer map.

Verification on the latest head:

- worktree-local build: passed
- `test_declared_memref.py`: 42 passed
- `test_memref.py`: 110 passed
- GitHub CI: all checks passed

The serialization repro above still fails semantically despite the green test suite.

lyfne123:
Both confirmed and fixed in `62e860e6`.

### 1. Embedded MemRef serialization

Reproduced, and it is worse than losing the slot — `is_pinned_` goes with it, so the declaration comes back as an ordinary compiler allocation that `InitMemRef` no longer treats as declared:

```
before: pl.MemRef("buf", slots=2)[1]
after:  pl.MemRef("buf", pl.const(0, pl.INT64), 0)
```

The reader (`deserializer.cpp::DeserializeMemRef`) already accepted all three keys — only the writer was missed, so this was a one-sided gap. Fixed in `SerializeMemRef`. The fields are written **only when they carry something**, and the reader's defaults are exactly the omitted values, so an ordinary allocation still serializes byte-for-byte as before and old blobs still load.

Regression test added: a slot MemRef embedded in a `TileType`, asserting `assert_structural_equal` plus `is_pinned_` / `slot_count_` / `slot_index_` explicitly. Your point about the standalone tests using a different path is exactly right — they pass either way; the new test fails without the fix on the `is_pinned_` assertion.

### 2. Non-transitive comparator

Also correct, and it is my change that made it reachable. The comparator was fine while every post-pass address was constant — `off_a && off_b` was effectively always true and the name fallback was dead code. A declared allocation's runtime slot index made the mixed case real, and your A/B/C trace is exactly the violation.

Now ordered on one total key: constants first in address order, then dynamic addresses by name, with `name_hint_` breaking offset ties (two MemRefs in one base group can legitimately share an offset — a view over its parent). I kept the sort rather than removing it, since the address-order intent is still meaningful for the constant addresses that make up almost all of them.

### Verification

| Check | Result |
| --- | --- |
| `tests/ut/` | 1 failed, 8189 passed — the failure is the pre-existing `test_benchmark_l3_ignores_prepare_setup_groups` |
| `test_declared_memref.py` | 43 passed |
| serialization + memref tests (`-k "serial or memref"`) | 294 passed |
| Slots system tests (codegen) | 2 passed |
| Serialization test without the fix | fails: `the round trip turned a declaration into a compiler allocation` |

Good catch on both — the serialization one in particular was invisible to the whole green suite.

lyfne123:
Correct — fixed in `db3e669b`. Reproduced exactly:

```
before: buf: pl.Ptr = pl.tile.alloc(pl.Mem.Vec, 32768, pinned=True)
after:  buf:   pl.Ptr = pl.tile.alloc(pl.Mem.Vec, 32768, pinned=True)
        buf_1: pl.Ptr = pl.tile.alloc(pl.Mem.Vec, 32768, pinned=True)
```

Allocation identity in the IR is base_ **pointer** identity — it is what `MemRef::SameAllocation` compares, what keys `InitMemRef`'s `DeclaredAllocMap`, and what groups the address allocator's buffers — while the wire format carries the base only by name. So minting a fresh `Var` per embedded MemRef split one allocation into as many as there were MemRefs naming it, and a declaration lost the one property that makes slots slots.

I fixed it on the **reader** rather than by serializing the base as a node reference: the reader now interns one base Ptr per name. A name identifies an allocation — the parser enforces that for declarations, and compiler allocations are named uniquely — so interning restores exactly the sharing the writer meant to express, without changing the wire format or invalidating existing blobs. It also repairs `SameAllocation` for ordinary `tile.slice` views across a round trip, which had the same latent split.

Test strengthened as you asked: the previous one used a single slot, which round trips fine either way. It now carries slot 0 **and** slot 1, asserts `memrefs["t0"].base_ is memrefs["t1"].base_`, and checks `InitMemRef` still emits exactly one allocation. Without the fix it fails on `the round trip split the shared base`.

| Check | Result |
| --- | --- |
| `tests/ut/` | 1 failed, 8189 passed — the failure is the pre-existing `test_benchmark_l3_ignores_prepare_setup_groups` |
| `test_declared_memref.py` | 43 passed |
| pre-commit | clean |

That is three real defects from this feature that the green suite could not see (serialization fields, shared base, the reservation overlap) — all of the same shape: something outside the passes was relying on an invariant the slot work changed. Appreciate the persistence on these.

lyfne123:
Reproduced all seven before touching anything. **Six were real and are fixed in `3b5f93ac`; the seventh is real but its cause is pre-existing and repo-wide**, so I did not change it unilaterally — details at the bottom.

### P1 — base not bound to the real alloc Var

Confirmed, with exactly your numbers:

```
m1 on "aaa"        [16384, 32768)
x  on "mem_vec_3"  [16384, 32768)   <- co-live, different buffer
```

You were right that name interning only made the MemRefs share a base *among themselves*. That Var is not the alloc statement's, so `PinnedAllocCollector` (keyed on the alloc's LHS) never matched the group and it was sized as if it held one MemRef. A second symptom confirmed it: the alloc lines vanished from the dump entirely, because the printer no longer saw the base as body-defined.

Fixed the way you originally suggested — the base is serialized as a **node**, so the node graph hands back the exact same object every other reference resolves to. The name is still written and the interning fallback stays for blobs that predate `base_node`.

### P1 — unbound slots escaped the capacity check

Confirmed: `slots=12` produced a 196608-byte Vec allocation and verified clean against the 188416-byte space. The verifier walks bound tiles, and each slot MemRef spans only its own slot, so eleven unbound slots were invisible. It now folds in what each alloc statement reserves — exact when a tile is bound to the first slot (`l0c[0]` / `l0c[i % 2]`, i.e. every realistic case), conservative otherwise, so it can over-report but never under-report.

### P2 — substitution

Both halves confirmed. `IRMutator::VisitExpr_(MemRefPtr)` rebuilt from base/offset/size only, so `is_pinned=True, slots=2, slot=1` came back `False, 1, None`; and the slot index was never followed, so a substituted index Var kept pointing at the old one — an index-only change did not even register as a change. Both fixed, tested through `ir.substitute_expr` directly.

### P2 — SSA/UseAfterDef collector

Confirmed: `VisitTypeExprFields` walked shape and views but never the MemRef. Now walks the slot index.

### P2 — TotalSize overflow

Confirmed: `slots=2**60` gave `pl.tile.alloc(pl.Mem.Vec, 0, pinned=True)`. Now rejected — wrapping is worse than a large number, since it produces a *small* allocation and then hands out addresses inside it.

### P2 — constructor compatibility

Confirmed against `e77ccedb`: the declaration ctors were `(span)` and `(name, span)`, and I had inserted `slots` ahead of `span`. `slots` now follows `span`; both positional forms work again, `slots=` is unaffected, and the stub + DSL overloads match the binding.

### P2 — interning scope: could not reproduce harm

Two functions each declaring `"buf"` round-trip to the same two allocations with their own sizes (16384 and 4096), and ROUNDTRIP verification passes — allocs are emitted per function body and the passes are function-scoped, so the shared Var is inert. With `base_node` landed the interning is now a legacy-blob path only. I left it rather than adding per-function scoping I could not show was needed.

### P2 — StructuralEqual: real, but not slot-specific

Your repro is right. The cause is broader than slots: **`EqualType` does not compare `memref_` at all**. On `main`, two programs binding tiles to *different named allocations* already compare equal:

```
different allocation NAMES: EQUAL
```

So this is long-standing behaviour, not something the slot fields introduced, and making memory part of type equality would change comparison semantics for every transform test in the repo. I extended `EqualMemRef` (the node path, where it is free and correct) and left the type path alone. **If you want type equality to include memrefs, that is worth its own PR** — happy to open one; it is not a change I want to slip in here. It does explain the explicit field assertions in the round-trip tests, which are the right mitigation meanwhile.

### Verification

| Check | Result |
| --- | --- |
| `tests/ut/` | 1 failed, 8193 passed — the failure is the pre-existing `test_benchmark_l3_ignores_prepare_setup_groups` |
| `test_declared_memref.py` | 47 passed (5 new) |
| Slots system tests (codegen) | 2 passed |
| pypto-lib `decode_attention_swa` | PASS against golden |
| pre-commit | clean |

Each new test was negative-controlled — reverting its fix makes it fail with the reported symptom (`substitution demoted a declaration to a compiler allocation`, `DID NOT RAISE`, the overlap, and so on).

lyfne123:
Three fixed in `a899906a`; the fourth I measured rather than argued — details below.

### P1 + P2 — both capacity bugs, same root cause

Reproduced both exactly:

```
dynamic  (slots=12, l0c[i % 12]): ACCEPTED -> alloc(pl.Mem.Vec, 196608)   # over 188416
high-only(slots=8,  only slot 7): REJECTED                                # 131072 fits
```

You are right on both, and they are the same mistake: I was **reconstructing** the footprint from the addressed tiles. That cannot work in either direction — a slot nobody binds leaves no MemRef to see, and a dynamic address leaves no constant to sum. Deriving the buffer's base from the lowest bound tile was the specific unsound step behind the false rejection.

So the check moved to where the number is already exact: `AllocateMemoryAddresses` builds `SpaceFootprint` from each buffer's true reserved size. The address verifier goes back to its member-based high-water — it can under-report, but it can never reject a legal program, which is the right failure direction for a redundant net.

```
dynamic:   REJECTED -> Vec buffer usage (196608 bytes) exceeds platform limit (188416 bytes)
high-only: ACCEPTED -> alloc(pl.Mem.Vec, 131072)
```

Writing the runtime-index test surfaced **a further defect neither of us had listed**: a resolved runtime slot index becomes part of the byte offset (`i % 12 * 16384`), and the parser accepted only add/sub/mul there — so such a program printed but did not reparse. Fixed by falling back to the same index-expression parser the slot subscript and TileView fields already use.

### P2 — docs

Correct, and worse than stale: they stated the rule this PR inverts.

- `29-init_memref.md:104` "tiles bound to one allocation must not be live at the same time"
- `31-memory_reuse.md:120` "Two tiles independently bound to one allocation must not be live at the same time"

Both now say per **slot**, note that different slots are *meant* to be co-live, and cover the runtime-index case. Chinese counterparts updated with them (`docs/zh/.../29`, `.../31`).

### P2 — structural equality: measured, not argued

You are right that it is the most-used comparison path, so I stopped asserting and tested it: I made `EqualType` compare `memref_` and ran the suite.

**433 failed, 7749 passed.**

That is not slot fallout — it is what comparing programs across the memory passes depends on, and it is why `EqualType` ignores `memref_` outright today (on `main`, two programs bound to different *named allocations* already compare equal). So this is a deliberate repo-wide semantic, not something the slot fields broke, and flipping it is a decision with a 433-test blast radius.

I have left it alone and kept the `EqualMemRef` (node-path) extension. **Happy to open a separate PR** that makes memory part of type equality and works through those 433 — but it should be its own change with its own review, not a rider here.

### Verification

| Check | Result |
| --- | --- |
| `tests/ut/` | 1 failed, 8195 passed — the failure is the pre-existing `test_benchmark_l3_ignores_prepare_setup_groups` |
| `test_declared_memref.py` | 49 passed (3 new capacity cases: over-capacity constant, over-capacity dynamic, high-slot-only accepted) |
| Slots system tests (codegen) | 2 passed |
| pre-commit | clean |

Hzfengsy (approved):

Issue #1475, opened 2026-05-22
[Bug] MemoryReuse over-aggressively reuses tile buffers, serializing independent matmuls (fa_qks cube/vector pipeline stall)


lwDavid:
### Diagnosis

**pypto** — The `MemoryReuse` pass (run between `28_after_InitMemRef` and `29_after_MemoryReuse`) reuses tile buffers as soon as a value's liveness ends. The reuse is functionally correct, but it introduces a WAR (write-after-read) hazard across pipe stages that serializes otherwise-independent compute, collapsing pipeline overlap.

### Description

Observed in the Qwen3-14B fused decode attention kernel (`models/qwen3/14b/decode_layer.py` in pypto-lib), function `fa_qks`.

**Symptom (trace):** On the cube (AIC) core, the two QK matmuls inside the `fa_qks` loop body run serially instead of overlapping — the second matmul's K-load (MTE2) stalls waiting on the first matmul. The vector (AIV) core shows the same pattern.

**Root cause, traced through the IR dumps:**

In `28_after_InitMemRef.py` (`fa_qks_aic`), the two matmuls use *distinct* Mat buffers:

```
mem_mat_12 = alloc(Mem.Mat, 32768)   # k_tile0, matmul 0
mem_mat_13 = alloc(Mem.Mat, 4096)    # q_padded0, matmul 0
mem_mat_17 = alloc(Mem.Mat, 32768)   # k_tile1, matmul 1
mem_mat_18 = alloc(Mem.Mat, 4096)    # q_padded1, matmul 1
```

After `29_after_MemoryReuse.py`, the second matmul's buffers are collapsed onto the first's — only two Mat allocs remain, and both matmuls' K-loads reference `mem_mat_12`:

```
mem_mat_12 = alloc(Mem.Mat, 32768)   # now used by BOTH k_tile0 and k_tile1
mem_mat_13 = alloc(Mem.Mat, 4096)    # now used by BOTH q_padded0 and q_padded1
# mem_mat_17 / mem_mat_18 eliminated; Left/Right/Acc likewise collapsed
```

**Resulting hazard in codegen** (`kernels/aic/fa_qks_aic.cpp`):

```
line 90:  TMOV(v44, v32)                              // matmul0: last read of K-buffer mem_mat_12
line 91:  set_flag(PIPE_MTE1, PIPE_MTE2, EVENT_ID1)   // signal MTE1 done with mem_mat_12
...
[...]

PR #2276, opened 2026-08-04
chore(runtime): bump simpler to c6346935 and adapt to log-tier retirement

lyfne123:
- chore(runtime): bump simpler to c6346935 and adapt to log-tier retirement

### DIFF
--- a/docs/en/dev/03-logging.md
+++ b/docs/en/dev/03-logging.md
@@ -161,7 +161,7 @@ respect them.
 | Option | Default | Drives |
 | ------ | ------- | ------ |
 | `--pypto-log-level` | `ERROR` | PyPTO C++ logger via `set_log_level(LogLevel[name])` |
-| `--runtime-log-level` | unset (keeps `v5`) | PyPTO runtime logger via `configure_log(level)` — note this **does not** pass `sync_pypto=True` |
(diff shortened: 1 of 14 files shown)


lyfne123:
Closing: the runtime submodule is being advanced separately, so this bump is superseded.

Recording what the run surfaced so it is not lost:

**1. simpler #1475 retired the `V0..V9` INFO sub-tiers** in favour of a `TIMING` (25)
severity level. `parse_level` maps any unknown name — including a leftover `"v5"` /
`"v9"` — to `TIMING`, so the stale literals kept working by coincidence rather than by
contract. `[STRACE]` markers also moved from `LOG_INFO_V9` to `LOG_TIMING`.
The one behavioural bug was in `_sync_to_pypto`, which mapped threshold 25 onto `WARN`;
since 25 is now simpler's *default* threshold, that silenced two PyPTO log bands out of
the box.

**2. The per-run resource isolation refactor broke PyPTO's persistent distributed mode.**
`Worker._execute_pending_domain_releases()` gained a required `resources` argument, and
CommDomain bookkeeping split from a single `Worker._live_domains` into an additional
per-run `resources.live_domains` that the end-of-run cursor sweeps via
`_release_all_live_domains(resources)`. `distributed_runner` detaches a retained domain
from `Worker._live_domains` only, so the handle stayed in the run's registry and was
freed at the first fence — breaking retention. This is tracked separately as a
private-API dependency problem.

PR #2294, opened 2026-08-05
fix: write recv_counts directly for self-send in all_to_all_v

### LINKED ISSUE #2295: WIP - builtin.tensor.all_to_all_v: non-deterministic self-send due to TNOTIFY fire-and-forget
### Summary

`builtin.tensor.all_to_all_v` unconditionally calls `TNOTIFY(Set)` on the
self-destination slot to publish `recv_counts`, but `TNOTIFY` is fire-and-forget
with no completion signal.

### Root cause

```
for dest in 0..P-1:
    TNOTIFY(self, rows, Set)    // fire-and-forget — may or may not land
    TPUT(dst, src)              // data transfer
    pipe_barrier(PIPE_ALL)      // waits for TPUT only, NOT TNOTIFY
dsb(DSB_DDR)                    // DDR mem barrier — does not drain interconnect
```

Whether the self-write lands before the downstream consumer reads `recv_counts`
depends on:

1. **Pipeline depth** — how many in-flight TNOTIFYs the interconnect can buffer.
2. **Self-loopback path** — different NPU generations route self-destination
   through different physical paths with different timing.
3. **`dsb(DSB_DDR)` semantics** — orders local stores but a self-destined
   TNOTIFY is a remote-initiated write and may not be drained.

This makes the test **non-deterministic across CI runners** (different NPU
hardware) and potentially across runs on the same runner.

### Impact

- `test_host_tensor_all_to_all_v[2]` fails with `recv_counts=0` on affected HW.
- CI flakiness: PR #2280 and PR #2294 both hit this on unlucky runners while
  the original merge PR #2243 landed on a runner where it passed.

### Why only `all_to_all_v`

`all_to_all_v` is the sole builtin that **must** include self in its loop
(the data-transfer `TPUT` copies the self-destination block from input to
output).  Every other builtin skips self with `if (peer == my_rank) continue;`
before `TNOTIFY`.

### Fix

PR #2294 adds a `dest == my_rank` guard using `__atomic_store_n` (GCC/CCE
builtin, address-space aware for `__gm__` pointers, `__ATOMIC_RELEASE`) —
synchronous, completes before next instruction, no interconnect.

georgebisbas:
On some NPU hardware generations, `TNOTIFY(Set)` to self may be dropped by the
interconnect.  The root cause is non-deterministic: `TNOTIFY` is fire-and-forget
with no completion signal — whether the self-write lands before the consumer
reads `recv_counts` depends on pipeline depth, self-loopback path timing, and
the fact that `dsb(DSB_DDR)` orders local stores but may not drain in-flight
self-destined remote writes (see #2295 for details).

This is the only builtin that must include self in its loop (the data-transfer
`TPUT` needs to copy the self-destination block from input to output).  Every
other builtin skips self entirely with `if (peer == my_rank) continue;` before
`TNOTIFY`.

**Fix**: adds a `dest == my_rank` guard that writes `recv_counts` via
`__atomic_store_n` with `__ATOMIC_RELEASE` — a synchronous store guaranteed to
complete before the next instruction, no interconnect involved.  The builtin is
address-space aware (works with `__gm__` pointers) and already used elsewhere
in the codebase (`comm_sim.cpp`, `sdma_completion_scheduler.h`, etc.).

**Test**: `tests/st/distributed/test_l3_host_tensor_all_to_all_v.py::`\
`TestL3HostTensorAllToAllV::test_host_tensor_all_to_all_v[2]`

Closes #2295.

### DIFF
--- a/python/pypto/runtime/builtins/collectives/all_to_all_v/templates/kernel.cpp.in
+++ b/python/pypto/runtime/builtins/collectives/all_to_all_v/templates/kernel.cpp.in
@@ -131,9 +131,18 @@ extern "C" __aicore__ __attribute__((always_inline)) void kernel_entry(__gm__ in
     if (rows64 > max_recv) rows64 = max_recv;
     int32_t rows = static_cast<int32_t>(rows64);
 
-    __gm__ int32_t *remote_recv_slot = CommRemotePtr(comm_ctx, recv_counts_base + my_rank, dest);
-    pto::comm::Signal count_sig(remote_recv_slot);
-    pto::comm::TNOTIFY(count_sig, rows, pto::comm::NotifyOp::Set);
+    if (dest == my_rank) {
+      // Self-send: write recv_counts via compiler atomic builtin — a
+      // TNOTIFY to self may be dropped by the NPU interconnect on some
+      // hardware generations.  __atomic_store_n is a GCC/Clang builtin
+      // that the CCE compiler supports natively (no <atomic> needed),
+      // and it is address-space aware (works with __gm__ pointers).
+      __atomic_store_n(recv_counts_base + my_rank, rows, __ATOMIC_RELEASE);
+    } else {
+      __gm__ int32_t *remote_recv_slot = CommRemotePtr(comm_ctx, recv_counts_base + my_rank, dest);
+      pto::comm::Signal count_sig(remote_recv_slot);
+      pto::comm::TNOTIFY(count_sig, rows, pto::comm::NotifyOp::Set);
+    }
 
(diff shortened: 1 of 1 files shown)


georgebisbas:
Closing: deep analysis of the NPU TNOTIFY(Set) implementation in pto-isa/include/pto/comm/a2a3/TNotify.hpp shows it is a plain *sigPtr = value store with dcci + dsb(DSB_DDR) — deterministic on all NPU hardware for self-destination. The original recv_counts=0 failure cannot be explained by the TNOTIFY code path. Reopening if reproducible evidence emerges.

PR #2317, opened 2026-08-07
examples: distributed pld teaching tutorials (steps 01-07) + walkthroughs + CI gate

georgebisbas:
## Summary

Starts the distributed (`pld`) teaching tutorials in `examples/distributed/`:
seven small, golden-validated programs that teach the execution model and the
distributed substrate one concept at a time, plus a tutorial walkthrough page
(en + zh) per step and the tutorial-index / abstractions-map page.

Reveal discipline: every builtin is built by hand from primitives first and
only revealed later (step 04 builds an N-rank barrier from `notify`/`wait`
before `pld.tensor.barrier` appears; all-reduce reveals land in follow-ups).
Progression: every step uses only concepts introduced in earlier steps (or the
prerequisite chapter); a later-taught idea appears only as an explicit pointer,
never as required knowledge.

## Tutorial steps (this PR)

| Step | Program | Teaches |
| ---- | ------- | ------- |
| 01 | `01_hello_rank.py` | Rank identity, `pld.world_size()`, `DistributedConfig`, per-rank dispatch |
| 02 | `02_programming_model.py` | Three levels: `@pl.jit.host` → `@pl.jit` → `@pl.jit.incore` |
| 03 | `03_window_buffer.py` | Symmetric window memory (`alloc_window_buffer`/`window`), own-slice, 4 KiB floor |
| 04 | `04_barrier.py` | `notify(AtomicAdd)`/`wait(Ge)` reusable N-rank barrier, then reveal `pld.tensor.barrier` |
| 05 | `05_remote_load_store.py` | Tile-level RMA ring shift (`remote_load`/`remote_store`) |
| 06 | `06_put_get.py` | Tensor-level push/pull (`put`/`get`); correct get protocol (each rank notifies the rank that reads from it) |
| 07 | `07_dynamic_rank_count.py` | Dynamic rank count (`NR = pl.dynamic("NR")`): one source compiles and runs at any P; CI-validated at P=2/3/4 |

Steps 08–16 (all-reduce three ways + reveal, broadcast, allgather,
reduce_scatter, all_to_all, composition) are planned in follow-up PRs.

## Docs

- `docs/en+zh/user/distributed/05-tutorials.md` — 16-step tutorial index + the
  abstractions map (every `pld` abstraction → chapter section → tutorial step).
- `docs/en+zh/user/distributed/06-hello_rank.md … 12-dynamic_rank_count.md` —
  tutorial walkthroughs (prerequisites, run it, walkthrough, edge cases with
[...]

PR #915, opened 2026-04-08
feat(jit): implement @pl.jit preprocessor over @pl.program pipeline

huxinyuan1215:
## Summary

Implements the `@pl.jit` decorator as specified in issue #878 / #915.

- **`python/pypto/jit/decorator.py`** — `JITFunction` class with `@pl.jit` and `@pl.jit.incore` decorators. Supports Style A (inline `pl.at(level=CORE_GROUP)` scope) and Style B (explicit multi-function cross-call). `__call__` specializes, runs the full pass pipeline, and caches the resulting `CompiledProgram` in an L1 in-memory cache keyed by `(source_hash, shapes, dtypes, scalar_values)`. Dynamic dims (via `bind_dynamic`) are stored as `None` in the key so different concrete values share the same cache entry. `platform` is extracted from `RunConfig` at compile time and stored in `CompiledProgram` so execution uses the correct target.
- **`python/pypto/jit/specializer.py`** — AST transformer that converts `@pl.jit` source into type-annotated `@pl.program` DSL source. Inlines static shape dims as `ConstInt`, preserves dynamic dims as `Var`, rewrites `a.shape[i]` references, removes `bind_dynamic` calls, and promotes `pl.dynamic("M")` declarations to module level. Entry functions with `with pl.incore():` scopes are emitted as `FunctionType.Opaque` (not `Orchestration`) so that `OutlineIncoreScopes` can outline them correctly.
- **`python/pypto/jit/cache.py`** — Cache key construction with PyPTO version stamp for automatic invalidation on upgrades.
- **`python/pypto/language/typing/tensor.py`** — Adds `bind_dynamic(dim, dynvar)` method to `Tensor`.
- **`python/pypto/language/__init__.py`** — Exports `jit`, restores `PtrType`/`Ptr` aliases.

Bug fixes:
- Propagate dynamic dims from `@pl.jit.incore` deps back to entry function so params use `Var(M)` not `ConstInt(64)` (`_propagate_dynamic_dims_from_deps`, `_backfill_entry_dynvar_bindings`).
- Emit `FunctionType.Opaque` (not `Orchestration`) for entry functions containing `with pl.incore():` scopes; `OutlineIncoreScopes` only processes `Opaque` functions, so using `Orchestration` caused `tile.alloc` to be misplaced.
- Forward `platform` from `RunConfig` through `_compile()` to `ir.compile()` so `CompiledProgram` stores the correct execution platform instead of defaulting to `"a2a3sim"`.
[...]

PR #2330, opened 2026-08-10
examples: distributed all-reduce tutorials (steps 08-11) + walkthroughs + CI P=4 legs

georgebisbas:
## Summary

Second PR of the distributed teaching ladder: the all-reduce in
depth. Four small, golden-validated programs (steps 08–11) build the
collective by hand three ways — mesh, two-phase, ring — and then reveal the
builtin, each with a walkthrough page (en + zh) and a CI P=4 leg so the
comparisons are actually observable.

> **Rebased on current `main` (2026-08-11) and review-fixed.** #2317 is merged
> (squash `016e69a8`); this branch is `main` + the steps 08–11 work, squashed
> into one commit (head `659def1e`). Review round addressed: the ring uses
> neighbour-ready handshakes instead of per-round full-mesh barriers (all
> `pl.store`s rebound), the reveal docs state the InCore composite accepts the
> full `ReduceOp` family + `FP16`/`FP32` in both modes (the `Sum`+`FP32` limit
> is the HOST-builtin ring path), and the quickstart/index wording is
> corrected. Ring + reveal re-sim-validated at P=2/P=4 on a2a3sim.

## Tutorial steps (this PR)

| Step | Program | Teaches |
| ---- | ------- | ------- |
| 08 | `08_allreduce_mesh.py` | All-reduce v1 (mesh): every rank reads every peer's slice and sums locally, behind a dedicated-row notify/wait barrier |
| 09 | `09_allreduce_two_phase.py` | All-reduce v2: reduce-scatter + all-gather (`SIZE // P` chunks, `SIZE % P == 0`) |
| 10 | `10_allreduce_ring.py` | All-reduce v3 (ring): chunked rotation around the ring, `2·(P−1)` rounds, neighbour-ready handshakes |
| 11 | `11_allreduce_reveal.py` | **The reveal**: `pld.tensor.allreduce` (mesh + ring modes); the walkthrough diffs the lowered IR against the hand-rolled versions |

Each program uses the class-form factory (`build_* (nr)` → `@pl.program`) so
the rank count folds into static window shapes — the same source serves any P
via `-d` (window shapes must be statically known; see the plan's implementation
note).

## Docs

- `docs/en+zh/user/distributed/13-allreduce_mesh.md` … `16-allreduce_reveal.md`
  — one walkthrough per step (The idea → Run it → Walkthrough → Edge cases →
  See also, cost card + three-column debugging table each; step 16 ships a
  guide to the IR diff as the teaching artifact).
- `docs/en+zh/user/distributed/05-tutorials.md` — steps 08–11 marked ✅ shipped;
  steps 12–15 (the remaining collectives) and step 16 (composition) remain
  planned.
- Cross-links: `01-collectives.md` §Runnable Examples (allreduce row → the four
  walkthroughs), `00-model.md` Quickstart ↔ the mesh walkthrough (the
  quickstart is the mesh pattern).
- `mkdocs.yml` — nav entries for 13–16.

## CI

- `examples-tests` (from #2317) gains the P=4 legs for steps 08–11 — mesh /
  two-phase / ring collapse to the same exchange at P=2, so the P=4 legs are
  where the comparisons are observable; step 11 runs both `--mode mesh` and
  `--mode ring`.
- Docs gates (nav / en-zh parity / pre-commit / `mkdocs build --strict`) green.

## Verification

- Steps 08–11 sim-validated on `a2a3sim` at P=2/3/4 (reveal in both modes),
  goldens vs torch with tolerance.
- Developer gate: 4-card NPU run
  (`-p a2a3 -d 0,1,2,3`) via `task-submit` where available; else 2-card +
  sim-P=4 evidence.

### DIFF
--- a/docs/en/user/distributed/00-model.md
+++ b/docs/en/user/distributed/00-model.md
@@ -13,6 +13,13 @@
 
 The simplest distributed program — two ranks sum their data, both see the same result.
 
+> This is the **mesh all-reduce** pattern — stage in, barrier, read every
+> peer's slice and sum — which the [mesh allreduce walkthrough](13-allreduce_mesh.md)
+> builds step-by-step in the [tutorial ladder](05-tutorials.md). At two ranks
+> every algorithm collapses to this one exchange; the
+> [ring allreduce walkthrough](15-allreduce_ring.md) is the last manual step
+> before the builtin is revealed.
+
 ```python
 import pypto.language as pl
 import pypto.language.distributed as pld
@@ -207,6 +214,7 @@ InCore kernel (@pl.jit.incore)
 
 ## See Also
 
+- [05-tutorials](05-tutorials.md) — The step-by-step distributed tutorial ladder
 - [01-collectives](01-collectives.md) — Built-in collectives and their semantics
 - [02-primitives](02-primitives.md) — The substrate beneath the collectives
 - [03-execution](03-execution.md) — DistributedWorker lifecycle and production patterns
--- a/docs/en/user/distributed/05-tutorials.md
+++ b/docs/en/user/distributed/05-tutorials.md
@@ -1,14 +1,16 @@
 # Distributed Tutorials
 
 The `pld` vocabulary taught step by step: a sixteen-step tutorial series, one
-concept per program. Seven runnable examples ship now — from "hello rank"
-through point-to-point moves and the dynamic rank count; steps 08–16 (the
-collectives) are planned.
+concept per program. Eleven runnable examples ship now — from "hello rank"
+through point-to-point moves, the dynamic rank count, and the all-reduce trio
+plus its reveal; steps 12–15 (the remaining collectives) and step 16
+(composition) are planned.
 
 > **Prerequisites:** the [Distributed Programming](../distributed/index.md)
 > chapter — read it once for the vocabulary, then come back here to build the
-> same ideas by hand. Hardware: two devices for steps 01–06, three or more for
-> step 07 (any rank count), four for the collective comparisons in steps 08–16.
+> same ideas by hand. Hardware: two devices for steps 01–06, any count ≥ 2 for
+> step 07 (three or more to see the ring differ from P=2), four for the
+> collective comparisons in steps 08–11.
 
 ## The idea
 
@@ -44,8 +46,8 @@ idea from the primitives before a builtin replaces it:
 ## Suggested reading order
 
 Read the steps in order — **01 → 02 → 03 → 04 → 05 → 06 → 07 → 08 → 09 → 10 →
-11 → 12 → 13 → 14 → 15 → 16**. Every page repeats this block. Steps 08–16 are
-planned; 01–07 ship in the first PR (below).
+11 → 12 → 13 → 14 → 15 → 16**. Every page repeats this block. Steps 01–11 ship
+now; 12–16 remain planned.
 
 ## The 16 steps
 
@@ -58,18 +60,18 @@ planned; 01–07 ship in the first PR (below).
 | 05 | `05_remote_load_store.py` | Tile-level RMA: `remote_load`/`remote_store`; one-step ring shift | ✅ shipped |
 | 06 | `06_put_get.py` | Tensor-level p2p: `put`/`get`; push vs pull | ✅ shipped |
 | 07 | `07_dynamic_rank_count.py` | Dynamic rank count: `pl.dynamic("NR")`; one source, any P | ✅ shipped |
-| 08 | `08_allreduce_mesh.py` | All-reduce v1 (mesh): every rank reads every peer, sums locally | planned |
-| 09 | `09_allreduce_two_phase.py` | All-reduce v2: reduce-scatter + all-gather | planned |
-| 10 | `10_allreduce_ring.py` | All-reduce v3 (ring): chunked around the ring | planned |
-| 11 | `11_allreduce_reveal.py` | **The reveal**: `pld.tensor.allreduce` (mesh + ring); diff the IR | planned |
+| 08 | `08_allreduce_mesh.py` | All-reduce v1 (mesh): every rank reads every peer, sums locally | ✅ shipped |
+| 09 | `09_allreduce_two_phase.py` | All-reduce v2: reduce-scatter + all-gather | ✅ shipped |
+| 10 | `10_allreduce_ring.py` | All-reduce v3 (ring): chunked around the ring | ✅ shipped |
+| 11 | `11_allreduce_reveal.py` | **The reveal**: `pld.tensor.allreduce` (mesh + ring); diff the IR | ✅ shipped |
 | 12 | `12_broadcast.py` | One-to-all; reveal `pld.tensor.broadcast` | planned |
 | 13 | `13_allgather.py` | All-to-all slices; reveal `pld.tensor.allgather` | planned |
 | 14 | `14_reduce_scatter.py` | All-to-chunks; reveal `pld.tensor.reduce_scatter` | planned |
 | 15 | `15_all_to_all.py` | Personalized exchange; reveal `pld.tensor.all_to_all` | planned |
 | 16 | `16_putting_it_together.py` | Compose `broadcast` + `allreduce` + `allgather` in one kernel | planned |
 
-Steps 08–16 are **planned** — they arrive in later PRs. The walkthroughs below
-(06–12) cover steps 01–07, which ship together.
+Steps 12–16 are **planned** — they arrive in later PRs. The walkthroughs below
+(06–16) cover steps 01–11.
 
 ## The abstractions map
 
--- a/docs/en/user/distributed/15-allreduce_ring.md
+++ b/docs/en/user/distributed/15-allreduce_ring.md
@@ -1,0 +1,120 @@
+# All-Reduce 3: Ring — Constant Per-Step Size
+
+The two-phase shape from step 09, but each rank only ever moves data with its
+**left neighbour**. Instead of reading `P-1` peers per stage, the chunks rotate
+around the ring: `2 * (P-1)` steps, each moving one `N/P`-sized chunk — so the
+per-step transfer stays `N/P` as `P` grows under weak scaling (for a fixed
+`N` the chunk shrinks instead). Synchronization is
+neighbour-local too: each rank notifies its **right** neighbour after a store
+and waits on its **left** neighbour before a read — no full barrier per round.
+
+> **Prerequisites:** [14-allreduce_two_phase](14-allreduce_two_phase.md).
+> Four sim devices recommended (the constant per-step size is only observable
+> at P≥4); `SIZE` must divide evenly by the rank count.
+
+**Suggested reading order:** 01 → … → 09 → **10** — this page is step 10.
+
+## The idea
+
+Two-phase halves mesh's traffic but still reads every peer per stage — O(P)
+peers, each a `N/P` chunk. The ring removes the "every peer" part: arrange the
+ranks in a circle and pass chunks to the **left neighbour**. The same
+reduce-scatter + all-gather split now takes `P-1` steps each:
+
+- **Reduce-scatter (P-1 steps):** in each step a chunk travels one hop and is
+  added at its destination. After `P-1` steps every rank holds the reduced
+  chunk it owns.
+- **All-gather (P-1 steps):** the reduced chunks keep circulating, and each
+  rank copies every chunk as it passes. After `P-1` more steps every rank has
+  the full result.
+
+Total bytes equal two-phase (`2 * (P-1) / P * N`), but each step moves only
+`N/P` — under weak scaling (a workload that grows with `P`) the per-step size
+stays constant as `P` grows, which is what keeps the ring efficient at large
+world sizes; for a fixed `N` the chunk would shrink instead.
+
+## Run it
+
+```bash
+# P=4 (the constant per-step size shows here) and P=2:
+python examples/distributed/10_allreduce_ring.py -p a2a3sim -d 0,1,2,3
+python examples/distributed/10_allreduce_ring.py -p a2a3sim -d 0,1
+```
+
+Expected output:
+
+```text
+OK
+```
+
+## Walkthrough
+
+The kernel is monolithic (one InCore function) and the signal generalises the
+two-phase idea: `[2 * (nr-1), nr]` — **one row per round**. The chunk index
+arithmetic is the heart of the schedule:
+
+```python
+left = (my_rank - 1 + nranks) % nranks      # never negative
+
+# Reduce-scatter: (nr-1) steps.
+for s in pl.range(nranks - 1):
+    step = s + 1
+    recv_add_idx = (my_rank - step - 1 + nranks) % nranks
+    left_send_idx = (left - step + nranks) % nranks
+    # Wait for the left neighbour's round-s chunk (signal row s), then:
+    pld.system.wait(signal, offsets=[s, left], expected=1, cmp=pld.WaitCmp.Ge)
+    recv = pld.tile.remote_load(scratch, peer=left,
+                                offsets=[0, left_send_idx * chunk_elems],
+                                shape=[1, chunk])
+    acc = pl.load(scratch, [0, recv_add_idx * chunk_elems], [1, chunk])
+    acc = pl.add(acc, recv)
+    scratch = pl.store(acc, [0, recv_add_idx * chunk_elems], scratch)
+    # The store stages next round's send: signal the right neighbour (row s+1).
+    pld.system.notify(signal, peer=right, offsets=[s + 1, my_rank],
+                      value=1, op=pld.NotifyOp.AtomicAdd)
+
+# All-gather: (nr-1) steps (rows nranks-1 .. 2*(nranks-1)-1), copying the
+# left neighbour's send chunk into the local chunk.
+```
+
+- **`left = (my_rank - 1 + nranks) % nranks`.** The `+ nranks` keeps the
+  dividend non-negative — a bare `(my_rank - 1) % nranks` yields `-1` at
+  rank 0 under truncating modulo (the step-06 lesson, now on the index side).
+- **Chunks rotate, not ranks.** In round `s`, the chunk you add from the left
+  and the chunk you forward both shift by one (`- step`), so every chunk
+  visits every rank exactly once per phase.
+- **Neighbour-ready handshakes, not barriers.** Each round gets its own signal
+  row — notify the **right** neighbour after a store, wait on the **left**
+  neighbour before a `remote_load`. The monotonic `Ge(1)` counters never leak
+  between rounds (the step-09 discipline, now at `2*(P-1)` rows), and only the
+  two adjacent ranks ever synchronize: O(P) signals **per rank** (O(P²)
+  system-wide) — versus the O(P²) per rank a full-mesh barrier per round would
+  cost.
+
+**Cost card (per rank):** `2 * (P-1) / P * N` total — the same as two-phase —
+but in `2*(P-1)` steps of `N/P` bytes each. Under weak scaling the per-step
+size stays **constant as P grows** — unlike mesh's `(P-1) * N` per step —
+which is the ring's reason to exist.
+
+## Edge cases
+
+> **Fatal pitfall — a negative left-neighbour index.** `(my_rank - 1) %
+> nranks` is `-1` at rank 0 under truncating modulo — an invalid peer the
+> `remote_load` then targets. **Fix:** always write
+> `(my_rank - 1 + nranks) % nranks` (at P=2 rank 0's left neighbour is rank 1;
+> the `+ nranks` keeps the dividend non-negative).
+
+| Symptom | Likely cause | Fix |
+| ------- | ------------ | --- |
+| Hang at P=2 | Negative dividend in the left-neighbour index | `(my_rank - 1 + nranks) % nranks` |
+| Wrong chunks in the result | Chunk index arithmetic off by one in a round | Trace `recv_add_idx` / `left_send_idx` for `s=0` on paper |
+| Two handshakes share a signal row | Row index reused across RS and AG | RS rows `0..P-2`, AG rows `P-1..2(P-1)-1` |
+| Result correct at P=2 only | P=2 has a single round, hiding rotation bugs | Run P=4 and check every chunk position |
+| Same result on every rank but ≠ torch sum | Reduction order differs (not a bug) | Compare with a tolerance |
+
+## See also
+
+- [05-tutorials](05-tutorials.md) — the tutorial index (this step = row 10)
+- [14-allreduce_two_phase](14-allreduce_two_phase.md) — the two-phase shape (step 09)
+- [01-collectives](01-collectives.md) §AllReduce — reference (Ring Mode, signal shape, `Sum`/`FP32`)
+- Next step: [16-allreduce_reveal](16-allreduce_reveal.md) — the builtin that picks mesh or ring for you
--- a/docs/en/user/distributed/15-allreduce_ring.md
+++ b/docs/en/user/distributed/15-allreduce_ring.md
@@ -64,11 +64,11 @@ for s in pl.range(nranks - 1):
     # Wait for the left neighbour's round-s chunk (signal row s), then:
     pld.system.wait(signal, offsets=[s, left], expected=1, cmp=pld.WaitCmp.Ge)
     recv = pld.tile.remote_load(scratch, peer=left,
-                                offsets=[0, left_send_idx * chunk_elems],
+                                offsets=[0, left_send_idx * chunk],
                                 shape=[1, chunk])
-    acc = pl.load(scratch, [0, recv_add_idx * chunk_elems], [1, chunk])
+    acc = pl.load(scratch, [0, recv_add_idx * chunk], [1, chunk])
     acc = pl.add(acc, recv)
-    scratch = pl.store(acc, [0, recv_add_idx * chunk_elems], scratch)
+    scratch = pl.store(acc, [0, recv_add_idx * chunk], scratch)
     # The store stages next round's send: signal the right neighbour (row s+1).
     pld.system.notify(signal, peer=right, offsets=[s + 1, my_rank],
                       value=1, op=pld.NotifyOp.AtomicAdd)
--- a/docs/en/user/distributed/16-allreduce_reveal.md
+++ b/docs/en/user/distributed/16-allreduce_reveal.md
@@ -59,10 +59,13 @@ y = pl.store(recv, [0, 0], y)
   takes `[nr, 1]`; ring takes `[2*(nr-1), nr]` — exactly the row-per-round
   signal step 10 taught. The factory folds both `nr` and `mode` in, so one
   source builds either variant (the class-form pattern from steps 08-10).
-- **What the builtin accepts (read this twice):** `pld.tensor.allreduce` takes
-  the full `ReduceOp` family (`Sum`/`Max`/`Min`/`Prod`) and `FP16`/`FP32` in
-  **both** modes — the mesh and ring ST suites run every operator and `FP16`
-  through the real pipelines.
+- **What the builtin accepts (read this twice):** `pld.tensor.allreduce` — the
+  **InCore composite** lowering used here — takes the full `ReduceOp` family
+  (`Sum`/`Max`/`Min`/`Prod`) and `FP16`/`FP32` in **both** modes — the mesh and
+  ring ST suites run every operator and `FP16` through the real pipelines.
+  A narrower `Sum`+`FP32`-only contract exists, but only on the separate
+  **HOST builtin** ring path (`builtin.tensor.allreduce_ring`, not used by
+  this tutorial) — see `01-collectives.md` §AllReduce.
 
 ### The IR diff (the teaching artifact)
 
--- a/docs/en/user/distributed/16-allreduce_reveal.md
+++ b/docs/en/user/distributed/16-allreduce_reveal.md
@@ -1,0 +1,111 @@
+# The Reveal: `pld.tensor.allreduce` in One Call
+
+You built all-reduce three ways by hand — mesh, two-phase, ring. Now the
+builtin: `pld.tensor.allreduce(data, signal, op=..., mode=...)` replaces the
+whole schedule with one call, and the IR diff shows what it lowers to: the
+hand-rolled pattern you wrote, or a better one.
+
+> **Prerequisites:** [13-allreduce_mesh](13-allreduce_mesh.md) ·
+> [14-allreduce_two_phase](14-allreduce_two_phase.md) ·
+> [15-allreduce_ring](15-allreduce_ring.md). Four sim devices recommended.
+
+**Suggested reading order:** 01 → … → 10 → **11** — this page is step 11.
+
+## The idea
+
+The reveal discipline, completed: step 04 revealed the barrier after you built
+it; this step reveals the collective after you built three of them. The point
+of the hand-rolled steps was not that you should write all-reduce by hand —
+it was that you should know *what the builtin chooses between*.
+
+`pld.tensor.allreduce` is the builtin for the whole ladder: it takes your
+window and signal, does the barrier + cross-rank reduce + store-back, and
+hands the reduced slice back. `mode=` picks the algorithm — `"mesh"` (default)
+or `"ring"`. The same golden as steps 08-10; none of the schedule.
+
+## Run it
+
+```bash
+# Both modes, at P=4 (and P=2):
+python examples/distributed/11_allreduce_reveal.py -p a2a3sim -d 0,1,2,3
+python examples/distributed/11_allreduce_reveal.py -p a2a3sim -d 0,1,2,3 --mode ring
+python examples/distributed/11_allreduce_reveal.py -p a2a3sim -d 0,1
+```
+
+Expected output:
+
+```text
+OK
+```
+
+## Walkthrough
+
+Your stage-in and stage-out stay; only the middle is replaced:
+
+```python
+# Phase 1 — stage this rank's slice into its window slot.
+local = pl.load(x, [0, 0], [1, SIZE])
+data = pl.store(local, [0, 0], data)
+
+# Phase 2 — the builtin: barrier + reduce + store-back, in one call.
+data = pld.tensor.allreduce(data, signal, op=pld.ReduceOp.Sum, mode=mode)
+
+# Phase 3 — stage-out: the reduced slice back to the local output.
+recv = pl.load(data, [0, 0], [1, SIZE])
+y = pl.store(recv, [0, 0], y)
+```
+
+- **The signal is yours, and its shape tells you which mode you're in.** Mesh
+  takes `[nr, 1]`; ring takes `[2*(nr-1), nr]` — exactly the row-per-round
+  signal step 10 taught. The factory folds both `nr` and `mode` in, so one
+  source builds either variant (the class-form pattern from steps 08-10).
+- **What the builtin accepts (read this twice):** `pld.tensor.allreduce` takes
+  the full `ReduceOp` family (`Sum`/`Max`/`Min`/`Prod`) and `FP16`/`FP32` in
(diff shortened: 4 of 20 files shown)


docs/en/user/distributed/00-model.md
```diff
@@ -13,6 +13,11 @@
 
 The simplest distributed program — two ranks sum their data, both see the same result.
 
+> The same program is built step-by-step in the
+> [ring allreduce walkthrough](15-allreduce_ring.md), the last manual step of
+> the [tutorial ladder](05-tutorials.md) before the builtin is revealed. This
+> quickstart is the 2-rank, all-in-one version.
+
```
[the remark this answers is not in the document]
georgebisbas: Fixed — the quickstart now identifies itself as the mesh all-reduce pattern (stage in, barrier, read every peer, sum) and links to 13-allreduce_mesh.md instead of calling itself "the same program" as the ring walkthrough; zh mirror updated to match.

docs/en/user/distributed/05-tutorials.md
```diff
@@ -1,14 +1,15 @@
 # Distributed Tutorials
 
 The `pld` vocabulary taught step by step: a sixteen-step tutorial series, one
-concept per program. Seven runnable examples ship now — from "hello rank"
-through point-to-point moves and the dynamic rank count; steps 08–16 (the
-collectives) are planned.
+concept per program. Eleven runnable examples ship now — from "hello rank"
+through point-to-point moves, the dynamic rank count, and the all-reduce trio
+plus its reveal; steps 12–16 (the remaining collectives) are planned.
```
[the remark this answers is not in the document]
georgebisbas: Fixed — the overview now states steps 12–15 cover the remaining collectives and step 16 covers composition, in both languages.

examples/distributed/10_allreduce_ring.py
```diff
@@ -0,0 +1,253 @@
...
+            """
+            ctx = pld.get_comm_ctx(scratch)
+            my_rank = pld.rank(ctx)
+            nranks = pld.nranks(ctx)
+            left = (my_rank - 1 + nranks) % nranks
+
+            # Phase 1 — stage-in: copy each local input chunk into scratch.
+            for c in pl.range(nranks):
+                src_tile = pl.load(x, [0, c * chunk_elems], [1, chunk])
+                pl.store(src_tile, [0, c * chunk_elems], scratch)
```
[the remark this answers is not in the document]
georgebisbas: Fixed — all four pl.store results in 10_allreduce_ring.py are now rebound to scratch/y, and the walkthrough 15 code snippets (en + zh) were updated to match.

examples/distributed/10_allreduce_ring.py
```diff
@@ -0,0 +1,253 @@
...
+                            op=pld.NotifyOp.AtomicAdd,
+                        )
+                for peer in pl.range(nranks):
+                    if peer != my_rank:
+                        pld.system.wait(
+                            signal,
+                            offsets=[rs_round, peer],
+                            expected=1,
+                            cmp=pld.WaitCmp.Ge,
+                        )
```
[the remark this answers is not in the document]
georgebisbas: Fixed — the full-mesh per-round barriers are replaced with neighbour-ready handshakes: notify the right neighbour after stage-in/store, wait on the left neighbour before each remote_load. Module docstring and walkthrough 15 (en + zh) updated to describe the neighbour-local synchronization and its O(P) total signal cost. Ring re-sim-validated at P=2 and P=4 on a2a3sim (golden OK).

examples/distributed/11_allreduce_reveal.py
```diff
@@ -0,0 +1,179 @@
...
+        default="0,1,2,3",
+        help="comma-separated device ids (any count >= 2 dividing SIZE)",
+    )
+    parser.add_argument(
+        "--mode",
+        type=str,
+        default="mesh",
+        choices=["mesh", "ring"],
+        help="builtin mode: mesh (default) or ring (Sum + FP32-only)",
+    )
```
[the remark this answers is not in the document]
georgebisbas: Fixed — the --device help text no longer mentions "dividing SIZE"; it now states any count >= 2 is supported.

docs/en/user/distributed/16-allreduce_reveal.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+- **The signal is yours, and its shape tells you which mode you're in.** Mesh
+  takes `[nr, 1]`; ring takes `[2*(nr-1), nr]` — exactly the row-per-round
+  signal step 10 taught. The factory folds both `nr` and `mode` in, so one
+  source builds either variant (the class-form pattern from steps 08-10).
+- **What the builtin narrows (read this twice):** ring is **`Sum` + `FP32`-only** —
+  a compile-time check; mesh accepts `FP16`/`FP32` and the full `ReduceOp`
+  family (`Sum`/`Max`/`Min`/`Prod`). If you need ring with `Max`, the builtin
+  won't give it — that is a real constraint, not a missing feature.
```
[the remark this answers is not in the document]
georgebisbas: Fixed — the walkthrough and the module docstring now state the Sum + FP32-only limit applies to the HOST builtin ring path only; the InCore composite this example calls accepts the full ReduceOp family and FP16/FP32 in both modes (as the ring intrinsic ST suite exercises). zh mirror updated too.

docs/en/user/distributed/16-allreduce_reveal.md
```diff
@@ -0,0 +1,110 @@
...
+- **The signal is yours, and its shape tells you which mode you're in.** Mesh
+  takes `[nr, 1]`; ring takes `[2*(nr-1), nr]` — exactly the row-per-round
+  signal step 10 taught. The factory folds both `nr` and `mode` in, so one
+  source builds either variant (the class-form pattern from steps 08-10).
+- **What the builtin narrows (read this twice):** the InCore composite this
+  example calls (explicit signal) accepts the full `ReduceOp` family
+  (`Sum`/`Max`/`Min`/`Prod`) and `FP16`/`FP32` in **both** modes — the
+  `Sum`+`FP32`-only restriction belongs to the **HOST builtin** ring path, not
+  to this InCore form. The ring intrinsic ST suite exercises `Max`/`Min`/`Prod`
+  and `FP16`.
```
lyfne123: [P1] The explicit-signal InCore composite does not currently support the full `ReduceOp` family. `DeduceTensorAllReduceType` accepts only `ReduceOp.Sum`, both mesh and ring lowerings assert the same restriction, and the existing unit tests require `Max` to fail. The ring intrinsic ST also covers only FP32 + Sum. Please document the current Sum-only contract, remove the unsupported ST-coverage claim, and update the Chinese mirror and the related troubleshooting row below.

docs/en/user/distributed/16-allreduce_reveal.md
```diff
@@ -0,0 +1,110 @@
...
+Compile with pass dumps enabled and diff the lowered IR of this step against
+your hand-rolled programs:
+
+- `--mode mesh` expands into **step 08's pattern**: a ready barrier on the
+  `[nr, 1]` signal, then `remote_load` + accumulate chunks — the mesh you
+  wrote, possibly chunked to UB-sized tiles.
+- `--mode ring` expands into **step 10's pattern**: `2*(nr-1)` rounds of
+  neighbour-ready handshakes on the `[2*(nr-1), nr]` signal, chunked `N/P`
+  transfers — the ring you wrote, with the push/read side chosen by the
+  lowering.
```
lyfne123: [P1] The current builtin ring lowering does not use step 10's neighbour-ready handshake. Each reduce-scatter and all-gather round calls `EmitNotifyAll` followed by `EmitWaitAll`, i.e. a full-mesh barrier. The hand-written example now has a different synchronization protocol. Please describe that difference (and update the Chinese mirror) rather than saying the lowered IR reproduces the step 10 handshake.

docs/en/user/distributed/16-allreduce_reveal.md
```diff
@@ -0,0 +1,110 @@
...
+**Cost card (per rank):** whatever mode you picked — `(P-1) * N` (mesh) or
+`2 * (P-1) / P * N` in `N/P` steps (ring). The builtin chooses the algorithm;
+you still choose the mode.
+
+## Edge cases
+
+> **Fatal pitfall — the wrong signal shape for the mode.** Pass a `[nr, 1]`
+> signal with `mode="ring"` (or the ring shape with mesh) and the lowering
+> rejects it — the signal *is* the contract for which schedule the builtin
+> runs. **Fix:** mesh → `[nr, 1]`; ring → `[2*(nr-1), nr]`.
```
lyfne123: [P2] Shape validation is currently asymmetric. Ring mode validates the `[2*(NR-1), NR]` relationship, but mesh lowering does not reject a sufficiently large ring-shaped INT32 signal; it only relies on the slots it indexes. Either add mesh shape validation if this is intended to be a strict contract, or document the actual behavior instead of promising rejection in both directions.

docs/en/user/distributed/15-allreduce_ring.md
```diff
@@ -0,0 +1,117 @@
...
+- **Neighbour-ready handshakes, not barriers.** Each round gets its own signal
+  row — notify the **right** neighbour after a store, wait on the **left**
+  neighbour before a `remote_load`. The monotonic `Ge(1)` counters never leak
+  between rounds (the step-09 discipline, now at `2*(P-1)` rows), and only the
+  two adjacent ranks ever synchronize — O(P) signals total, not the O(P²) a
+  full barrier per round would cost.
+
+**Cost card (per rank):** `2 * (P-1) / P * N` total — the same as two-phase —
+but in `2*(P-1)` steps of `N/P` bytes each. The per-step size is **constant
+as P grows**, unlike mesh's `(P-1) * N` per step; that is the ring's reason to
```
lyfne123: [P2] The scaling accounting mixes fixed-size and weak-scaling assumptions. For fixed `N`, an `N/P` chunk shrinks as `P` grows; it is constant only if `N` grows proportionally with `P`. Also, the neighbour-ready protocol performs O(P) notifications per rank but O(P²) across the whole system, so “O(P) signals total” is inaccurate. Please qualify the scaling model and scope the signal count explicitly in both language versions.

docs/en/user/distributed/15-allreduce_ring.md
```diff
@@ -0,0 +1,117 @@
...
+exist.
+
+## Edge cases
+
+> **Fatal pitfall — a negative left-neighbour index.** `(my_rank - 1) %
+> nranks` is `-1` at rank 0 under truncating modulo: the `remote_load` then
+> targets an invalid peer and the ring deadlocks at P=2 (the only case where
+> rank 0's left neighbour is itself). **Fix:** always write
+> `(my_rank - 1 + nranks) % nranks`.
```
lyfne123: [P3] At P=2, rank 0's correct left neighbour is rank 1, not itself. The unsafe expression can produce the invalid peer `-1` under truncating modulo, but that is unrelated to self-neighbouring. Please remove or correct the parenthetical explanation in both language versions.

lyfne123 (changes requested):
The hand-written algorithms are internally consistent and CI is green, but the new walkthroughs currently contain two blocking mismatches with the implementation: the supported reduction operators and the builtin ring synchronization protocol. I also found three smaller accuracy issues in the signal-shape and scaling explanations. Please update both the English pages and their Chinese mirrors.

docs/en/user/distributed/16-allreduce_reveal.md
georgebisbas: Thanks — I verified this against the current code, and the walkthrough's claim already matches the implementation: the builtin accepts the full `ReduceOp` family and FP16/FP32 in both modes.

- `DeduceTensorAllReduceType` (`src/ir/op/distributed/allreduce.cpp:111-113`) validates `kSum <= op <= kProd` — Sum/Max/Min/Prod are all accepted.
- `tests/ut/ir/test_distributed_ops.py` (`test_tensor_allreduce_accepts_every_reduce_op`, ~236-246) parametrizes Sum/Max/Min/Prod and asserts all are accepted (Max does not fail).
- The composite lowering (`src/ir/transforms/lower_composite_ops_pass.cpp`) maps Sum→`tile.add`, Max→`tile.maximum`, Min→`tile.minimum`, Prod→`tile.mul` in mesh mode, and the ring lowering accepts the same `[kSum..kProd]` range.
- Codegen (`src/codegen/distributed/distributed_ops_codegen.cpp:54-65`) emits TADD/TMAX/TMIN/TMUL for the four variants.
- The ring intrinsic ST (`tests/st/distributed/collectives/test_l3_tensor_allreduce_ring_intrinsic.py:230-246`, `test_ring_allreduce_reduce_ops`) exercises every operator through the real ring pipeline with no op-level skip, and `test_ring_allreduce_fp16_arbitrary_lengths` covers FP16.

I also removed the one remaining unsupported carve-out (the "HOST builtin is Sum+FP32-only" line) since the host codegen path supports the same four variants, and aligned the zh mirror and the troubleshooting row. If there is a specific path (a backend, an op, a shape) that still restricts the family, point me at it and I'll document that case.

docs/en/user/distributed/16-allreduce_reveal.md
georgebisbas: Fixed. The walkthrough now states that `--mode ring` lowers each round to a full-mesh barrier (`EmitNotifyAll`/`EmitWaitAll`) rather than step 10's neighbour-ready handshake — same schedule and chunks, different synchronization — and calls that difference out as the point of the IR diff. The zh mirror, the troubleshooting table, and the reveal example docstring are updated to match.

docs/en/user/distributed/16-allreduce_reveal.md
georgebisbas: Fixed. The walkthrough now documents the actual asymmetry instead of promising symmetric rejection: ring mode strictly validates the `[2*(nr-1), nr]` signal; mesh validates that a static column count is 1 (so a static ring-shaped signal is rejected) but does not validate the row count — it only reads the slots it indexes. The zh mirror and troubleshooting table are updated.

docs/en/user/distributed/15-allreduce_ring.md
georgebisbas: Fixed. The scaling model is now qualified: the per-step `N/P` transfer is constant as `P` grows *under weak scaling* (a workload that grows with `P`), with a note that for fixed `N` the chunk shrinks instead; and the signal count is scoped as O(P) notifications per rank (O(P²) system-wide), contrasted with the O(P²) per rank a full-mesh barrier per round would cost. Both language versions are updated.

docs/en/user/distributed/15-allreduce_ring.md
georgebisbas: Fixed. The parenthetical is corrected: at P=2 rank 0's left neighbour is rank 1 (its left and right neighbours coincide — not rank 0 itself), and the `(my_rank - 1) % nranks` hazard is truncating modulo producing the invalid peer `-1`, unrelated to self-neighbouring. Both language versions are updated.

georgebisbas:
@lyfne123 Thanks for the detailed review — round 2 is pushed.

- **Supported reduction operators:** I verified the claim against the code — the deducer, unit tests, both mesh and ring lowerings, codegen, and the ring intrinsic ST all cover Sum/Max/Min/Prod and FP16 (evidence with file/line citations in the thread). The walkthrough keeps the full-family statement, and I removed the last unsupported carve-out (the "HOST builtin is Sum+FP32-only" line) since the host codegen path supports the same four variants. If there is a specific path that still restricts the family, point me at it and I'll document that case.
- **Builtin ring synchronization:** the walkthrough now states that `--mode ring` lowers each round to a full-mesh barrier (`EmitNotifyAll`/`EmitWaitAll`), not step 10's neighbour-ready handshake, and frames that difference as the point of the IR diff.
- **Signal-shape asymmetry, scaling model, and the P=2 left-neighbour parenthetical:** all corrected in both English and Chinese pages, plus the troubleshooting table and the reveal example docstring.

All five threads have replies; four are resolved (the reduction-operators thread is left open for your confirmation). CI is re-running on the amended commit.

YunjiQin (commented):
Three consistency notes on the example sources, left inline. None of them are correctness bugs — I walked the ring index arithmetic and the handshake chain and the algorithm is right — they are all about the ladder staying legible as a teaching sequence, where a style drift costs more than it would in ordinary code.

examples/distributed/10_allreduce_ring.py
```diff
@@ -0,0 +1,268 @@
...
+    @pl.program
+    class RingAllreduce:
+        @pl.function(type=pl.FunctionType.InCore)
+        def ring_step(
+            self,
+            x: pl.Tensor[[1, SIZE], pl.FP32],
+            y: pl.Out[pl.Tensor[[1, SIZE], pl.FP32]],
+            scratch: pl.InOut[pld.DistributedTensor[[1, SIZE], pl.FP32]],
+            signal: pl.InOut[pld.DistributedTensor[[total_rounds, nr], pl.INT32]],
+            chunk_elems: pl.Scalar[pl.INDEX],
```
YunjiQin: **`chunk_elems` is redundant, and mixing it with `chunk` obscures the arithmetic this file exists to teach.**

It is threaded through `ring_step` and `per_rank` as a `pl.Scalar[pl.INDEX]`, but the host passes `chunk_elems = SIZE // nr` (line 200) — exactly the compile-time closure constant `chunk` already computed on line 55. The two are always equal, so the parameter buys nothing.

The cost is that offsets and shapes end up spelled differently on the same line:

```python
recv = pld.tile.remote_load(scratch, peer=left,
                            offsets=[0, left_send_idx * chunk_elems],  # runtime scalar
                            shape=[1, chunk])                          # compile-time const
```

A reader working through `recv_add_idx` / `left_send_idx` on paper — which the walkthrough explicitly asks them to do — will reasonably assume the split is meaningful. Suggest dropping the parameter from both signatures and using `chunk` for offsets and shapes alike.

examples/distributed/08_allreduce_mesh.py
```diff
@@ -0,0 +1,195 @@
...
+import pypto.language as pl
+import pypto.language.distributed as pld
+import torch
+from pypto import ir
+from pypto.ir.distributed_compiled_program import DistributedConfig
+
+SIZE = 64
+
+
+def build_mesh_allreduce(nr: int):
```
YunjiQin: **The ladder switches program-definition idiom here, and only half the switch is explained.**

Steps 01–07 use `@pl.jit.host` / `@pl.jit.incore` with `program.compile(x, y, config=RunConfig(...))`. Steps 08–11 use the `@pl.program` class form with `ir.compile(program, platform=..., distributed_config=...)` (line 172).

The docstring explains *why a factory is needed* (window shapes must be statically known and depend on `nr`) — that part is clear and correct. What it does not explain is why the factory also changes the decorator family and the compile entry point: a factory could equally return `@pl.jit.host`-decorated functions with `nr` closed over.

This lands immediately after step 07 taught "one source, any P" via `pl.dynamic("NR")` under `@pl.jit`, so a reader arriving at step 08 sees three things change at once with one of them motivated. A sentence or two in `13-allreduce_mesh.md` §Walkthrough covering the decorator/compile switch (or, if the class form is not actually required, keeping `@pl.jit` inside the factory) would close the gap.

examples/distributed/08_allreduce_mesh.py
```diff
@@ -0,0 +1,195 @@
...
+
+            # Phase 1 — stage this rank's slice into its window slot.
+            local = pl.load(x, [0, 0], [1, SIZE])
+            data = pl.store(local, [0, 0], data)
+
+            # Phase 2 — barrier: notify every peer, wait on every peer slot.
+            # Each rank owns a dedicated row (offsets=[my_rank, 0]);
+            # AtomicAdd/Ge(1) means the wait only passes once every peer has
+            # staged its slice.
+            for peer in pl.range(nr):
```
YunjiQin: **Four adjacent teaching files use four different rank-count spellings.**

| File | kernel loops | host orchestration loop |
| --- | --- | --- |
| 08 | `pl.range(nr)` — closure constant | `pl.range(pld.world_size())` |
| 09 | `pl.range(nranks)` — `pld.nranks(ctx)` | `pl.range(pld.world_size())` |
| 10 | `pl.range(nranks)` | `pl.range(nr)` |
| 11 | — | `pl.range(nr)` |

All four are functionally equivalent here (`nr == world_size` by construction), so nothing is broken. But steps 08–11 are meant to be read back-to-back and diffed against each other, and step 07 has just finished making the point that the rank count is a runtime property. Having the very next step drop `nranks` for a closure constant, and then having 10/11 drop `pld.world_size()` on the host side, sends the opposite signal.

Suggest picking one convention and applying it across all four: `nranks = pld.nranks(ctx)` inside kernels (carries step 07's lesson forward), `pld.world_size()` in host orchestration (matches step 07's source). `nr` then stays reserved for what genuinely needs to be compile-time — the window shapes — which is exactly the distinction the factory docstrings are trying to draw.

examples/distributed/10_allreduce_ring.py
georgebisbas: Agreed — removed in 6a918fa2 (the parameter, both signatures, the call site, and the host-side `chunk_elems = SIZE // nr`). All eight body uses now read `chunk`, so the index arithmetic is expressed in exactly one name.

You were right that it bought nothing: `chunk` is already the compile-time closure constant, and threading an always-equal `pl.Scalar[pl.INDEX]` alongside it just gave the reader two names for one quantity in the file whose whole subject is the index arithmetic.

Verified in the simulation image at P=2 and P=4.

examples/distributed/08_allreduce_mesh.py
georgebisbas: Fixed in 6a918fa2. The four files now use one spelling each, by role:

| File | kernel loops | host orchestration loop |
| --- | --- | --- |
| 08 | `pl.range(nranks)` — `pld.nranks(ctx)` | `pl.range(pld.world_size())` |
| 09 | `pl.range(nranks)` | `pl.range(pld.world_size())` |
| 10 | `pl.range(nranks)` | `pl.range(pld.world_size())` |
| 11 | — | `pl.range(pld.world_size())` |

The rule, applied uniformly: a **kernel** derives the rank count at runtime from its comm context (`pld.nranks(ctx)`), a **host orchestrator** uses `pld.world_size()`, and a bare `nr` is reserved for compile-time *shapes* only — never a loop bound.

Step 08 no longer has an `nr` at all (see the other thread), so the `pl.range(nr)` spelling is gone from the series entirely.

examples/distributed/08_allreduce_mesh.py
georgebisbas: Addressed in 6a918fa2, and your instinct to push on this was right — the explanation I had written was wrong in three separate ways.

**I tried the conversion you suggested.** It does not work, but not for the reason the docstring claimed. Converting 08 to the `@pl.jit` family fails at compile time with:

```
ValueError: @pl.jit: missing inferred tensor metadata for parameter 'signal'
  python/pypto/jit/specializer.py:1986 in _build_params
```

`signal` is a window shaped `[pld.world_size(), 1]`, and `@pl.jit` must statically infer shape and dtype for every parameter it forwards to a dependency. It is not distributed tensors as such — steps 03, 05, 06 and 07 all pass `pld.DistributedTensor` through `@pl.jit` deps happily; every one of those windows is statically shaped. It is specifically a runtime-sized window dim. (There is a second, independent gap too: `@pl.jit`'s constant folding reads only `func.__globals__` (`specializer.py:2101`), never `__closure__`, so a factory closure constant referenced in a HOST body cannot resolve either. Both look worth their own issues; neither is a regression — the message dates to #915.)

**But the factory was never needed here**, which is what the old text got wrong. Step 08 has no tile shape depending on the rank count, so the barrier signal row count can simply stay dynamic: `NR = pl.dynamic("NR")` in the annotations, `pld.world_size()` in the host body. 08 is now a module-level `@pl.program` with **no factory**, matching `tests/st/distributed/collectives/test_l3_allreduce.py` — the system test for this same collective.

So the switch is still required, but the honest reason is the *dynamic* signal shape, not a compile-time one; and the compile entry point follows from it, since `ir.compile(...)` is simply the class form's entry point. The walkthrough now says exactly that, with the error string quoted.

**Three corrections to the docs while I was in there**, all of which were my errors:

- It claimed *"every rank-parametrized collective in `tests/st/distributed/` uses this same class-form factory"*. False — three of five do not (`test_l3_allreduce.py`, `test_l3_allgather.py`, `test_l3_broadcast.py` are module-level with `pl.dynamic`).
- It attributed the switch to `@pl.program` snapshotting the defining frame's locals. Real limitation, wrong conclusion for this file.
- A troubleshooting row told readers a `[nr, 1]` shape must be a closure constant — which would have sent them to "fix" working code. Replaced with the actual error and its real remedy.

**Steps 09/10 do keep the factory**, and step 14 now states the sharp reason the docs previously lacked: their chunk size `SIZE // nr` is a **tile shape**, and tile shapes must be known when the kernel is compiled. A signal row count is not, which is why 08 does not need one. The idiom switch now marks a real constraint boundary rather than an arbitrary one.

Verified in the simulation image (pypto `10286517`, pto-isa `f51c92f6`): all four examples pass at P=2 and P=4, plus `11 --mode ring` at P=4.

georgebisbas:
@YunjiQin all three of your points are addressed and pushed — summary here since the branch has moved a fair bit since your review.

**Your three items**

| Item | Resolution |
| --- | --- |
| `chunk_elems` redundant with `chunk` | Removed — parameter, both signatures, call site, and the host-side `chunk_elems = SIZE // nr`. All 8 body uses now read `chunk`. |
| Idiom switch at 08 half-explained | 08 no longer has a factory at all, and the explanation was rewritten because it was wrong. Detail below. |
| Four rank-count spellings | Unified by role: kernels use `pld.nranks(ctx)`, host orchestrators use `pld.world_size()`, bare `nr` is reserved for compile-time shapes. `pl.range(nr)` is gone from the series. |

**On the idiom switch — you were right to push, and I was wrong twice**

I tried the `@pl.jit` conversion you suggested. It fails:

```
ValueError: @pl.jit: missing inferred tensor metadata for parameter 'signal'
  python/pypto/jit/specializer.py, _build_params
```

`@pl.jit` must statically infer shape and dtype for every parameter it forwards to a dependency, and `signal` is a window shaped `[pld.world_size(), 1]`. Not distributed tensors in general — steps 03/05/06/07 pass them through `@pl.jit` fine; specifically a runtime-sized window dim. Filed as #2450, along with #2449 for a second, independent gap (`@pl.jit` constant folding reads only `__globals__`, never `__closure__`).

But investigating it showed **the factory was never needed for step 08**. Its signal row count is not a tile shape, so it can stay dynamic: `NR = pl.dynamic("NR")` plus `pld.world_size()`. Step 08 is now a module-level `@pl.program` with no factory, matching `tests/st/distributed/collectives/test_l3_allreduce.py` — the system test for this same collective.

The old walkthrough was wrong in three ways, all mine:

- It claimed *every* rank-parametrized collective in `tests/st/distributed/` uses a class-form factory. Three of five do not — `test_l3_allreduce.py`, `test_l3_allgather.py`, `test_l3_broadcast.py` are module-level with `pl.dynamic`.
- It blamed the switch on `@pl.program` snapshotting the defining frame's locals. Real limitation, wrong conclusion for this file.
- A troubleshooting row told readers a `[nr, 1]` shape must be a closure constant — which would have sent them to "fix" working code.

Steps 09/10 keep their factories, and step 14 now gives the reason the docs previously lacked: their chunk size `SIZE // nr` is a **tile shape**, and tile shapes must be known at compile time. So the idiom switch now marks a real constraint boundary instead of an arbitrary one.

Also fixed `01-collectives.md` Ring Mode (`77d278cf`), which documented the `@pl.jit`-closure factory as the pattern on the grounds that "the specializer folds the closure constants" — it does not, per #2449.

**Verification** — simulation image, pypto `10286517` / pto-isa `f51c92f6`: all four examples pass at P=2 and P=4, plus `11 --mode ring` at P=4. Local gates green: ruff check + format, EN↔ZH parity, docs nav, English-only, headers.

Re-review when you have a moment — happy to revisit any of it.

docs/en/user/distributed/16-allreduce_reveal.md
georgebisbas: Following up after re-verifying everything against current `main` (`fef2831c`).

**First, you were right about one thing and I was wrong.** My earlier reply said I had removed the "HOST builtin is Sum+FP32-only" carve-out "since the host codegen path supports the same four variants". That claim was incorrect. The gate is still there and still live:

- `src/ir/op/distributed/collective.cpp:62` — `CheckSupportedSumFp32BuiltinVariant`
- called at `:194` and `:909`

So the HOST builtin ring path (`builtin.tensor.allreduce_ring`) genuinely is `Sum` + `FP32`-only. The docs say exactly that and are correct as written — `docs/en/user/distributed/01-collectives.md:103` and `16-allreduce_reveal.md:67`. Nothing needs changing there; my sentence about removing it was the error, not the docs.

**On the actual P1 — the explicit-signal InCore composite — the full `ReduceOp` family is supported.** Re-verified on current main, since line numbers moved since August 11:

- `src/ir/op/distributed/allreduce.cpp:107` — `DeduceTensorAllReduceType` validates `op_value >= kSum && op_value <= kProd`. Not Sum-only, and not mode-dependent.
- `src/ir/transforms/lower_composite_ops_pass.cpp:881` (mesh) and `:1171` (ring) — **both** assert the same `[kSum, kProd]` range, and both dispatch through the shared `Reduce(op, ...)` helper (line ~293) mapping Sum/Max/Min/Prod to `tile.add` / `tile.maximum` / `tile.minimum` / `tile.mul`.
- `tests/st/distributed/collectives/test_l3_tensor_allreduce_ring_intrinsic.py:228-238` — `test_ring_allreduce_reduce_ops` is parametrized over `(Sum, Max, Min, Prod)` and runs through the real ring pipeline with no op-level skip.

**One thing worth flagging, because it is a genuine trap.** There *is* a Sum-only assert in that lowering file:

```cpp
// src/ir/transforms/lower_composite_ops_pass.cpp:1783
INTERNAL_CHECK_SPAN(op_value == static_cast<int>(ReduceOp::kSum), span)
    << "pld.tensor.reduce_scatter lowering supports ReduceOp::kSum only ...";
```

That one belongs to `pld.tensor.reduce_scatter`, a different collective — not allreduce. A grep for `ReduceOp::kSum` in this file surfaces it right next to the allreduce paths, and I suspect that is what the original review found. It does not constrain `pld.tensor.allreduce` in either mode.

**Net:** the tutorial text is accurate as it stands — steps 08-11 use the InCore composite, which takes the full family; the narrower `Sum`+`FP32` contract belongs to the separate HOST builtin ring path, which these tutorials do not use, and which the docs already document correctly.

If you still see a specific path that restricts the family — a backend, an op, a shape — point me at it and I will document that case. Otherwise, could you re-review? This is currently the only review blocking the PR; CI is green across all 15 jobs including `dist-system-tests` and `examples-tests`.

examples/distributed/08_allreduce_mesh.py
YunjiQin: Thanks — the investigation and the revised step 08 explanation make sense. There is one remaining consistency gap in the example docstrings:

- Step 09 still says `A rank-count factory (as in step 08)` and attributes the factory to the `nr`-dependent window/signal shape. Step 08 no longer has a factory, and the actual compile-time constraint here is `chunk = SIZE // nr` being used as a tile shape.
- Step 10 still says the signal shape and chunk size “both must be statically known” and refers to steps 08/09. That conflicts with the corrected explanation that a signal row count alone need not be static; it is the chunk tile shape that forces compile-time `nr` here.
- Step 11 likewise says its factory is “as in steps 08-10” and attributes it to the signal shape. At minimum, the step 08 reference is stale; the rationale should distinguish mode specialization and its signal layout from the tile-shape constraint in steps 09/10.

Could you align these three docstrings with the corrected explanation before merge?

georgebisbas: Aligned in 9b37ba8f — all three, plus the step 11 walkthrough page (en + zh), which was the last place still describing the factory without the distinction.

**Step 09** — the factory now has one reason, and it is not the signal:

> A rank-count factory — the first step that needs one. Step 08 has none: its signal row count stays `pl.dynamic`. What forces a compile-time `nr` here is `chunk = SIZE // nr`, used as a **tile shape** (the `[1, chunk]` on every `pl.load` / `remote_load` below) [...] The `[2, nr]` signal then falls out of the same constant for free — but a signal shape on its own would not have required it.

**Step 10** — same single reason as 09; the "both must be statically known" claim is gone, since it was the exact sentence that contradicted the correction.

**Step 11** — you were right that this one needed its own rationale rather than a pointer at 08-10, and chasing it turned up something I could not settle by reading. The `pld.tensor.allreduce` docstring (`tensor_ops.py:624-628`) says the ring signal is `[2*(NR-1), NR]` "where `NR` must be a compile-time constant". But the lowering does not enforce that — `lower_composite_ops_pass.cpp:1182-1194` only *cross-checks* `shape[0] == 2*(shape[1]-1)` when both dims are `ConstInt`, and explicitly skips when either is dynamic.

So rather than repeat last round's mistake and assert it, I tested it: a module-level `@pl.program` copy of step 11 with `NR = pl.dynamic("NR")`, a `[2*(NR-1), NR]` ring signal and `pld.window(signal_buf, [2*(pld.world_size()-1), pld.world_size()], ...)` in the host body **compiles and passes its golden** on `a2a3sim` at `-d 0,1,2` and `-d 0,1,2,3`. A compile-time `nr` is not required in step 11 at all. The docstring now says so:

> A factory over `(nr, mode)`, and for a different reason than steps 09/10: the builtin owns the chunking, so no **tile shape** here depends on the rank count — `nr` need not be compile-time at all (the ring layout, the more constrained of the two, compiles and passes its golden with a `pl.dynamic` rank count). What must be fixed when the kernel is traced is `mode`: it picks both the lowering `pld.tensor.allreduce` emits and the signal layout the kernel is annotated with, and those are two different shapes, not two extents of one — mesh `[nr, 1]`, ring `[2*(nr-1), nr]`.

That makes the ladder's idiom boundary land where the real constraint is: 08 needs nothing, 09/10 need a compile-time `nr` for a tile shape, 11 needs a build-time `mode` and folds `nr` in beside it only so one source can spell both layouts.

The overstated line in the `pld.tensor.allreduce` docstring is a separate matter — tracked locally, not touched here, since it is API documentation rather than tutorial text.

Re-verified after the change: all four examples plus `11 --mode mesh` / `--mode ring` print `OK` at P=4 on `a2a3sim`; `pre-commit` (incl. en-zh parity, markdownlint, ruff, pyright) and `mkdocs build --strict` clean.

examples/distributed/10_allreduce_ring.py
georgebisbas: Dropped — `chunk_elems` is gone from both `ring_step` and `per_rank`, and offsets and shapes now both use the closure constant `chunk`:

```python
recv = pld.tile.remote_load(scratch, peer=left,
                            offsets=[0, left_send_idx * chunk],
                            shape=[1, chunk])
```

Your read of the cost was right: the split made the index arithmetic look like it carried a distinction it did not, on exactly the lines the walkthrough asks readers to work through on paper.

examples/distributed/08_allreduce_mesh.py
georgebisbas: Converged on the convention you proposed — `pld.nranks(ctx)` inside kernels, `pld.world_size()` in host orchestration, and `nr` reserved for what genuinely has to be compile-time.

Current state across the four files:

| File | kernel loops | host orchestration loop |
| --- | --- | --- |
| 08 | `pld.nranks(ctx)` (`:72`) | `pl.range(pld.world_size())` (`:130`) |
| 09 | `pld.nranks(ctx)` (`:74`) | `pl.range(pld.world_size())` (`:166`) |
| 10 | `pld.nranks(ctx)` (`:83`) | `pl.range(pld.world_size())` (`:201`) |
| 11 | — (one builtin call, no rank loop) | `pl.range(pld.world_size())` (`:112`) |

That also made the point you were making about step 07 land properly: `nr` now appears only where it is a real compile-time constraint, and the just-pushed 9b37ba8f makes each file say which constraint that is — a tile shape in 09/10, `mode` specialization in 11, and none at all in 08.

YunjiQin (approved):

lyfne123 (approved):

Issue #2354, opened 2026-08-12
[Bug] Advancing the Simpler pin onto the Buffer API stalls DeepSeek V4 decode after the first token (TENSOR_WAIT_TIMEOUT)


lterrac:
> [!IMPORTANT]
> **Update 2026-08-13:** this is **not** caused by the port in #2355 — it reproduces identically on #2345's own head (`75e4bcc4`) with our code absent, 3 runs. The "look in pypto / the port" narrowing below is withdrawn; see [this comment](https://github.com/hw-native-sys/pypto/issues/2354#issuecomment-5278994812). The four exclusions still stand.

### Summary

Advancing pypto's `runtime/` (Simpler) pin onto the Buffer device-memory API makes DeepSeek V4 Flash W8A8 serving **start and upload correctly but stall in decode after the first generated token**. Startup is clean (346.6 GB of resident weights uploaded, 0 faults, `model loaded` in 74.0s); the first token comes out; the second decode step dies with a tensor-wait timeout.

This is filed to record what has already been **excluded**, so the next person does not re-pay for it.

### Symptom

```
AICore error 507901: bounded device drain failed
Device[0] fault, ret=0x7110011
PTO2 runtime failed: orch_error_code=8 sched_error_code=100 runtime_status=-8
  orch_error_code=8 TENSOR_WAIT_TIMEOUT - waiting for tensor data timed out: the producing
  task never completed, or a consumer never released its fanout reference
  sub_class=S1:running-stalled (detail=1) completed=71/9 running=3 ready=0 waiting=0
  stuck_task_id=4294967357 stuck_core=41
→ finalize_native_run failed with code -8 (run_id=2 slot=0 generation=2 dispatch_id=3 run_epoch=3)
```

`stuck_task_id` decodes as `(ring 1 << 32) | 61`. Note `run_id=2` / `generation=2`: the **second** run of the persistent program, not the first.

Inference harness: `req 0: TRUNCATED 1/32`, then zero tokens on `req 1`. Both the streaming and non-streaming completion paths are affected identically, so it is not an API-surface problem.

### Environment

| Component | Value |
|---|---|
| pypto | `08f6c70f` + the pin-advance port (7 layers, see below) |
| Simpler | `8a82bf9e` |
| pypto-lib | `f0d352e` |
| pypto-serving | `71fd275` |
| PTOAS | v0.57 (as pinned by `toolchain/versions.env`) |
[...]

Issue #1681, opened 2026-06-04
[RFC] SuperscalarNPU backend: DDR + TREG register-file architecture


lyfne123:
### Summary

Add a new compiler backend, **SuperscalarNPU**, whose only memory spaces are **DDR** (off-chip) and **TREG** — a register file of **256 fixed 4 KB blocks (1 MB total)** addressed by block index. Unlike the Ascend backends, SuperscalarNPU has no cube/vector cores and no L1/L0 buffer hierarchy: on-chip data lives in TREG and is allocated by **register renaming** (liveness-based coalescing + block-index assignment with a register-pressure limit). This RFC tracks the end-to-end development of the backend, from IR generation through (eventually) code generation.

This issue is the umbrella tracker. Phase 1 + 2 (IR generation) are implemented in #1680.

### Motivation

PyPTO's pass pipeline and memory model are currently Ascend-centric: memory-space inference targets `Vec/Mat/Left/Right/Acc`, `AllocateMemoryAddr` assigns byte addresses, and the default pass strategy runs cube/vector/cross-core passes (`AutoTileMatmulL0`, `ExpandMixedKernel`, `InjectGMPipeBuffer`, `SplitVectorKernel`, ...). None of this fits a register-file NPU.

SuperscalarNPU is the first non-Ascend backend and a forcing function to validate that the `BackendHandler` + `MemoryAllocatorPolicy` abstractions actually decouple backend behaviour from shared passes. It also introduces a genuinely different allocation model (register renaming over a fixed-size register file) that the existing byte-addressed allocator did not support.

### Design

**Memory model**
- New `MemorySpace::TREG`. SoC = single compute core carrying a 1 MB TREG (256 × 4 KB, 4 KB-aligned) + DDR; memory graph `DDR ↔ TREG`.

**Register-renaming allocation**
- Extend `MemoryAllocatorPolicy` with `AddressUnitBytes(space)` (default 1) and `MaxAddressUnits(space)` (default unbounded).
- `AllocateMemoryAddr` stores **block indices** for register-file spaces (unit > 1), reserves consecutive blocks for tiles larger than one block, and raises a user-facing error when more blocks are simultaneously live than exist (register pressure). The `AllocatedMemoryAddr` verifier scales stored addresses by the unit so the byte-based capacity check stays unit-agnostic.
[...]

PR #134, opened 2026-02-04
Addop: Submit the first PR.

wangqin1723-max:
Added new block-level operators across multiple categories:
\- Elementwise operations: Add, Sub, Mul, Div, Mod, Min, Max, Equal, etc.
\- Unary operations: Abs, Exp, Log, Sqrt, Reciprocal, etc.
\- Memory operations: Load, Store
\- Reduction operations: ReduceSum, ReduceMax, etc.
\- Transform operations: Transpose, Reshape
\- Broadcast operations

Enhanced Python language API and IR op registry with comprehensive
operator definitions. Added extensive examples and updated test suite
to improve coverage and remove legacy pipeline tests.

PR #2355, opened 2026-08-12
Advance the Simpler pin onto the Buffer device-memory API (port, not a bump)

lterrac:
> [!IMPORTANT]
> **Update 2026-08-13 — this PR is being superseded in part, and reduced on purpose.**
> #2345 does the same pin advance (it moves `runtime` to `e4ab544a`, 16 commits *ahead* of the `8a82bf9e` used here) and covers the same renames, emitters and the `tests/ut/runtime` rewrite this PR lists as a gap. So the port below is no longer the contribution.
> What survives is one measured thing: **#2345's H2D path copies, and this one does not.** Once #2345 merges, this PR will be reduced to the `copy_to` / `copy_from` in-place host-buffer naming (see *What is still worth merging* below) rather than rebased whole.
> Still a draft, but no longer because of us: #2354 reproduces on #2345's own head with our code absent, so the decode stall is a blocker for that PR too.

Advances pypto's `runtime/` (Simpler) pin onto the Buffer device-memory API and ports pypto across it.

### What is still worth merging

`DistributedWorker.copy_to` in #2345 stages every H2D through a freshly created POSIX-shm Buffer plus a full host→host `ctypes.memmove` of the payload, and `alloc_tensor` is `malloc` + `copy_to`, so all 346.6 GB of DeepSeek V4's resident weights pay an extra copy. This PR instead *names* the mapping the caller already holds — `wrap_fork_inherited(ptr, nbytes, …)` per shard, with the backend classified by the caller (`FORK_SHM` when `tensor.is_shared()`, `FORK_COW` read-only otherwise, per the capability matrix in `src/common/task_interface/buffer.h`) — and stages through shm **only** for host memory allocated after the fork, which cannot be named at all.

Measured 2026-08-13, one node, one session, single-variable: `pypto-serving` `71fd275`, `pypto-lib` `f0d352e`, PTOAS v0.57 both sides, 8 × 910B2, `--dp 8 --ep 8`, warm kernel cache, 346.6 GB resident:

| stack | resident upload | `model loaded` |
|---|---|---|
| #2345 alone | 497.7s | 518.1s |
| #2345 + #2292 | 98.9s | 119.3s |
| **this PR + #2292** | **25.6s** | **33.0s** |
| this PR alone | 66.5s | 74.0s |

**These absolutes are setup-dependent and should not be quoted as properties of either branch.** `pypto-serving`#134 measured the same class of host-copy cost at ~492s on this node and ~62s on another (page-cache state, mount, driver all matter — see that PR's discussion). What transfers is the mechanism and the direction: one extra full copy of the whole weight set, removed rather than parallelised. On a setup where the copy is cheap the gap will be much smaller.

497.7s is essentially the pre-`pypto-serving`#134 number (492.8s) *on this node*, i.e. here #2345 gives back the win of the loader change that maps the prepacked sidecar `MAP_SHARED` precisely so the upload need not copy. Cold runs agree independently: isolating the upload as `kernel-cache STORED` → `resident main weights uploaded` (both stacks pay the same ~150s of kernel compile) gives ~484.6s on #2345 vs ~20.6s here, with `model loaded` 731.3s vs 263.4s. Reported on #2345 with the code pointers.

### Why the pin had to move

Simpler #1702 removed the last of three serializers on the device-memory path (process-wide `_child_prov_lock`, run admission, and the GIL on the four bindings). pypto pinned `3165cc89`, which predates it, so #2292's concurrent shard upload was a **no-op** on that pin. #2345 now carries the pin advance instead; #2292 applies to its head with zero conflicts (two hunks, offsets only) and is worth 5.03× there (497.7 → 98.9s) versus 2.84× on this pin — it parallelises the extra memcpy too, which is why the copy still has to go.

| DeepSeek V4 Flash W8A8, 8 cards, 346.6 GB, warm cache | without #2292 | with #2292 |
|---|---|---|
| resident-weight upload | 66.5s (3 runs, ±0.6) | **23.4s / 23.8s** |
| `model loaded` total | 74.0s | **31.0s / 31.2s** |
| effective bandwidth | 5.2 GB/s | **14.8 GB/s** |

The old pin reaches `model loaded` in 73.7s, so the pin itself is performance-neutral — the 2.84× is #2292's, unlocked by the pin.

### What the port turned out to be

Not a bump. Seven layers, each found only by running the next step — recorded here because #2345 solves the same problems and the traps are worth having written down:

1. `L2TaskArgs` → `ChipTaskArgs` (Simpler #1681) — 74 sites in codegen and the collectives templates.
2. `Tensor` → `ChipTensor` at chip level, including every AIV kernel's `__gm__` preamble. Note the trap: `simpler.task_interface` exports **both** names — the plain `Tensor` is #1729's wire ABI and stays valid, so `Tensor.make(...)` on the Python path must *not* be renamed.
3. `AccessMode.READWRITE`, not `READ_WRITE`.
4. Host memory allocated **after** the fork cannot be named with `wrap_fork_inherited` at all; it stages through `create_buffer` (POSIX shm).
5. `make_tensor_arg` now takes `(worker, tensor)`; generated code has no worker in scope, so the submitter binds one via ContextVar.
6. `Tensor.make` is gone — a `DeviceTensor` is named from its allocating Buffer, which forces the owning chip onto the handle (equal-sized shards share the same numeric address on every chip).
7. Comm windows moved from `ctx.buffer_ptrs[name]` (addresses) to `ctx.buffers[name]` (Buffers).

### Verified

- 8-chip hardware check: shard uploads through the Buffer API, bogus free refused, write into a COW mapping refused, registry frees — 4/4.
- DeepSeek V4: 346.6 GB uploaded, 0 faults, `model loaded`, prefill produces a token.
- `tests/ut/{codegen,ir,language,jit,debug}`: 8441 passed, 2 skipped.

### Known gaps

- **#2354** (the decode stall) — now attributed to the ABI adoption itself, not to this port: it reproduces on #2345's head with our code absent, 3 runs, same `stuck_task_id=4294967357`.
- ~44 `tests/ut/runtime` tests still mock the old contract (`malloc(nbytes, worker_id) -> int`, `Tensor.make(child_memory=True)`). #2345 rewrites these, which is one reason to reduce this PR rather than rebase it.
- `execute_compiled`'s one-shot path builds task args before its worker exists, which the Buffer ABI cannot support; it now raises with an explanation. Fixing it has to decide what the two-pass DFX path does.

### DIFF
--- a/python/pypto/backend/pto_backend.py
+++ b/python/pypto/backend/pto_backend.py
@@ -497,7 +497,7 @@ def _generate_arg_unpacking(func: _ir_core.Function, *, uses_spmd: bool = False)
         assert isinstance(param.type, _ir_core.TensorType)
         c_type = param.type.dtype.to_c_type_string()
         lines.append(f"    // Unpack tensor: {param_name}")
-        lines.append(f"    __gm__ Tensor* {param_name}_tensor = reinterpret_cast<__gm__ Tensor*>(args[{i}]);")
+        lines.append(f"    __gm__ ChipTensor* {param_name}_tensor = reinterpret_cast<__gm__ ChipTensor*>(args[{i}]);")
         if param_name == "__gm_pipe_buffer" and uses_spmd:
             lines.append("    // SPMD: shard GM pipe workspace by logical block_idx to avoid overlap.")
             lines.append("    int64_t __pypto_gm_block_num = static_cast<int64_t>(__pypto_spmd_block_num);")
--- a/python/pypto/runtime/_buffer_bridge.py
+++ b/python/pypto/runtime/_buffer_bridge.py
@@ -1,0 +1,151 @@
+# Copyright (c) PyPTO Contributors.
+# This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+# CANN Open Software License Agreement Version 2.0 (the "License").
+# Please refer to the License for details. You may not use this file except in compliance with the License.
+# THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+# INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+# See LICENSE in the root of the software repository for the full text of the License.
+# -----------------------------------------------------------------------------------------------------------
+
+"""Shared plumbing for naming device memory to Simpler's Buffer-based APIs.
+
+pypto's public surface hands out integer device pointers (``DeviceTensor.data_ptr``),
+while Simpler's device-memory and task-arg APIs take ``Buffer`` objects. Both worker
+levels therefore need the same three things, and they need to agree:
+
+* a registry from ``(worker_id, ptr)`` back to the allocating ``Buffer`` — the pointer
+  alone cannot be re-wrapped, because a Buffer carries the identity a consumer keys on;
+* a POSIX-shm staging Buffer for host memory that a copy cannot name directly;
+* one conversion from :class:`DeviceTensor` to a wire ``Tensor``.
+
+L2 and L3 differ only in *how* they allocate (``malloc`` on a leaf worker vs
+``alloc_child_tensor`` on a chip worker), so only that part stays in each subclass.
+Keeping the rest here is what stops the two levels from drifting apart.
+"""
+
+from typing import Any
+
+
+class BufferBridge:
+    """Mixin: Buffer registry, shm staging, and DeviceTensor → wire ``Tensor``."""
+
+    def _init_buffer_bridge(self) -> None:
+        self._device_buffers: dict[tuple[int, int], Any] = {}
+        self._staging_buffers: dict[tuple[int, int], Any] = {}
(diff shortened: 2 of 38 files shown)


lterrac:
Closing: superseded by #2273, which landed the Buffer-ABI adaptation and the pin advance together. `main` now pins Simpler `1f27a157`, 44 commits past simpler#1702, so the blocker this PR existed to remove is gone and #2292 is no longer a no-op on `main` (rebased there now).

One piece of this PR does survive, and it is not carried by #2273: `copy_to` / `copy_from` still stage every device copy through a POSIX-shm Buffer plus a full host→host `memmove`. That is deliberate on `main` — it is what lets an ordinary post-fork tensor be a legal endpoint — but a range registered through `inherited_host_tensors` can be named in place instead, and `alloc_tensor` being `malloc` + `copy_to` means the copy is paid once per resident weight (346.6 GB for DeepSeek V4 Flash W8A8). It is written against `main` as a fast path with the staging kept as the fallback, so the relaxed contract is unchanged; I will open it once it is re-measured on `main` rather than quote the pre-#2273 numbers.

#2354 stays open and is not this PR's: the decode stall reproduces on two independently written ports of this ABI with the same `stuck_task_id`, so it belongs to the ABI adoption, not to either port. Whether it is present on `main` is measuring now.

PR #2363, opened 2026-08-13
fix(language): forward pad_value through pl.slice

Hzfengsy:
## Summary

`pl.tensor.slice` and `pl.tile.slice` both accept `pad_value`, but the unified `pl.slice` dispatcher never declared it, so the parameter was unreachable through the facade on either path — `pl.slice(x, shape, offset, pad_value=0)` raised `TypeError: slice() got an unexpected keyword argument 'pad_value'`, and callers had to drop to the level-specific spelling to set a padding mode. `pl.slice` now declares and forwards it.

This needs no cross-path guard. Unlike `clamp`, which only a Tensor can honour, `pad_value` exists at both levels, so forwarding is the whole fix.

Forwarding alone would have degraded a diagnostic, so the second half of the change repairs that — see below. Callers that omit `pad_value` are unaffected, and the Tensor-only `clamp` rejection on the Tile path is unchanged.

## Changes

- `python/pypto/language/op/unified_ops.py` — `pl.slice` declares `pad_value` and forwards it on both dispatch branches, in the same positional slot both callees already use: after `drop_dims`, before the Tensor-only `clamp`. An AST sweep of `tests/`, `examples/`, `python/`, and `docs/` found all 253 existing `pl.slice` call sites pass exactly three positional arguments, and the three that use `clamp` spell it as a keyword, so moving `clamp` one slot rebinds nothing.

- `python/pypto/ir/utils.py` — new `caller_warning_stacklevel()`, which walks out to the first stack frame outside the `pypto` package.

- `python/pypto/language/op/tensor_ops.py`, `python/pypto/language/op/tile_ops.py` — the two `slice` "pad_value has no effect" warnings take their `stacklevel` from that helper instead of a literal `2`.

  A literal `2` names the *immediate* caller, which was always user code before — and is the dispatcher once `pl.slice` forwards. The consequence is worse than a misleading filename: Python's default filter dedupes on `(text, category, lineno)` recorded in the frame named by `stacklevel`, so a fixed library line collapses every distinct call site onto one record. Measured on the unfixed branch, three distinct user call sites produced **1** warning instead of 3 — a kernel with twenty misused `pl.slice` calls would have seen one warning pointing into `unified_ops.py` and nineteen silent drops. Walking to the first non-`pypto` frame restores one warning per real call site and stays correct through any depth of forwarding.

- `tests/ut/language/test_unified_ops.py` — `TestUnifiedSlicePadValue` (15 tests): forwarding on both paths, the `0`/`inf` literal sugars, positional binding, `pad_value` alongside `clamp`, `None`/omitted staying byte-identical to the previous IR, and both DSL-parser paths (the parser reaches the wrapper dynamically, so kwarg resolution is a separate risk from the direct call).

  Every forwarding test pairs its structural match against the explicit `pl.tensor.*` / `pl.tile.*` call with a negative assertion that the no-`pad_value` IR differs, so the test fails rather than passing vacuously if both sides drop the kwarg. Two tests cover the warning defect specifically — attribution and per-call-site dedup — because `pytest.warns` checks neither `filename` nor `lineno` and installs `simplefilter("always")`, which defeats the dedup it needs to observe. Both were confirmed to fail against a reverted `stacklevel=2`.

## Verification

Run against the pushed commit `12e1cfcd`, rebased onto `main`:

- `ruff check` (5 changed files): exit 0, all checks passed
- `ruff format --check` (5 changed files): exit 0, already formatted
- `cmake --build build --parallel`: exit 0
- `pytest tests/ut/language/test_unified_ops.py`: exit 0, 160 passed
- `pytest tests/ut/ -n auto --maxprocesses 8`: exit 0, **9511 passed, 3 skipped** in 2:40

Separately, the two warning-regression tests were run against a reverted
`stacklevel=2` to confirm they are not vacuous: 4 failed, with the dedup test
reporting `assert 1 == 3`.

No behaviour change for existing callers, no interface removal, and no
migration step.

### DIFF
--- a/python/pypto/ir/utils.py
+++ b/python/pypto/ir/utils.py
@@ -10,9 +10,11 @@
 """Utility functions for IR construction."""
 
 import inspect
+import os
 from collections.abc import Iterator, Sequence
 from contextlib import contextmanager
 from contextvars import ContextVar
+from pathlib import Path
 
 from pypto.pypto_core import DataType
 from pypto.pypto_core import ir as _ir
@@ -200,6 +202,46 @@ def has_partial_valid_region(expr: _ir.Expr) -> bool:
     if view is None:
         view = getattr(expr_type, "tile_view", None)
     return view is not None and bool(view.valid_shape)
+
+
+# Directory holding the ``pypto`` package, with a trailing separator so the
+# match is on a path *component*. A frame whose file lives under it is library
+# code, never the call site a user-facing warning should name. Without the
+# separator a sibling like ``<parent>/pypto_kernels/k.py`` would prefix-match
+# ``<parent>/pypto`` and a real user frame would be skipped.
+_PYPTO_PACKAGE_PREFIX = f"{Path(__file__).resolve().parent.parent}{os.sep}"
+
+
+def caller_warning_stacklevel() -> int:
+    """``stacklevel`` naming the nearest frame outside the ``pypto`` package.
+
+    A literal ``stacklevel=2`` names the *immediate* caller, which is user code
+    only when the warning site is called directly. Reached through a wrapper —
+    ``pl.slice`` dispatching to ``tensor.slice``, say — that frame is library
+    code instead, and the damage is worse than a misleading filename: Python's
+    default filter dedupes on ``(text, category, lineno)`` recorded in the
+    frame at ``stacklevel``, so every user call site collapses onto one fixed
+    library line and only the first warning of many is ever shown.
+
+    Walking out to the first non-``pypto`` frame keeps one warning per real
+    call site through any depth of forwarding. Call it from the frame that
+    invokes :func:`warnings.warn`.
+
+    Returns:
+        A ``stacklevel`` for :func:`warnings.warn`, at least 1
+    """
+    frame = inspect.currentframe()
+    if frame is not None:
+        frame = frame.f_back  # the caller, i.e. the frame that will warn
+    level = 1
+    while frame is not None:
+        if not frame.f_code.co_filename.startswith(_PYPTO_PACKAGE_PREFIX):
+            return level
+        frame = frame.f_back
+        level += 1
+    # Every frame is library code (e.g. the DSL parser drives the wrapper with
+    # no user frame below it). Name the outermost one rather than a bare 1.
+    return max(level - 1, 1)
 
 
 def _to_int32_scalar(value: int | _ir.Expr, span: _ir.Span) -> _ir.Expr:
--- a/python/pypto/language/op/tensor_ops.py
+++ b/python/pypto/language/op/tensor_ops.py
@@ -124,7 +124,7 @@ __all__ = [
 ]
 
 from pypto.ir.op import tensor_ops as _ir_ops
-from pypto.ir.utils import _normalize_expr, has_partial_valid_region
+from pypto.ir.utils import _normalize_expr, caller_warning_stacklevel, has_partial_valid_region
 from pypto.pypto_core import DataType
 from pypto.pypto_core import ir as _ir_core
 from pypto.pypto_core.ir import AtomicType, Expr, MemorySpace, PadValue, PtrType, TensorLayout
@@ -473,7 +473,9 @@ def slice(
             f"If you intend to narrow the valid region later via "
             f"tensor.set_validshape, you can ignore this warning; otherwise "
             f"pass valid_shape=... to tensor.slice.",
-            stacklevel=2,
+            # Not a literal 2: pl.slice forwards here, and a fixed level would
+            # name the dispatcher and collapse every call site's warning.
+            stacklevel=caller_warning_stacklevel(),
         )
 
     tensor_expr = tensor.unwrap()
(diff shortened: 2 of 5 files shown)


Hzfengsy:
Addressed in `322767ae` (folded into the original commit, force-pushed with lease).

**Fixed — `python/pypto/ir/utils.py`, library-frame detection (CodeRabbit):** the check
was a path *prefix* match, so a user file at `<parent>/pypto_kernels/k.py` prefix-matched
the package directory `<parent>/pypto` and that real user frame would have been skipped.
It now matches on a path *component* (`_PYPTO_PACKAGE_PREFIX`, with a trailing `os.sep`).

Regression test `test_sibling_directory_is_not_mistaken_for_library_code` compiles a probe
under exactly such a sibling path and asserts the helper names it. Confirmed non-vacuous:
against the unfixed prefix it fails with `assert 2 == 1`.

**Not changed — sixth-positional `clamp`:** I measured the behaviour before acting, and the
premise of both bot findings does not reproduce. Neither value is silently rebound; the
existing bool guard rejects both loudly:

```
pl.slice(x, [16,16], [0,0], None, None, True)
# TypeError: fillpad pad_value cannot be bool (True). Use pl.PadValue.zero /
#            pl.PadValue.max / pl.PadValue.min, or one of the literals 0, 0.0,
#            math.inf, -math.inf.

pl.slice(x, [16,16], [0,0], None, None, False)   # same TypeError
```

So "`False` can similarly become unintended zero padding" and "binds `True` to `pad_value`
and leaves `clamp=False`" both describe a silent failure that cannot occur.

What remains is a genuine trade-off rather than a defect, so I have left it for a
maintainer decision and the thread unresolved:

- **As-is** — `pad_value` sits in the same positional slot as in `pl.tensor.slice` and
  `pl.tile.slice`, so `pl.slice(*args)` and `pl.tensor.slice(*args)` keep meaning the same
  thing at every arity. An out-of-tree caller passing `clamp` positionally breaks loudly.
- **`pad_value` after `clamp`** — never changes an existing call's meaning, but the
  dispatcher's positional order then differs from both callees.

The suggested `isinstance(pad_value, bool)` shim does not avoid a break either: it silently
reinterprets a bool, and makes `pl.slice(...)` and `pl.tensor.slice(...)` disagree on
identical positional arguments.

An AST sweep of `tests/`, `examples/`, `python/`, and `docs/` found all 253 existing
`pl.slice` call sites pass exactly three positional arguments, and the three using `clamp`
spell it as a keyword — so nothing in this repository is affected either way.

Verification at `322767ae`: `ruff check` / `ruff format --check` clean, `cmake --build` exit 0,
`pytest tests/ut/` **9512 passed, 3 skipped**.

Hzfengsy:
### `system-tests-direct` red — pre-existing, runner-dependent, not from this PR

Recording the diagnosis so the red check is not read as a regression here.

**Root cause of the job failure.** 29 tests "fail", but 28 are cascade: the first device
fault marks the runner unusable and everything after it reports `runner is unusable after a
prior device failure`. The single real failure both times is a GEMV-acc case
(`tile_gemv_acc_1x256x64_fp32_chunks2_partial_final`, then
`tile_gemv_acc_1x256x64_fp32_chunks2_nK`):

```
aclrtSynchronizeStreamWithTimeout (AICPU) failed: 507018 ACL_ERROR_RT_AICPU_EXCEPTION
PTO2 scheduler timeout sub_class=S1:running-stalled (detail=1)
  completed=0/1 running=1 ready=0 waiting=0 orch_done=1 stuck_task_id=4294967296 stuck_core=2
recover_device_or_mark_unusable: bounded device drain failed: 507014 (force reset will follow)
```

**It tracks the runner, not the commit.**

| head | runner | `system-tests-direct` |
| --- | --- | --- |
| `1eb0a861` (main — this PR's base) | npu-8 | pass |
| `8e27c013` (**main**) | npu-4 | **fail** |
| `12e1cfcd` (this PR, v1) | npu-2 | pass |
| `322767ae` (this PR, v2) | npu-1 | **fail** |
| `322767ae` (this PR, v2) | npu-7 | **fail** |

Main fails this job too, on npu-4. Separately, main has produced both a pass and a fail for
the *same* commit (`8e27c013`, `71020585`), which on its own establishes nondeterminism
independent of any PR.

**Why it cannot be this PR.** The delta between the passing head `12e1cfcd` and the failing
head `322767ae` is a trailing path separator in a `warnings.warn` `stacklevel` helper plus
one unit test. `caller_warning_stacklevel()` is reached only from the two `slice` no-op
`pad_value` warnings, which no system test triggers — so it is never invoked during these
runs, and a Python warning stacklevel cannot raise an AICPU exception on device. The
first-failing test exercises single-row GEMV-acc, which this PR does not touch.

I re-ran the job once; it failed on a different machine with the same signature. I am
deliberately not re-running again to chase a green: a pass would only mean a favourable
runner draw and would hide what looks like a real, hardware-variant-dependent GEMV-acc
defect on main that deserves its own investigation.

lyfne123 (approved):

PR #2367, opened 2026-08-13
feat(runtime): add submit task timing slots

Shenggan:
For L3 kernels, PyPTO’s current benchmark runtime cannot avoid including launch skew or all-rank barrier time in the reported device_wall duration. This makes it difficult to measure the actual payload time.
Simpler already provides selective task timing slots. This change exposes that capability through PyPTO’s `pl.submit` and `pl.spmd_submit` APIs via `timing_slot=0..15`, allowing users to bracket only the target device tasks and read the corresponding `task_slot_<N>` trace span.

- Adds `timing_slot=<literal 0..15>` to `pl.submit` and `pl.spmd_submit`.
- Lowers the selected slot to runtime `set_task_timing_slot()`.
- Emits one device trace span per completed slot: `simpler_run.runner_run.device_wall.task_slot_<N>`.
- Documents the API in English and Chinese.
- Adds parser and orchestration codegen coverage.

See [allgather_mm.py#L366](https://github.com/Shenggan/pypto-ccfusion/blob/acf7fa3df99d454435791ecf02e652e0a89f3476/pypto-l3/allgather_mm.py#L366) and [allgather_mm.py#L435](https://github.com/Shenggan/pypto-ccfusion/blob/acf7fa3df99d454435791ecf02e652e0a89f3476/pypto-l3/allgather_mm.py#L435) to see how to use task timing slots.


docs/en/user/tasks/02-submit.md
```diff
@@ -126,7 +126,26 @@ a, tid = pl.spmd_submit(self.k1, x, core_num=8)
...
+leaving warmup and an alignment barrier untagged:
+
+```python
+_, barrier_tid = pl.submit(self.warmup_barrier, signal)
+mid, _ = pl.spmd_submit(self.stage_a, x, core_num=N, deps=[barrier_tid], timing_slot=0)
+out, _ = pl.spmd_submit(self.stage_b, mid, w, core_num=N, deps=[barrier_tid], timing_slot=0)
+```
+
+The resulting trace span is named
+`simpler_run.runner_run.device_wall.task_slot_0`. A slot is local to one device;
```
YunjiQin: **Blocking: the documented span name is the pre-#1877 spelling.**

This names the span `simpler_run.runner_run.device_wall.task_slot_<N>`, but the pinned runtime emits `chip.run.runner_run.device_wall.task_slot_<N>` — see the `_ROUNDS_TABLE_NAMES` literals in `runtime/src/common/platform/{sim,onboard}/host/c_api_shared.cpp`.

`simpler_run` is the spelling from before simpler #1877/#1893. `python/pypto/runtime/bench.py:287-292` documents both renames (`run_prepared` -> `simpler_run` -> `chip.run`) and is precisely why PyPTO resolves these names from the installed runtime's table at call time instead of hardcoding them. A user following this doc will grep for a span that does not exist.

Same line in the Chinese mirror (`docs/zh/user/tasks/02-submit.md:118`) and in the PR description.

python/pypto/language/parser/ast_parser.py
```diff
@@ -7093,6 +7101,46 @@ def _parse_submit_allow_early_resolve_kwarg(self, method_name: str, keywords: li
...
+        kw = next((k for k in keywords if k.arg == "timing_slot"), None)
+        if kw is None:
+            return None
+        if (
+            not isinstance(kw.value, ast.Constant)
+            or not isinstance(kw.value.value, int)
+            or isinstance(kw.value.value, bool)
+        ):
+            raise ParserSyntaxError(
+                f"'{method_name}' timing_slot must be an integer literal in 0..15",
```
YunjiQin: **Blocking: `timing_slot=-1` never reaches the range branch, and the message it does hit is misleading.**

In Python's AST, `-1` is `UnaryOp(USub, Constant(1))`, not `ast.Constant`. So `timing_slot=-1` fails the *literal-type* check here and reports `timing_slot must be an integer literal in 0..15` — but `-1` **is** an integer literal, so the message points the user at the wrong problem. The friendlier `must be in 0..15, got {slot}` at L7129 is only ever reachable for `slot >= 16`.

`TestTimingSlotParsing.test_invalid_slot_rejected` passes only because both messages share the `timing_slot must` prefix, so it does not distinguish the two branches.

Either fold a leading unary minus before the range check, or reword this message to `timing_slot must be a non-negative integer literal in 0..15`.

src/codegen/orchestration/orchestration_codegen.cpp
```diff
@@ -2371,6 +2371,17 @@ class OrchestrationStmtCodegen : public CodegenBase {
     }
   }
 
+  // Selective device timing (pl.submit(..., timing_slot=N)). The runtime folds
+  // every task tagged with a slot into one device-domain dispatch-to-finish span.
+  // The parser validates the user-facing literal; retain this codegen guard for
+  // hand-built or deserialized IR that bypasses the parser.
+  void EmitTaskTimingSlot(const std::string& task_var, const CallPtr& call) {
+    if (!call->HasAttr("task_timing_slot")) return;
```
YunjiQin: **The attr key should be a declared constant, not a bare literal.**

`docs/en/dev/ir/01-hierarchy.md:522-543` makes this an explicit convention: a key one layer writes and another reads is a contract, and spelling it as a bare literal at each site leaves that contract with no single rename point. `Call` / `Submit` attr keys are declared in `include/pypto/ir/expr.h` — 15 `kAttr*` constants live there today (`kAttrCoreNum`, `kAttrDevice`, `kAttrPredicate`, `kAttrDumpVars`, ...), each with a lifecycle comment naming the writer, the readers, and whether the key is ever stripped.

`"task_timing_slot"` appears bare on this line and the next. Suggest:

```cpp
/**
 * @brief Reserved attr key for the selective device-timing tag — ``pl.submit(..., timing_slot=N)``.
 *
 * Value type: ``int`` (0..15). Written by the DSL parser; read only by
 * orchestration codegen (``EmitTaskTimingSlot`` -> ``Arg::set_task_timing_slot``).
 * Never stripped — it rides the generic attrs vector through every Submit rebuild.
 */
inline constexpr const char* kAttrTaskTimingSlot = "task_timing_slot";
```

`tests/lint/check_function_attr_key_parity.py` only covers `Function` attrs, so nothing catches this automatically. The Python side (`extra_attrs["task_timing_slot"]`) is consistent with existing practice there (`"dump_vars"`, `"arg_direction_overrides"` are bare too) and needs no change.

To be clear on the carrier itself: `attrs_` is the **right** choice here. Per `01-hierarchy.md:136-145`, `kwargs_` means language-level keyword args belonging to the callee's schema, and a `Submit`'s callee is a `GlobalVar` with no kwarg schema — `timing_slot` is launch-site metadata, not a kernel parameter. And per the `core_num_` rationale at `expr.h:1123-1128`, a first-class field is reserved for values that participate in the use-def chain; a plain `int` does not. Only the key declaration is missing.

src/codegen/orchestration/orchestration_codegen.cpp
```diff
@@ -2371,6 +2371,17 @@ class OrchestrationStmtCodegen : public CodegenBase {
...
+    const int slot = call->GetAttr<int>("task_timing_slot");
+    CHECK_SPAN(slot >= 0 && slot < 16, call->span_) << "task_timing_slot must be in 0..15, got " << slot;
```
YunjiQin: **`16` is hardcoded in three places with no shared constant.**

The slot count is owned by the runtime — `NUM_TASK_TIMING_SLOTS = 16` in `runtime/src/common/platform/include/common/device_phase.h:142`. It is currently re-spelled here, in the parser (`ast_parser.py:7127`), and in the user docs.

This file already has the established pattern eight lines below:

```cpp
// Mirrors the runtime's ``MAX_TENSOR_DIMS`` — the capacity of the fixed
// ``uint32_t indices[]`` array in ``CorePredicateOperand``.
static constexpr size_t kRuntimeMaxTensorDims = 5;
```

Suggest a matching `static constexpr int kRuntimeNumTaskTimingSlots = 16;` with a comment naming `NUM_TASK_TIMING_SLOTS` as the mirror source.

YunjiQin:
**Two docs outside this diff still describe the `pl.submit` surface without `timing_slot=`.**

Posting as a plain comment since neither file is in the diff, so there is no line to anchor to:

- `docs/en/dev/language/02-manual_dependencies.md:32-33` enumerates the full `pl.submit` / `pl.spmd_submit` kwarg surface (`deps=`, `allow_early_resolve=`, `predicate=`) and does not mention `timing_slot=`. Same for the `docs/zh` mirror.
- `docs/en/user/tasks/03-tuning.md:196` has the troubleshooting row "**`predicate` / `allow_early_resolve` rejected under `pl.cluster()`**", which now also applies to `timing_slot` — this PR adds it to the same rejection list in `_reject_submit_dispatch_metadata_in_cluster`.

YunjiQin (changes requested):

YunjiQin (approved):

PR #2369, opened 2026-08-13
feat(distributed): add deferred task completion waits

MaxwellF1:
## Summary

A blocking `pld.system.wait` occupies its AIV until a remote signal is ready. If a task graph fills the available AIVs with waiters before the publisher can run, those waiters can prevent forward progress.

This PR adds `pld.system.defer_wait`. A dedicated waiter task registers `signal >= expected` and returns from the kernel, allowing the physical AIV to be reused. Its scheduler `TaskId` remains incomplete until the condition is satisfied, so continuation tasks use the existing `deps=[wait_tid]` mechanism.

This reuses the deferred-completion API already available in the pinned Simpler runtime. It does not modify Simpler, change the runtime gitlink, or alter the existing blocking `pld.system.wait`.

## Changes

- Add the `pld.system.defer_wait` DSL and IR operation.
- Add argument, type, range, and comparison validation.
- Validate deferred waiters as dedicated pure-AIV tasks and reject unsafe continuation work inside the waiter kernel.
- Lower deferred registrations through a checked adapter to Simpler's `AsyncCtx`.
- Add the hidden raw dispatch context only to generated kernels that use deferred completion.
- Reuse ordinary `TaskId` dependencies; no new dependency type or `completion_deps` API is introduced.
- Add English and Chinese documentation.
- Add unit, simulator, and onboard coverage.

## Notes

- V1 supports `WaitCmp.Ge` on an INT32 distributed counter.
- A deferred waiter does not resume after the condition is satisfied; continuation work must be placed in a dependent task.
- Deferred counters use non-negative, monotonically increasing values.

## Verification

- Build and affected unit tests passed.
- The full unit suite introduced no new failures.
- A2/A3 and A5 simulator tests passed.
- A2/A3 onboard tests passed, including AIV saturation, repeated epochs, and already-ready counter reuse.

### DIFF
--- a/include/pypto/ir/transforms/utils/scope_outline_utils.h
+++ b/include/pypto/ir/transforms/utils/scope_outline_utils.h
@@ -171,7 +171,7 @@ class DeferredWaitContractValidator {
  private:
   explicit DeferredWaitContractValidator(Span span) : span_(std::move(span)) {}
 
-  Result ValidateStmt(const StmtPtr& stmt) const {
+  [[nodiscard]] Result ValidateStmt(const StmtPtr& stmt) const {
     if (auto seq = As<SeqStmts>(stmt)) {
       Result total;
       bool registration_started = false;
@@ -209,9 +209,9 @@ class DeferredWaitContractValidator {
     if (auto loop = As<ForStmt>(stmt)) {
       CHECK_SPAN(loop->kind_ == ForKind::Sequential || loop->kind_ == ForKind::Unroll, stmt->span_)
           << "deferred waiter supports only sequential/unrolled scalar registration loops";
-      ValidateScalarExpr(loop->start_, stmt->span_, /*allow_tensor_read=*/false);
-      ValidateScalarExpr(loop->stop_, stmt->span_, /*allow_tensor_read=*/false);
-      ValidateScalarExpr(loop->step_, stmt->span_, /*allow_tensor_read=*/false);
+      static_cast<void>(ValidateScalarExpr(loop->start_, stmt->span_, /*allow_tensor_read=*/false));
+      static_cast<void>(ValidateScalarExpr(loop->stop_, stmt->span_, /*allow_tensor_read=*/false));
+      static_cast<void>(ValidateScalarExpr(loop->step_, stmt->span_, /*allow_tensor_read=*/false));
       auto body_result = ValidateStmt(loop->body_);
       CHECK_SPAN(body_result.has_deferred_wait, stmt->span_)
           << "deferred waiter loop must register at least one pld.system.defer_wait condition";
@@ -231,7 +231,7 @@ class DeferredWaitContractValidator {
       return body_result;
     }
     if (auto branch = As<IfStmt>(stmt)) {
-      ValidateScalarExpr(branch->condition_, stmt->span_, /*allow_tensor_read=*/false);
+      static_cast<void>(ValidateScalarExpr(branch->condition_, stmt->span_, /*allow_tensor_read=*/false));
       auto then_result = ValidateStmt(branch->then_body_);
       auto else_result = branch->else_body_.has_value() ? ValidateStmt(*branch->else_body_) : Result{};
       Result result;
@@ -247,7 +247,7 @@ class DeferredWaitContractValidator {
     return {};
   }
 
-  Result ValidateCall(const CallPtr& call, const Span& span) const {
+  [[nodiscard]] Result ValidateCall(const CallPtr& call, const Span& span) const {
     CHECK_SPAN(call, span) << "deferred waiter contains a non-call side-effect statement";
     if (IsOp(call, "pld.system.defer_wait")) {
       INTERNAL_CHECK_SPAN(call->args_.size() == 3, span)
@@ -255,9 +255,9 @@ class DeferredWaitContractValidator {
       auto offsets = As<MakeTuple>(call->args_[1]);
       INTERNAL_CHECK_SPAN(offsets, span) << "Internal error: pld.system.defer_wait offsets were not verified";
       for (const auto& offset : offsets->elements_) {
-        ValidateScalarExpr(offset, span, /*allow_tensor_read=*/false);
+        static_cast<void>(ValidateScalarExpr(offset, span, /*allow_tensor_read=*/false));
       }
-      ValidateScalarExpr(call->args_[2], span, /*allow_tensor_read=*/false);
+      static_cast<void>(ValidateScalarExpr(call->args_[2], span, /*allow_tensor_read=*/false));
       return Result{true, 1, true};
     }
     CHECK_SPAN(false, span) << "deferred waiter is registration-only; unexpected operation '"
@@ -275,7 +275,8 @@ class DeferredWaitContractValidator {
                                                                                 : ScalarEffect::kPure;
   }
 
-  ScalarEffect ValidateScalarExpr(const ExprPtr& expr, const Span& span, bool allow_tensor_read) const {
+  [[nodiscard]] ScalarEffect ValidateScalarExpr(const ExprPtr& expr, const Span& span,
+                                                bool allow_tensor_read) const {
     if (!expr) return ScalarEffect::kPure;
     CHECK_SPAN(!AsTensorTypeLike(expr->GetType()), span)
         << "deferred waiter scalar bookkeeping cannot produce a tensor value";
@@ -292,7 +293,7 @@ class DeferredWaitContractValidator {
       // The first tensor.read argument is the tensor source. Skip that
       // position, not every ExprPtr that happens to alias it.
       for (size_t i = 1; i < call->args_.size(); ++i) {
-        ValidateScalarExpr(call->args_[i], span, /*allow_tensor_read=*/false);
+        static_cast<void>(ValidateScalarExpr(call->args_[i], span, /*allow_tensor_read=*/false));
       }
       return ScalarEffect::kTensorRead;
     }
--- a/include/pypto/ir/transforms/utils/scope_outline_utils.h
+++ b/include/pypto/ir/transforms/utils/scope_outline_utils.h
@@ -129,6 +129,24 @@ class PostStoreAliasCollector : public IRVisitor {
 }
 
 /**
+ * @brief Per-waiter completion-condition budget.
+ *
+ * Must match the runtime's ``MAX_COMPLETIONS_PER_TASK``
+ * (``runtime/src/{arch}/runtime/{rt}/runtime/aicore_completion_mailbox_types.h``).
+ * PyPTO's C++ never includes a runtime header, so the two cannot be linked
+ * directly; the generated adapter in ``pto_backend.py`` carries a
+ * ``static_assert`` against the real constant to catch drift at kernel-compile
+ * time.
+ *
+ * Exceeding the budget on device is not a hang: the AIV writes
+ * ``ASYNC_WAIT_OVERFLOW`` into the completion slab and the scheduler surfaces it
+ * as ``sched_error_code=102`` while aborting the run. This compile-time bound
+ * exists for attribution — a span-anchored error naming the offending
+ * ``pl.at`` scope — not to avoid a worse device-side failure.
+ */
+inline constexpr int64_t kMaxDeferredConditionsPerWaiter = 64;
+
+/**
  * @brief Validate the dedicated deferred-waiter kernel body contract.
  *
  * A waiter may use scalar expressions and sequential scalar control flow to
@@ -138,7 +156,7 @@ class PostStoreAliasCollector : public IRVisitor {
  * registration upper bound is proved statically. Conditional registration
  * (the common ``if peer != rank`` form) may execute zero times; Simpler accepts
  * that as an already-complete waiter. Loops with an unknown trip count are
- * rejected because they cannot prove the runtime's 64-condition limit.
+ * rejected because they cannot prove the per-waiter condition budget.
  */
 class DeferredWaitContractValidator {
  public:
@@ -162,9 +180,9 @@ class DeferredWaitContractValidator {
     auto result = validator.ValidateStmt(body);
     INTERNAL_CHECK_SPAN(result.has_deferred_wait, span)
         << "Internal error: deferred-wait finder/validator disagreement";
-    CHECK_SPAN(result.static_max_count <= 64, span)
-        << "deferred waiter supports at most 64 conditions, but may statically register "
-        << result.static_max_count;
+    CHECK_SPAN(result.static_max_count <= kMaxDeferredConditionsPerWaiter, span)
+        << "deferred waiter supports at most " << kMaxDeferredConditionsPerWaiter
+        << " conditions, but may statically register " << result.static_max_count;
     return result;
   }
 
@@ -213,21 +231,25 @@ class DeferredWaitContractValidator {
       static_cast<void>(ValidateScalarExpr(loop->stop_, stmt->span_, /*allow_tensor_read=*/false));
       static_cast<void>(ValidateScalarExpr(loop->step_, stmt->span_, /*allow_tensor_read=*/false));
       auto body_result = ValidateStmt(loop->body_);
-      CHECK_SPAN(body_result.has_deferred_wait, stmt->span_)
-          << "deferred waiter loop must register at least one pld.system.defer_wait condition";
+      // A loop that registers nothing contributes nothing to the budget, so it
+      // needs no statically known trip count. Returning before the arithmetic
+      // below also keeps `trip_count` out of the divisor when it is unknown.
+      if (!body_result.has_deferred_wait) return body_result;
       auto trip_count = transform_utils::EvalConstTripCount(loop);
       CHECK_SPAN(trip_count > 0, stmt->span_)
           << "deferred waiter registration loop must have a statically known positive trip count "
-             "so the 64-condition limit can be proved";
+             "so the per-waiter condition budget can be proved";
       CHECK_SPAN(trip_count == 1 || body_result.safe_after_registration, stmt->span_)
           << "deferred waiter loop may execute tensor reads or continuation work after registering its "
              "first condition; loop bodies may contain only pure scalar bookkeeping, control flow, "
              "and defer_wait registrations";
       // Saturate just above the supported limit rather than multiplying
-      // unchecked: enormous constant bounds must fail the <=64 contract, not
+      // unchecked: enormous constant bounds must fail the budget contract, not
       // wrap int64_t and accidentally pass it.
       body_result.static_max_count =
-          body_result.static_max_count > 64 / trip_count ? 65 : body_result.static_max_count * trip_count;
+          body_result.static_max_count > kMaxDeferredConditionsPerWaiter / trip_count
+              ? kMaxDeferredConditionsPerWaiter + 1
+              : body_result.static_max_count * trip_count;
       return body_result;
     }
     if (auto branch = As<IfStmt>(stmt)) {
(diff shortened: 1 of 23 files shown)


python/pypto/backend/pto_backend.py
```diff
@@ -303,6 +304,104 @@ def _run_ptoas(
...
+        static_cast<uint32_t>(expected),
+        COMPLETION_ENGINE_SDMA,
+        COMPLETION_TYPE_COUNTER,
+        0,
+    };
+    if (!register_completion_condition(ctx, token) &&
+        (ctx.completion_error_code == nullptr || *ctx.completion_error_code == PTO2_ERROR_NONE)) {
+        pto2::detail::defer_error(ctx, PTO2_ERROR_ASYNC_REGISTRATION_FAILED);
+    }
+    pto2::detail::defer_flush(ctx);
```
YunjiQin: **[A1 + B5, non-blocking but recommended before merge] This block duplicates the runtime's public `save_expected_notification_counter`, and its error branch is unreachable.**

### 1. Lines 380-391 are a field-for-field copy of a public API

`pto_async_kernel_api.h` already ships this, and it is on the public-surface list in that header's top comment:

```cpp
inline __aicore__ void save_expected_notification_counter(
    AsyncCtx &ctx, volatile __gm__ void *counter_addr, uint32_t expected_value) {
    CompletionToken token{ (uint64_t)counter_addr, expected_value,
                           COMPLETION_ENGINE_SDMA, COMPLETION_TYPE_COUNTER, 0 };
    (void)register_completion_condition(ctx, token);
    pto2::detail::defer_flush(ctx);
}
```

All five token fields match. Suggest replacing 380-391 with:

```cpp
save_expected_notification_counter(
    ctx,
    reinterpret_cast<volatile __gm__ void*>(base + offset * kElementBytes),
    static_cast<uint32_t>(expected));
```

### 2. The `defer_error(PTO2_ERROR_ASYNC_REGISTRATION_FAILED)` at 387-389 is dead code

`register_completion_condition` returns false in exactly three cases, and all three are already excluded before this point:

| Reason it returns false | Already excluded by |
| --- | --- |
| `task_token.is_invalid()` | L319 `async_ctx_is_deferred` (returns or traps) |
| `completion_count` / `completion_entries` null | L339-344 `slab_is_valid` (returns or traps) |
| Overflow | It writes `PTO2_ERROR_ASYNC_WAIT_OVERFLOW` itself, so the second conjunct is false |

The `ctx.completion_error_code == nullptr` disjunct is dead for the same reason: `slab_is_valid` already required it non-null.

Letting the runtime's own `ASYNC_WAIT_OVERFLOW` stand is also more accurate — `REGISTRATION_FAILED` is what the mailbox drain layer uses (`pto_async_wait.h`). These lines fall out naturally with the replacement above.

### 3. Remaining concern: the error paths still reach into `pto2::detail` across repos

The change above only cleans up the success path. The error paths (329, 353, 365-366, 375-376) still need `pto2::detail::defer_error` / `defer_flush`, and that header states explicitly:

> Everything else lives in `pto2::detail` and is **reserved for backend adapters / internal use**.

This is not something this PR can fix on its own: no function on the public surface can write an arbitrary error code, and the flush is mandatory — without the `dcci ... CACHELINE_OUT` writeback the scheduler reads a stale cache line at `scheduler_completion.cpp:112` and the error code never reaches the host.

Worth raising with the runtime owners; two public entry points would let the adapter drop `pto2::detail` entirely:

```cpp
inline __aicore__ void report_deferred_error(AsyncCtx &ctx, int32_t error_code);
//   = defer_error + defer_flush; covers 365-366 / 375-376
inline __aicore__ void publish_async_ctx_invalid(AsyncCtx &ctx);
//   encapsulates the corrupted-token publish path (count=0 + token reset + flush);
//   covers 329 / 353
```

More broadly, ~90% of this adapter is runtime policy (AsyncCtx decoding, slab validation, token construction, error codes, flush discipline) living in a Python string literal in PyPTO. The `#error` guard only checks that the header *exists*, not its version, so a runtime ABI change produces no compile-time signal here. Moving the whole adapter into `pto_async_kernel_api.h` (exposed as e.g. `register_counter_completion_from_args(args, base, offset, expected)`) would give a clean "undeclared identifier" error against an older runtime and put the policy under runtime review. Until then, please note in a comment above the adapter which lines mirror runtime internals and which header they track.

include/pypto/ir/transforms/utils/scope_outline_utils.h
```diff
@@ -105,6 +105,233 @@ class PostStoreAliasCollector : public IRVisitor {
...
+    // deferred registration.
+    if (!ContainsDeferredWait(body)) return {};
+
+    DeferredWaitContractValidator validator(span);
+    auto result = validator.ValidateStmt(body);
+    INTERNAL_CHECK_SPAN(result.has_deferred_wait, span)
+        << "Internal error: deferred-wait finder/validator disagreement";
+    CHECK_SPAN(result.static_max_count <= 64, span)
+        << "deferred waiter supports at most 64 conditions, but may statically register "
+        << result.static_max_count;
```
YunjiQin: **[A2, non-blocking but recommended before merge] This `64` has no link to the runtime's `MAX_COMPLETIONS_PER_TASK`.**

There are now two independent 64s in the tree:

- Runtime: `inline constexpr int32_t MAX_COMPLETIONS_PER_TASK = 64;` in `aicore_completion_mailbox_types.h`
- Here: a bare literal — and PyPTO's C++ never includes any runtime header

The generated adapter (`pto_backend.py:344`) uses the real constant, so the *device* side tracks the runtime while this *compile-time* check does not. The divergence is harmful in both directions:

- Runtime raises it to 128 → PyPTO still rejects at the 65th condition (false rejection)
- Runtime lowers it to 32 → PyPTO admits up to 64, and the device aborts the whole run with `PTO2_ERROR_ASYNC_WAIT_OVERFLOW`

Suggest doing both halves:

1. Name the constant here and record where it comes from:

```cpp
// Must match the runtime's MAX_COMPLETIONS_PER_TASK
// (runtime/src/{arch}/runtime/{rt}/runtime/aicore_completion_mailbox_types.h).
// The adapter in pto_backend.py carries a static_assert anchoring the two.
static constexpr int64_t kMaxDeferredConditionsPerWaiter = 64;
```

2. Add the anchor in the adapter so a runtime change fails at kernel-compile time:

```cpp
static_assert(MAX_COMPLETIONS_PER_TASK == 64,
    "PyPTO's DeferredWaitContractValidator hard-codes 64; see scope_outline_utils.h");
```

That static_assert is currently the only mechanism that could detect the drift automatically.

include/pypto/ir/transforms/utils/scope_outline_utils.h
```diff
@@ -105,6 +105,233 @@ class PostStoreAliasCollector : public IRVisitor {
...
+    }
+    if (auto loop = As<ForStmt>(stmt)) {
+      CHECK_SPAN(loop->kind_ == ForKind::Sequential || loop->kind_ == ForKind::Unroll, stmt->span_)
+          << "deferred waiter supports only sequential/unrolled scalar registration loops";
+      static_cast<void>(ValidateScalarExpr(loop->start_, stmt->span_, /*allow_tensor_read=*/false));
+      static_cast<void>(ValidateScalarExpr(loop->stop_, stmt->span_, /*allow_tensor_read=*/false));
+      static_cast<void>(ValidateScalarExpr(loop->step_, stmt->span_, /*allow_tensor_read=*/false));
+      auto body_result = ValidateStmt(loop->body_);
+      CHECK_SPAN(body_result.has_deferred_wait, stmt->span_)
+          << "deferred waiter loop must register at least one pld.system.defer_wait condition";
```
YunjiQin: **[B3, follow-up] "Every loop must register" contradicts this class's own doc comment and the user docs.**

The class comment above says a waiter *"may use scalar expressions and sequential scalar control flow to compute/register conditions"*, and `docs/en/dev/distributed_ops.md` says *"Pure scalar bookkeeping and control flow may run between registrations"*. This assertion rejects any non-registering loop, however pure:

```python
# Fold a linear rank into mesh coordinates -- pure scalar, no memory access
coord = 0
for d in pl.range(3):
    coord = coord * mesh_dim + (my_rank // stride_of_dim[d]) % mesh_dim
pld.system.defer_wait(signal, offsets=[coord, 0], expected=epoch, cmp=pld.WaitCmp.Ge)
```

→ `deferred waiter loop must register at least one pld.system.defer_wait condition`

Note this is coupled to the `YieldStmt` issue: any meaningful pure-scalar loop necessarily carries a value (otherwise its result cannot leave the loop and `Simplify` at pass 5 deletes it before the outliner runs), so such a loop trips the yield rejection as well. Fixing that one is a prerequisite; fixing this one alone changes nothing observable.

Once yields are accepted, this assertion can simply be dropped: the entry point's `ContainsDeferredWait(body)` early return plus `INTERNAL_CHECK_SPAN(result.has_deferred_wait, span)` already guarantee the waiter registers at least once overall, and a zero-registration loop contributes `0 * trip_count = 0`, so the `<= 64` proof is unaffected.

If the restriction is intentional, the class comment and the user doc need to say so explicitly — something like "loops inside a deferred waiter must register on every iteration; pure scalar loops (e.g. address folding) are unsupported, hoist the computation out of the loop or unroll it". Right now the two cannot both stand.

include/pypto/ir/transforms/utils/scope_outline_utils.h
```diff
@@ -105,6 +105,233 @@ class PostStoreAliasCollector : public IRVisitor {
...
+      Result result;
+      result.has_deferred_wait = then_result.has_deferred_wait || else_result.has_deferred_wait;
+      result.static_max_count = std::max(then_result.static_max_count, else_result.static_max_count);
+      result.safe_after_registration =
+          then_result.safe_after_registration && else_result.safe_after_registration;
+      return result;
+    }
+    CHECK_SPAN(false, stmt ? stmt->span_ : span_)
+        << "deferred waiter body supports only scalar assignments/control flow and "
+           "pld.system.defer_wait registrations";
```
YunjiQin, line 283: **[B1, follow-up] `YieldStmt` is missing from the whitelist, so legitimate scalar bookkeeping is rejected with a misleading message.**

`ValidateStmt` enumerates six statement kinds (`SeqStmts` / `EvalStmt` / `AssignStmt` / `ReturnStmt` / `ForStmt` / `IfStmt`). `YieldStmt` is not among them, but `ConvertToSSA` (pass 4) inserts yields well before the outliner (pass 8) in all three of its documented cases (`docs/en/dev/passes/04-convert_to_ssa.md:42-44`): phi nodes for branch merges, iter_args for loop-carried values, and promotion of loop-escaping variables.

Three waiter bodies that are rejected today:

```python
# 1) Conditional expected -- there is no loop at all here
if my_rank == 0:
    expected = base_epoch
else:
    expected = base_epoch * 2
pld.system.defer_wait(signal, offsets=[0, 0], expected=expected, cmp=pld.WaitCmp.Ge)

# 2) Threshold incremented across iterations
wanted = epoch
for src in pl.range(4):
    pld.system.defer_wait(signal, offsets=[src, 0], expected=wanted, cmp=pld.WaitCmp.Ge)
    wanted = wanted + 1

# 3) Defined inside the loop, used after it (escaping-variable promotion --
#    the user cannot tell they wrote a "carry" at all)
for src in pl.range(4):
    slot = src * 2
    pld.system.defer_wait(signal, offsets=[slot, 0], expected=epoch, cmp=pld.WaitCmp.Ge)
pld.system.defer_wait(signal, offsets=[slot, 0], expected=epoch, cmp=pld.WaitCmp.Ge)
```

All three hit this catch-all and report *"deferred waiter body supports only scalar assignments/control flow and pld.system.defer_wait registrations"* — but scalar assignments and control flow is exactly what the user wrote, and nothing in the message points at the yield. It also contradicts `docs/en/dev/distributed_ops.md` ("Pure scalar bookkeeping and control flow may run between registrations").

Suggest adding a branch after the `As<IfStmt>` case and before this catch-all:

```cpp
if (auto yield = As<YieldStmt>(stmt)) {
  // SSA phi / iter_arg carries. ConvertToSSA inserts these for scalars that
  // cross a branch merge or a loop iteration; they are bookkeeping, not
  // continuation work. ValidateScalarExpr's tensor-type guard rejects a tensor
  // carry, and allow_tensor_read=false keeps a yielded value from smuggling in
  // a post-registration memory read.
  auto effect = ScalarEffect::kPure;
  for (const auto& value : yield->value_) {
    effect = MergeScalarEffects(
        effect, ValidateScalarExpr(value, stmt->span_, /*allow_tensor_read=*/false));
  }
  return Result{false, 0, effect == ScalarEffect::kPure};
}
```

`has_deferred_wait=false` / `static_max_count=0` keeps yields out of the 64-condition accounting; `allow_tensor_read=false` makes `effect` always `kPure`, which is correct since yielding a scalar after a registration is safe; a tile carry is still rejected, but by `ValidateScalarExpr`'s tensor-type guard with the accurate "cannot produce a tensor value" message.

Worth three cases in the waiter test file: phi-carried scalar (accepted), iter_arg-carried scalar (accepted), tile carried across iterations (still rejected, correct message).

If widening the accepted set is out of scope for this PR, please at least split `YieldStmt` out of the catch-all with an actionable message, and adjust the doc sentence above to match.

YunjiQin (commented):
Follow-up review focused on the deferred-completion adapter and the waiter contract validator. Nothing here is a correctness blocker — every finding is either fail-closed (over-rejection) or a maintainability concern, and the A2/A3 onboard STs passing is consistent with that.

Two items I would suggest addressing before merge, both small:

- **A1** — the adapter's registration tail duplicates the runtime's public `save_expected_notification_counter`, and the error branch inside it is unreachable.
- **A2** — the `64` in the validator is a bare literal with no link to the runtime's `MAX_COMPLETIONS_PER_TASK`, so a runtime change would diverge silently in either direction.

Two follow-up items on waiter usability (**B1**, **B3**): the validator whitelists statement kinds against post-SSA IR, so compiler-inserted `YieldStmt` nodes and non-registering scalar loops are reported as user errors with messages that do not describe what the user wrote. B3 is downstream of B1.

For reference, I verified the failure mode of exceeding 64 conditions: it is **not** a hang. `scheduler_completion.cpp:112` reads the slab's `error_code` on every task completion, CASes it into `sm_header->sched_error_code`, and sets `completed_ = true`, which aborts the run and surfaces as `sched_error_code=102` on the host. The compile-time limit is still worth keeping, but for attribution (a span-anchored error naming the `pl.at` scope) rather than for hang avoidance.

src/backend/common/pto_ops_distributed.cpp
```diff
@@ -574,6 +575,199 @@ static std::string MakeWaitCodegenPTO(const CallPtr& op, codegen::CodegenBase& c
...
+    strides = dist_type->tensor_view_->stride;
+  }
+  if (strides.empty()) {
+    CHECK_SPAN(layout != ir::TensorLayout::NZ, op->span_)
+        << "pld.system.defer_wait does not support an NZ signal TensorView; use an ND/DN INT32 signal";
+    strides = ir::tensor_view_semantics::BuildLogicalStridesFromLayout(dist_type->shape_, layout);
+  }
+  INTERNAL_CHECK_SPAN(strides.size() == rank, op->span_)
+      << "pld.system.defer_wait signal stride rank " << strides.size() << " does not match signal rank "
+      << rank;
```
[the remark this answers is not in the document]
MaxwellF1: Fixed in `f78f2f70`.

Confirmed, and slightly worse than described: `MaterializeTensorStrides` (pass 30) fills every empty stride slot before codegen, so an NZ signal always reaches lowering carrying explicit strides — the guard never fired in a production compile, not just "normally". Reproduced the wrong flattening before the fix: a `[16,16]` NZ signal emitted `row*16 + 0*1`, plain row-major.

Moved the layout gate ahead of the stride-source choice as suggested, and added two parametrized tests covering both sources. There was no NZ coverage anywhere in the suite, which is why this went unnoticed.

One note on the test: it runs under `PassContext([])`. The autouse verification fixture makes an earlier property verifier reject an NZ signal during `MaterializeTensorStrides`, which masks the guard under test; a default (unverified) compile reaches lowering, and that is exactly where the wrong address would have been registered.

python/pypto/backend/pto_backend.py
MaxwellF1: Addressed in `de747752`.

**A1** — replaced with `save_expected_notification_counter`. Verified the five token fields match the pinned runtime (`runtime` gitlink `3165cc89`), and that the `REGISTRATION_FAILED` branch was unreachable for exactly the three reasons you listed. Checked on device rather than only in the wrapper text: the generated `deferred_wait.cpp` from the a2a3 onboard run contains the helper call and zero occurrences of the old token construction or `PTO2_ERROR_ASYNC_REGISTRATION_FAILED`.

**B5** — agreed, and out of scope for this PR: both the two-entry-point version and the full relocation need changes in `simpler` plus runtime-owner sign-off, and this feature is otherwise verified. For now the adapter carries a provenance comment recording which constructs are runtime policy, which header they track, and why the `pto2::detail` calls on the error paths are deliberate (no public entry point can publish an arbitrary error code, and the writeback is mandatory for the scheduler to observe it). I will open an issue with your two proposed signatures so it is tracked rather than lost in this thread.

include/pypto/ir/transforms/utils/scope_outline_utils.h
MaxwellF1: Both halves done — `863a8a9d` (constant) and `de747752` (anchor).

Named `kMaxDeferredConditionsPerWaiter` with the runtime path recorded, and added `static_assert(MAX_COMPLETIONS_PER_TASK == 64, ...)` in the adapter. The assert is already load-bearing: it compiled in the a2a3 onboard run, which is a live confirmation that the pinned runtime still says 64.

Also folded in your correction on the failure mode. The comment previously read "so the 64-condition limit can be proved", which invites reading the bound as hang avoidance; it now records that overflow surfaces as `sched_error_code=102` and that the compile-time bound exists for attribution — a span-anchored error naming the offending `pl.at` scope.

include/pypto/ir/transforms/utils/scope_outline_utils.h
MaxwellF1: Dropped in `863a8a9d`, together with B1.

One addition: dropping the assertion alone introduces a division by zero. A non-registering loop then reaches `static_max_count > kMax / trip_count`, and `trip_count` is `0` whenever `EvalConstTripCount` cannot fold the bounds. Handled by returning early when the loop registers nothing — which also drops the static-trip-count requirement for such a loop, since it is only there to bound registrations. Your mesh-coordinate example now compiles, including with a dynamic bound.

include/pypto/ir/transforms/utils/scope_outline_utils.h
MaxwellF1: Fixed in `863a8a9d`. Reproduced all three of your cases against the previous code first — every one hit the catch-all — then confirmed each is accepted after the change.

Added the `YieldStmt` branch essentially as sketched. One simplification: `allow_tensor_read=false` makes the accumulated effect unconditionally `kPure`, so I kept the loop for its validation side effect and return the result directly rather than threading a value that cannot vary.

Four tests added: phi-carried threshold, iter_arg-carried threshold, pure-scalar loop that registers nothing, and a tensor carry.

A correction on that last one: a tensor carry never reaches the yield guard. It is refused earlier by the `AssignStmt` tensor check on the binding that would define the carry (`cannot create or update payload tensors`). `ValidateScalarExpr`'s tensor-type guard on the yield is therefore defence in depth and unreachable from the DSL today — the test asserts the message that actually fires and says so, rather than claiming coverage it does not have.

Docs updated (EN + ZH) to state that SSA carries are accepted and that a non-registering loop needs no statically known trip count.

MaxwellF1:
All review comments addressed. Pushed as four commits on top of the lint fixes:

| Commit | Covers |
| --- | --- |
| `f78f2f70` | NZ signal layout rejected from either stride source (+2 tests) |
| `863a8a9d` | B1 `YieldStmt` carries accepted, B3 assertion dropped, condition budget named (+4 tests, EN/ZH docs) |
| `de747752` | A1 delegation to `save_expected_notification_counter`, A2 `static_assert`, B5 provenance comment |
| `357f2ccb` | Lane-placement rationale for `defer_wait` |

On the nitpick from the review body: `notify` and `wait` each carry a placement rationale in `system.cpp`, so the third op in the family having none invited exactly this question. `357f2ccb` records that the lane guarantee comes from the waiter contract rather than from `set_no_duplicate()` — `ScopeOutliner` admits a registration only inside a dedicated task-level `pl.at(CORE_GROUP)` scope, and `ExpandMixedKernel` then rejects any waiter whose rolled-up affinity is CUBE or MIXED, so a waiter converts directly to a pure-AIV function and is never duplicated onto the cube lane.

Only **B5** is deliberately not resolved here; issue to follow.

Verification on this branch:

| Gate | Result |
| --- | --- |
| C++ build | pass |
| Targeted UT | 465 passed |
| `tests/ut/ir/transforms` + `codegen` + `parser` | 3959 passed (3953 before, +6 new) |
| ST `a2a3sim` | 2 passed (16-waiter saturation, already-ready reuse) |
| ST `a5sim` | 1 passed (cross-rank correctness) |
| ST `a2a3` onboard | 2 passed, exit 0 — 48-waiter saturation x 3 epochs, devices 1,3 |

The onboard run matters for A1 specifically, since that commit changes generated device code: the `deferred_wait.cpp` it compiled contains the runtime helper call and the `static_assert`, with no residual token construction and no `TWAIT`.

One pre-existing failure is unrelated to this PR: `test_declared_memref.py::TestLoopCarryRoundtrip::test_signature_memref_base_prints_as_a_string` reproduces on an older base with none of these files touched.

YunjiQin (approved):

PR #2374, opened 2026-08-14
ci: bump ptoas to v0.58

yanghaoran29:
## Summary

Bump PTOAS to **v0.58** (`toolchain/versions.env`) and add the compatibility layer the new assembler requires under the default `--pto-level=level3` (PlanMemory skipped) compile path.

### Frontend-owned workspaces (PlanMemory is skipped)

PTOAS v0.58 refuses to synthesize scratch under level3, so PyPTO now materializes explicit tmp for the A2/A3 ops that require one: `tci` / `sort32` / `tsel` / `tsels` / `tcolsum` / `tprelu` / `trowexpandadd` / **`tcvt`** (non-saturating narrowing). Default compile stays `MemoryPlanner.PYPTO` → ptoas level3; the tmps are frontend-owned because PlanMemory never runs.

### Static-valid destination views for explicit-tmp overloads

v0.58 verifies the explicit-tmp forms against a **static** valid shape, while `pto.alloc_tile` types keep `v_row/v_col` dynamic. `tile.cast` now materializes a metadata-only static-valid alias (`pto.treshape`, `tcvt_static_dst`) over the destination — same approach `tprelu` / `tcolsum` already use. Because a later `tile.slice` of that alias cannot be legalized by PTOAS (`pto.subview` on a treshape handle), the alias is registered and `tile.slice` automatically unwraps it back to the original buffer.

### `tmrgsort` format2 tmp sizing

The format2 verifier sums each source's **physical** capacity, not the view's cols — a `pto.subview` of a wide parent (e.g. two 1024-wide top-k halves sliced out of a sorted `[1,8192]` tile) is measured at the parent width and rejects the synthesized tmp. Subview sources are now materialized into their own dense buffers (`pto.textract` identity copy) before the merge.

### Generated-C++ repairs in `_ptoas_preprocess.py` (all with TODOs, removed once the pinned ISA catches up)

- **`TFILLPAD<pto::TFillPadMode::Expand|InPlace>` → `TFILLPAD_EXPAND` / `TFILLPAD_INPLACE`**: v0.58's spelling vs the pinned ISA's flat intrinsics.
- **Mask restore after narrow TCI**: PTO-ISA `TCI_b32_normal` (`validCol < 64`) leaves the vector unit in count-mode; the next `TCVT` tail reads the leftover count as a bitmask and UB-OOBs. `set_mask_norm(); set_vector_mask(-1, -1);` is inserted after narrow TCI calls.
- **Unsigned TCI → scalar overload**: PTO-ISA `f51c92f`'s tmp path dispatches b16/b32 with a *runtime* `if`, so `TMULS_IMPL<unsigned>` instantiates unconditionally and fails to compile. Unsigned `TCI<..., tmp>` calls are rerouted to the scalar no-tmp overload (correct for unsigned in both orders). This un-skips `test_arange_repro`, `test_ci_uint32_ascend`, `test_ci_uint16_ascend`.

### Other mapping changes

- `tile.fillpad_expand` → `pto.tfillpad` (was unsupported).
- Tensor scatter stays off the broken PTOISA short-CI + column-broadcast path (padded linear CI + elementwise lowering).

## Test plan

- [x] legalize / flatten / codegen UTs for tci, tsel, tsels, tcvt scratch
- [x] onboard: `test_ci.py` (13 passed, 4 skipped), `test_cmp.py`, `test_fillpad_expand.py`, `test_row_expand_add.py`, `test_sels.py`
- [x] pypto-lib onboard: `qwen3_14b` decode/prefill, `deepseek_v4_flash_mtp` decode_swa / decode_hca / decode_csa (all green)

## Known leftover issues

Not fixed here — all are PTO-ISA (pin `f51c92f`) or version skew, not missing PyPTO workspace.

1. **A2/A3 TCI explicit-tmp descending is wrong** (PTOISA `f51c92f`): computes `-(start + i)` instead of `start - i`.
   - `test_ci.py` keeps 4 skips (`PTOISA_TCI_DESCENDING_BUG`).
   - A local ISA fix exists (descending as `2S-(S+i)` on a signed view); drop the skips after it lands and the pin is bumped.
   - The **unsigned** half of this bug is worked around by the preprocess reroute above, so `PTOISA_TCI_UNSIGNED_BUG` skips are removed in this PR.

2. **PTOAS v0.58 fillpad spelling vs pinned PTO-ISA** — handled by the preprocess rewrite; delete it once the pinned ISA exposes `TFillPadMode`.

3. **PTOISA short-CI + column-broadcast still wrong on A2/A3** — tensor scatter keeps the padded linear CI lowering until the ISA sequence is fixed.


yanghaoran29:
### Changes beyond the Simpler/PTOAS dependency pins

The large file count contains two different kinds of changes: mechanical API alignment and behavioral compatibility work. They are separated below so the review surface is clearer.

#### 1. Mechanical/API renames (no semantic change intended)

- Device-orchestration ABI names:
  - `L2TaskArgs` -> `ChipTaskArgs`
  - `L0TaskArgs` -> `CoreTaskArgs`
  - `L0TaskPredicate` -> `CoreTaskPredicate`
  - chip-side `Tensor` -> `ChipTensor`
- Tensor packing helpers:
  - the compact chip-argument path now uses `make_chip_tensor_arg`
  - `device_tensor_to_tensor` -> `device_tensor_to_chip_tensor`
- DFX runtime names:
  - `CallConfig.enable_l2_swimlane` -> `CallConfig.enable_chip_swimlane`
  - `l2_swimlane_records.json` -> `chip_swimlane_records.json`
  - the PyPTO-facing `RunConfig.enable_l2_swimlane` option is intentionally retained for compatibility and is mapped to the renamed runtime field
- Mapped SubWorker arguments now use the current indexed `MappedArg` interface (`args[i]`) instead of the old `args.tensor(i)` accessor.

These names are propagated through orchestration codegen, builtin collective templates, tests/goldens, and English/Chinese documentation. In particular, `ChipTensor` names only the compact device-visible descriptor; the host dispatch descriptor remains `Tensor`.

#### 2. Required behavioral additions

**Host dispatch ABI conversion and Buffer ownership**

- Added `to_worker_task_args(worker, args, ...)` to convert the compact `ChipStorageTaskArgs` built by PyPTO into the self-describing `TaskArgs` now required by `Worker.run`.
- Host tensors are converted while the owning Worker is available, and resident device tensors are resolved to the exact live runtime `Buffer`.
- `ChipWorker` and `DistributedWorker` retain Buffer handles behind the existing PyPTO pointer-based `DeviceTensor` API. Their malloc/free/copy implementations now translate public pointers to the corresponding handles.
- L3 keeps both `(worker_id, pointer) -> Buffer` and exact `DeviceTensor -> Buffer` ownership mappings. This is necessary because a raw address alone does not carry runtime ownership and equal virtual addresses can exist on different child workers.
- One-shot distributed execution now rejects resident `DeviceTensor` inputs. Such inputs must be allocated and dispatched through the same prepared `DistributedWorker`, which is the only place where canonical Buffer identity can be preserved.

Without this bridge, the old raw-pointer calls either fail at the new API boundary or lose the allocation identity/lifetime information used by the runtime.

**Distributed/CommDomain Buffer migration**

- Generated host orchestration now builds tensor views from `CommContext.buffers[name].tensor(...)` rather than recreating descriptors from `buffer_ptrs[name]`.
- Persistent CommDomain reset uses runtime-owned zero staging Buffers and Buffer-to-Buffer copies, and releases those staging objects only after active hierarchical runs have drained.
- SubWorker conversion accepts the new `MappedArg.buffer` and integer dtype wire representation while retaining the legacy in-process fallback.

This is more than a rename: `buffers` contains owned Buffer objects, not integer pointers, so generated code and cleanup ordering must preserve those objects.

**A2/A3 PTOAS scratch operands**

- Added an optional `tmp` operand to `tile.ci` and `tile.sort32`, including IR type/memory validation, Python printer support, and PTO emission.
- `FlattenTileNdTo2D` materializes the canonical scratch tile automatically only for A2/A3 before memory references are initialized. A5 keeps the original no-`tmp` form.

PyPTO runs PTOAS level 3 after its own memory planner, so PTOAS `PlanMemory` is skipped. In A2/A3 buffer mode, `ci` and `sort32` therefore need explicit scratch storage instead of relying on PTOAS to synthesize it.

**Split-AIV lane propagation and CPU simulator adapters**

- Split `TPUSH/TPOP` PTO now receives the logical AIV subblock id from the runtime payload (extended from i32 to the i64 PTO operand).
- CPU-simulator wrappers install a runtime-backed subblock-id hook and provide the missing explicit-lane `TPUSH/TPOP` adapters. The V2C left/right store path also preserves the required interleaved row stride.

This keeps explicit split-lane behavior consistent between hardware and CPU simulation; simulator physical-core TLS is not a reliable substitute for the logical lane selected by runtime dispatch.

**PTO-ISA CPU shim compatibility**

- The legacy cache/fence macro shim is now guarded so it is emitted only when the old macros are present. Newer PTO-ISA versions provide struct/inline-function definitions instead, and redefining them causes compilation conflicts.

#### 3. Coverage

The PR adds focused TaskArgs conversion tests and updates worker memory/ownership, distributed dispatch, CommDomain, codegen, simulator, swimlane, IR round-trip, and A2/A3 scratch coverage. The broad golden/doc churn is primarily the mechanical rename propagation described in section 1.

yanghaoran29:
Closing in favor of a refreshed PTOAS v0.60 PR based on current branch content.

PR #2387, opened 2026-08-14
feat(runtime): configurable distributed worker startup timeout

lwDavid:
## Summary

- Add an optional `startup_timeout_s` parameter to `DistributedCompiledProgram.prepare()` and `DistributedWorker`, threading a startup-readiness deadline through worker construction for the forked worker hierarchy.
- `None` (default) preserves Simpler's built-in deadline; a positive finite value overrides it for legitimately slow cold starts.
- Validate the override before `Worker` construction: non-positive, infinite, and NaN values are rejected with a clear `ValueError`.
- Document the setting in the distributed execution guide (English and Chinese).

## Testing

- `tests/ut/runtime/test_distributed_worker.py`: forwarding coverage for the helper, direct-worker, and `prepare()` paths, plus rejection cases for zero/negative/inf/NaN deadlines.

### DIFF
--- a/docs/en/user/distributed/03-execution.md
+++ b/docs/en/user/distributed/03-execution.md
@@ -24,7 +24,7 @@ directly via `DistributedWorker(compiled)`, importable from
 
 | Method | Description |
 | ------ | ----------- |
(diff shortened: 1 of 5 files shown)


lwDavid:
Duplicate: both commits on this branch were already squash-merged into main via #2340 (the validation hunk and its tests are present on main verbatim). The local branch was simply never rebased after the merge.

PR #2404, opened 2026-08-18
ci,docs: run the teaching examples, and verify the manual's first two pages

lyfne123:
Follow-up to #2394, which introduced the runnable-doc-block gate for the Performance chapter. A survey of the rest of the manual turned up two gaps, both the same shape — **teaching code that nothing executes**.

## 1. Only `examples/distributed` was run by CI

| Directory | Files | Was CI-run |
| --------- | ----- | ---------- |
| `distributed` | 7 | yes |
| `beginner` / `intermediate` / `advanced` / `utils` | 20 | **no** |
| `models` | 9 | **no** |

Everything the Tutorials chapter cites sits in that unrun set, so those pages could point at code that had stopped working and nothing would say so. This is the same gap that let a silent no-write kernel sit in the docs.

**All 20 were run on the simulator before wiring them up, and all 20 pass** — so the new step locks in a state that is good today rather than importing a backlog of failures.

## 2. The two highest-traffic pages had zero verified code

`00-getting_started.md` and `02-quickstart.md` carry 24 python blocks between them and not one was checked. Three blocks on each are now runnable and asserted.

Converting the residency example surfaced a constraint worth stating outright: **a `DeviceTensor` carries no shape/dtype for the `@pl.jit` specializer**, so that pattern has to dispatch a *compiled* program rather than the jit entry —

```
ValueError: @pl.jit: missing inferred tensor metadata for parameter 'b'
```

The page said `ir.compile(MyKernel)` and then dispatched through the kernel. It now compiles once in setup and explains why.

## `examples/models` is deliberately not wired up

Re-checked against the pinned pto-isa (`f51c92f6`), which is now fetchable here:

| Example | Old result | With the pinned pto-isa |
| ------- | ---------- | ----------------------- |
| `06_paged_attention_dynamic` | FAIL (`TFillPadMode`) | **PASS** |
| `07_paged_attention_multi_config` | FAIL (`template argument 1 is invalid`) | **PASS** |
| `09_paged_attention_spmd` | FAIL | PASS on 2 of 3 runs |

So two of the three were the environment, as suspected. The third is not: it fails
intermittently with `507018` (`ACL_ERROR_RT_AICPU_EXCEPTION`) and passed on both retries.
Wiring a flaky example into a required job would make the job flaky, so `models` stays out
until that one is understood — the rest of the directory is healthy.

## Verification

- **17 runnable doc blocks across the manual, all passing** on `a2a3sim` (13 from #2394 plus 6 here).
- `--check-parity` clean: every zh block byte-identical to its en counterpart.
- All 20 teaching examples pass on the simulator.
- `check_docs_en_zh_parity`, `check_docs_nav`, `check_english_only`, `check_docs_symbol_coverage`, markdownlint (296 files).

## Still uncovered

`language/00-types`, `language/04-scopes` and `tasks/03-tuning` have several blocks each, no runnable ones, and cite no examples at all — the remaining pure gaps, left for a follow-up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -498,6 +498,21 @@ jobs:
           echo "=== examples/distributed/07_dynamic_rank_count.py -d 0,1,2,3 ==="
           python examples/distributed/07_dynamic_rank_count.py -p a2a3sim -d 0,1,2,3
 
+      # The teaching ladders. These carry the manual's beginner/intermediate/
+      # advanced/utils citations, so leaving them unrun let the pages point at
+      # code nothing checked — the gap that let a silent no-write kernel sit in
+      # the docs. They default to a2a3sim and need no arguments.
+      - name: Run the teaching examples on the simulator
+        run: |
+          source activate.sh
+          set -e
+          for ex in examples/beginner/*.py examples/intermediate/*.py \
+                    examples/advanced/*.py examples/utils/*.py; do
+            case "$ex" in */__init__.py) continue ;; esac
+            echo "=== $ex ==="
+            python "$ex"
+          done
+
       # The user manual's runnable blocks. Teaching code that is never executed
       # rots silently — a kernel can compile, run, write nothing, and still read
       # correctly on the page — so the manual itself is executed here.
--- a/docs/en/user/00-getting_started.md
+++ b/docs/en/user/00-getting_started.md
@@ -28,19 +28,47 @@ memory and returns a `DeviceTensor` handle that `CompiledProgram` accepts in
 place of a `torch.Tensor`. The runtime treats the buffer as already resident
 and skips both H2D and D2H copies for that argument.
 
+<!-- doctest: setup -->
 ```python
+import pypto.language as pl
 import torch
 from pypto import ir
 from pypto.runtime import ChipWorker, RunConfig
 
-compiled = ir.compile(MyKernel)
+ROWS, COLS = 128, 128
+PLATFORM = "__PLATFORM__"
 
-with ChipWorker(config=RunConfig(platform="a2a3sim")) as w:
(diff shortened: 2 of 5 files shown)


lyfne123:
## Re-verified against the updated pins — no rebase needed

The branch is already on `488ba5d8` (1 ahead, 0 behind), and the runtime pin it was built against (`1f27a157`) is still what main pins, so there was nothing to replay.

What *did* change is that the pinned pto-isa is now reachable. It lives on the `hw-native-sys` remote rather than the gitcode mirror this checkout defaults to:

```
git fetch hwns          # github.com/hw-native-sys/pto-isa
f51c92f6...  ->  contains cache_line_t (41 files) and TFillPadMode (8 files)
```

That unblocked the two things I had to leave unverified.

### The earlier attribution was right, with one correction

`06_paged_attention_dynamic` and `07_paged_attention_multi_config` **pass** with the pinned pto-isa — those failures were my environment, as I said.

`09_paged_attention_spmd` is different, and I had it wrong to lump it in: it fails on `507018` (`ACL_ERROR_RT_AICPU_EXCEPTION`), and passed on both retries. That is flakiness in the example, not a version gap. The PR body now says so, and `models` stays out of CI because of that one rather than because of the pto-isa.

### Everything this PR gates was re-run under the pins CI uses

| Check | Result |
| ----- | ------ |
| The 20 teaching examples being wired up | **20/20 pass** |
| Runnable doc blocks across the manual | **17/17 pass** |
| `--check-parity` (zh code == en code) | clean |

The first pass was done under the older pto-isa, so this repeat matters: it confirms the new CI step is green under the toolchain CI actually runs, not just the one on this machine.

Hzfengsy (approved):

PR #2417, opened 2026-08-18
feat(ir): add LegalizeGraphBoundary (derived-scalar hoisting + boundary checks)

lyfne123:
> **Stacked on #2408 → #2414 → #2416.** A fork PR cannot target another fork branch, so this targets `main` and its diff includes the three earlier commits. **Review only the fourth commit.** Merge order: #2408, #2414, #2416, then this.

## Summary

The `host_build_graph` runtime records a Graph function's task topology on its first call and replays it afterwards. Replay patches exactly two things: boundary tensor addresses and boundary scalar values. Everything else — node count, shapes, dependency edges, block counts — is frozen into the recorded Definition.

Two classes of problem follow, and **both are silent at runtime**, which is why they are caught at compile time here rather than left to a log line.

## Step A — derived boundary scalars (the only silent wrong-answer path)

A boundary scalar is tracked by **pointer identity**: recording anchors the address of each `args.scalar(k)` slot and replay re-reads it. A value the body computes has no slot, so the runtime classifies it as static data and freezes the first call's value into the recording — with no warning on any later replay.

```python
@pl.function(graph="qwen_decoder_layer_v1")
def layer(self, cur, wq, layer_idx: pl.Scalar[pl.INDEX]):
    base = layer_idx * 5120     # derived: no argument slot, frozen at call #1
```

Step A rewrites this so the value arrives as a parameter and the arithmetic moves to each call site, where it becomes an ordinary pass-through scalar. A value is hoistable when its whole expression tree bottoms out in the Graph's own scalar parameters and constants — exactly what a call site can recompute, since it already supplies those parameters. Anything else (a task output, a tensor read, a runtime query) is rejected with a message naming the variable.

Two implementation notes for reviewers:

- Scalar arithmetic here is a `BinaryExpr` / `UnaryExpr` node, **not** a `Call`. The derivability check recurses through those two base classes, which covers all ~28 operator kinds at once; `As<T>` would not, since it matches one exact `ObjectKind` and each operator has its own.
- New parameters are **appended**, not prepended: `CoreTaskArgs` requires every tensor argument to precede every scalar one.

## Step D — boundary legality (silent fallbacks)

Every other constraint degrades to a silent non-graph fallback in a release build: the program stays correct and the feature simply does nothing. Rejected at compile time instead — an empty or over-32 tensor boundary, runtime-allocated (`Out`) tensor params, non-`In` scalar params, returning a computed value, over 1024 launched tasks, a Graph calling a Graph, a call site not supplying every parameter, and a launch carrying explicit dependencies or a dispatch predicate.

Two of these deserve a note:

- **The return check is deliberately narrow.** `return c` where `c` is an `InOut` parameter is the DSL's spelling for writing in place and lowers to an alias, so it is allowed. Only a genuinely *new* value is refused, since `rt_submit_graph` yields a valid task id solely on a cache **hit** — nothing can depend on a graph call's result. Rejecting all returns would have made the feature unusable in idiomatic PyPTO.
- **The predicate check is a wrong-answer path, not a fallback path.** A predicate on a graph launch is neither honoured nor rejected by the runtime — `graph_reset_outer_payload` silently zeroes it — so the region would run unconditionally.

## Pipeline position

After the final `Simplify`, immediately before `MaterializeRuntimeScopes`. Forced from both sides: `DeriveCallDirections` and `AutoDeriveTaskDependencies` must already have run so argument directions and cross-task edges are known, and `MaterializeRuntimeScopes` must not yet have run so no scope wrapper sits around the statements Step A moves.

`PassProperties` re-declares `CallDirectionsResolved` in `.produced` because the pass rewrites call arguments and their direction attrs, and `MaterializeRuntimeScopes` requires that property.

## Docs

New pass doc numbered 44; `44-47` renumbered to `45-48` in both languages, with every cross-reference, both index tables, the mkdocs nav and the `pass-doc-ordering` rule updated to match.

## Test plan

- New `tests/ut/ir/transforms/test_legalize_graph_boundary.py`: Step A hoisting, a pass-through scalar left alone, a non-Graph program proven structurally unchanged, and three Step D rejections asserted on their messages.
- `test_pass_manager_get_strategy_default` pins the Default strategy's pass list and was updated for the new entry.
- Full `tests/ut`: 9873 passed, 8 skipped. Plus clang-format, cpplint, ruff check/format, pyright, and every `tests/lint` script.

## Not in this PR

Derived tensor *slices* of a boundary tensor, and `tensor.create` inside a Graph body (a bare `alloc_tensors` poisons the recording outright), come in the P2 PR at the end of this stack.


lyfne123:
`system-tests-direct` is red for the same ambient device reason as #2408, not for anything in this change:

- **719 passed, 36 failed** — and all 36 report `RuntimeError: chip run lane is poisoned: finalize_native_run failed with code -100 (run_id=0 slot=0 generation=29 dispatch_id=0 run_epoch=29)`. The device faulted at generation 29 and every test scheduled after it inherited the poisoned lane.
- This run landed on **npu-8**; #2408's identical failure was on **npu-9**, at generation 35. Same signature, different host and different generation, which is what an ambient pool problem looks like rather than a deterministic regression.
- Every other job on this run passes.

On the change itself: `LegalizeGraphBoundary` returns the program unmodified unless it contains a `FunctionType::Graph` function, and no ST does — the pass short-circuits on the first scan. So the IR reaching codegen for every one of these tests is bit-identical to before.

Worth a re-run once the pool is healthy.

lyfne123:
Folded into #2416. The predicate audit there was small and is not separately testable — the widenings have no observable effect without something that consumes Graph IR, and this pass is that consumer. Both now land as one commit; nothing was dropped.

PR #2418, opened 2026-08-18
feat(ir): verify the Graph boundary contract, and widen it to Graph bodies

lyfne123:
> **Stacked on #2416.** #2408 and #2414 have merged; #2417 was folded into #2416. A fork PR cannot target another fork branch, so this targets `main` and its diff includes #2416's commit. **Review only the second commit.** Merge order: #2416, then this.

## Summary

#2417 rejects illegal graphs as it rewrites them. This adds the matching property verifier, so a later pass that reintroduces a violation is caught rather than silently emitting a program the runtime declines to record.

That safety net matters more here than for a typical property: almost every `host_build_graph` constraint degrades to a **silent** non-graph fallback in a release build. The program stays numerically correct and merely loses the speedup, which no correctness test can detect. This verifier is the automated detector, so it re-states the signature contract (boundary tensor count, no runtime-allocated outputs, scalars are inputs) and the call-site rules (only an Orchestration entry may call a Graph; no explicit deps; no dispatch predicate).

## Widening the three orchestration-body verifiers

This is the lockstep partner to the pass audit in #2416. `RuntimeScopesMaterialized`, `IterArgCarryClassified` and `OrchestrationReferencesResolved` all inspected Orchestration bodies only, so a Graph body satisfied them **vacuously** — which is exactly what would make them useless as a net for "some future pass copied `== Orchestration` and skipped Graph". All three now use `IsOrchestrationLike`.

The ordering was deliberate: widening a verifier *before* its producing pass would hard-fail every Graph program, whereas #2416's order only under-verified in the interim.

## A pre-existing hole this exposed

`OrchestrationCallTargetChecker` overrode only `VisitExpr_(CallPtr)`. `IRVisitor` dispatches `Submit` through a **separate handler that does not delegate to the Call path**, so `pl.submit(...)` callees have never been checked at all.

That was survivable while the checker only saw Orchestration bodies. It is not once Graph bodies are in scope: a Graph is launched from a manual scope precisely *as* a submit, so the undefined-callee check would have been silent for exactly the construct it is meant to protect. Both call-like kinds now route through one `CheckCallee`.

## Note on the caller check

It accepts `Opaque` alongside `Orchestration`: an entry function is Opaque until `OutlineIncoreScopes` promotes it, and the verifier also runs on IR that has not reached that point.

## Test plan

- New `tests/ut/ir/verifier/test_graph_boundary_legalized.py`: a well-formed graph produces no diagnostics (so the negative cases are not passing vacuously), plus the empty-boundary, runtime-allocated-output and nested-graph rejections, each asserted on rule name and message.
- Full `tests/ut`: 9877 passed, 8 skipped. Plus clang-format, cpplint, ruff check/format, pyright, and every `tests/lint` script.


lyfne123:
Three red checks here, none of them an assertion failure:

| Check | What actually happened |
| ----- | ---------------------- |
| `system-tests` | Timed out after 30 minutes — no test failures reported |
| `dist-system-tests` | Exit code 143 (SIGTERM, job budget) |
| `system-tests-direct` | The ambient device fault seen across this stack: `chip run lane is poisoned … generation=30`, on npu-6 |

**Decisive evidence that this is not the change:** #2421 and #2423 both *contain this commit* and both have `system-tests` and `dist-system-tests` **passing**. Same verifier code, same widened filters — so the two timeouts here are runner-specific, not caused by anything in the diff. The `system-tests-direct` fault has now appeared on npu-9 (#2408, gen 35), npu-8 (#2417, gen 29) and npu-6 (this one, gen 30): same signature, three different hosts, three different generations.

**On the risk that actually deserved checking here** — this commit adds a `Submit` callee check to `OrchestrationCallTargetChecker` that has never run before, so it could legitimately have rejected existing programs whose `pl.submit` callee is not in the Program. It does not: `unit-tests`, `codegen-tests`, `examples-tests`, `pypto-lib-model` and `system-tests-a5sim` all pass, and those run the full pipeline with verification over far more programs than the timed-out jobs. A real violation would fail there deterministically rather than as a timeout.

Worth a re-run once the device pool is healthy.

lyfne123:
Folded into #2416. Review pointed out that #2416 now *produces* `GraphBoundaryLegalized` with no verifier, and that the passes it widened to accept Graph bodies (`ClassifyIterArgCarry`, `MaterializeRuntimeScopes`, orchestration references) have verifiers that still skipped Graph — so FULL verification had a blind spot over exactly the IR #2416 starts emitting. A verifier belongs with its producing pass, so both land together. Nothing was dropped.

PR #2419, opened 2026-08-18
fix(core): stop flattening every op-deduction exception to ValueError

Hzfengsy:
## What

`OpRegistry::CreateImpl` is the single funnel through which every op `Call` is built. It appends the IR span to type-deduction failures, and did so by catching `const std::exception&` and rethrowing a fresh `ValueError`:

```cpp
try {
  result_type = deduce_type_fn(args, kwargs);
} catch (const std::exception& e) {
  std::string location = span.is_valid() ? " at " + span.to_string() : "";
  throw ValueError(std::string(e.what()) + location);   // flattens everything
}
```

`InternalError`, `TypeError` and `IndexError` all derive from `Error : std::runtime_error`, so that one handler flattened all of them.

## Why it matters

**The CHECK / INTERNAL_CHECK distinction was erased** for every `INTERNAL_CHECK*` site reachable from op type deduction. A compiler bug reached users as a value error, and the carefully ordered translator chain in `python/bindings/modules/error.cpp` never got the chance to run.

**The traceback vanished too** — the less obvious half. `InternalError` always reports `GetFullMessage()`; `ValueError` goes through `UserErrorMessage()`, which suppresses the trace unless `PTO_BACKTRACE=1`. So flattening the type silently also switched the trace off for exactly the errors where it is the primary artefact. Setting `PTO_BACKTRACE=1` would not have recovered it either: you'd get the *replacement* exception's frames, rooted at the catch block with everything below already unwound.

Before, an `INTERNAL_CHECK` inside a deduction function surfaced as:

```
ValueError: Internal error: <msg>
Check failed: false at /…/src/ir/op/testing.cpp:62 at kernel.py:12:15
```

After:

```
pypto.InternalError: Internal error: <msg>
Check failed: false at /…/src/ir/op/testing.cpp:62 at kernel.py:12:15

C++ Traceback (most recent call last):
 File "./python/bindings/modules/ir.cpp", line 745
   return OpRegistry::GetInstance().CreateUserFacing(op_name, args, span);
 File "./src/ir/op_registry.cpp", line 170
   result_type = deduce_type_fn(args, kwargs);
 File "./src/ir/op/testing.cpp", line 62
   INTERNAL_CHECK(false) << "Internal error: …";
```

The final frame is the deduction function itself — it exists only because the trace was carried over.

## How

**`Error::RethrowWithMessage`** (`include/pypto/core/error.h`) — a virtual that rethrows as the concrete type while adopting the stack trace captured at the original throw, so an intermediate frame can restate a message without paying either cost. Subclasses opt in with a one-line `PYPTO_ERROR_RETHROW_SUPPORT(X)`; `VerificationError` overrides by hand so its `diagnostics_` survive.

**`CreateImpl` splits its catch** — PyPTO exceptions keep type and trace and still get the span appended; non-PyPTO ones (a `std::bad_any_cast` from a wrong-typed kwarg) stay `ValueError` as before.

**Python passthrough** — the parser had the mirror-image trap: its broad `except Exception` handlers wrap strays as source-located `ParserError`s, which would re-hide an `InternalError` the moment it escaped C++. There are **seven** such handlers on the parse path — any one would have undone the fix. All now re-raise a shared `BUG_CLASS_EXCEPTIONS` first, while ordinary user errors are still wrapped exactly as before:

| Site | Shape |
| ---- | ----- |
| `ast_parser.py` — both op dispatchers | wraps as `InvalidOperationError` |
| `decorator.py` — both `parse_function` wrappers | wraps as `ParserSyntaxError` |
| `text_parser.py` — the exec path | wraps as `ParserSyntaxError` |
| `expr_evaluator.py` — `eval_expr` | wraps as `ParserTypeError` |
| `ast_parser.py` — `parse_attribute` | **swallows** (`except Exception: pass`) |

The last one is the worst of the set and the least obvious: because it swallows rather than wraps, an `InternalError` there was not merely re-typed but replaced outright by the unrelated `UnsupportedFeatureError: Standalone attribute access not supported` raised on the fall-through path. The last two were caught in review (thanks @chatgpt-codex-connector) and landed in 3ba3f8a5.

## For reviewers

**Two tests changed, and the reason is worth a look.** `test_tile_{put,get}_rejects_stage_larger_than_transfer` asserted `pytest.raises(ValueError)` against an `INTERNAL_CHECK_SPAN` whose own message begins `"Internal error:"`. `pld.tile.put` / `get` are materialized by `ConvertTensorToTileOps` and, per `docs/en/dev/distributed_ops.md`, never appear on the DSL surface — their staging tile is compiler-created, so an oversized stage is a pass bug and the check is classified correctly. The tests were encoding the flattening bug. Corrected to `pypto.InternalError`, keeping `match="must fit within"`.

**Blast radius is smaller than it looks.** Under `src/ir/op` the ratio is 1,389 `CHECK` to 21 `INTERNAL_CHECK`, so the common user-error path is byte-identical. The two tests above were the only fallout across the whole suite.

**Adding an `Error` subclass later?** End the class body with `PYPTO_ERROR_RETHROW_SUPPORT(YourError)`. Omitting it still compiles but degrades to a plain `Error` on rethrow — documented in `docs/en/dev/02-error-handling.md`.

**Not addressed here:** the `Check failed: <expr> at <abs cpp path>` tail visible in the samples above is a separate finding (`ast_parser.py` `_dispatch_op` not sanitizing) and is left alone deliberately.

## Testing

- Full suite green: **9,940 passed, 3 skipped, 0 failed**.
- 34 new tests: the rethrow mechanism per subclass (`tests/ut/core/test_error.py`), the registry contract (`tests/ut/ir/operators/test_op_registry.py`), and the parser passthrough (`tests/ut/language/parser/test_error_wrapping.py`), backed by two test-only ops and a `rethrow_with_message` binding.
- Verified the new tests actually discriminate rather than passing vacuously: reverting the `ast_parser.py` passthrough reproduces the exact `InvalidOperationError: … Internal error: synthetic pass bug` re-wrapping and fails three of them; reverting either half of the `expr_evaluator` / `parse_attribute` pair reproduces the compound `UnsupportedFeatureError` swallow.
- Trace-preservation assertions are chosen to be real discriminators — a file boundary in the registry test, a frame *count* in the mechanism test (both frames there share a file, so presence alone would be a tautology).
- Docs updated in `docs/en/` and `docs/zh/`.
- Review hardening in 0f5a2c89: two frame-count assertions could never hold on macOS (`_assert_has_trace` returns an empty list there after checking the no-symbolization fallback) and are now skipped; `test_unknown_span_appends_no_location` compares the unknown-span and located messages for exact equality, since `Span::unknown()` renders as `:-1:-1` and the old substring check would have accepted a dropped guard; and the `AssertionError` arm of `BUG_CLASS_EXCEPTIONS` is now covered at the parser layer.

### DIFF
--- a/docs/en/dev/02-error-handling.md
+++ b/docs/en/dev/02-error-handling.md
@@ -116,6 +116,11 @@ except Exception as e:
     raise InvalidOperationError(...) from e
 ```
 
+This covers speculative evaluations too, where the broad handler *swallows* rather than
+wraps (`try: ... except Exception: pass`, then fall through to another strategy). Those
+are the worse case: a swallowed `InternalError` is replaced by whatever unrelated error
+the fall-through path raises next.
+
 ### Platform support for stack traces
 
 `3rdparty/libbacktrace` tracks upstream [ianlancetaylor/libbacktrace](https://github.com/ianlancetaylor/libbacktrace).
--- a/docs/en/dev/02-error-handling.md
+++ b/docs/en/dev/02-error-handling.md
@@ -67,6 +67,55 @@ stricter — it rejects anything other than `0` or `1` — so stick to those two
 path — `libbacktrace`, `nanobind`, libc, the C++ standard library, and the `error.h` / `logging.h`
 throw sites (`kFileNameFilter` in `src/core/backtrace.cpp`).
 
+### Augmenting an error without flattening it
+
+An intermediate frame often wants to add context to an error already in flight — the op registry
+appends the IR span to every type-deduction failure, so a message thrown deep inside a deduction
+function still names the offending DSL line. Catching `const Error&` and constructing a fresh
+exception does that, but at two costs: the concrete type collapses to whatever the catcher throws,
+and the stack trace captured at the original throw is replaced by the catcher's own.
+
+Use `Error::RethrowWithMessage` instead. It is virtual and every subclass overrides it, so the
+exception rethrows as its own type carrying the frames of the original throw:
+
+```cpp
+try {
+  result_type = deduce_type_fn(args, kwargs);
+} catch (const Error& e) {
+  // Concrete type and original trace survive; only the message changes.
+  e.RethrowWithMessage(std::string(e.what()) + LocationSuffix(span));
+} catch (const std::exception& e) {
+  // Non-PyPTO exceptions carry no PyPTO trace to keep.
+  throw ValueError(std::string(e.what()) + LocationSuffix(span));
+}
+```
+
+Catching `const std::exception&` alone is the trap: `InternalError`, `TypeError` and `IndexError`
+all derive from `Error : std::runtime_error`, so a single handler that rethrows `ValueError`
+silently flattens every one of them — erasing the `CHECK` / `INTERNAL_CHECK` distinction for
+everything below the catch, and defeating the ordered translator chain in
+`python/bindings/modules/error.cpp` before it can run.
+
+**Adding a new `Error` subclass?** End the class body with `PYPTO_ERROR_RETHROW_SUPPORT(YourError)`,
+which defines the trace-adopting constructor and the override. A subclass that omits it still
+compiles, but rethrows as a plain `Error`. A subclass carrying extra state needs a hand-written
+override so that state survives too — `VerificationError` is the worked example.
+
+The Python parser has the mirror-image trap. Its handlers wrap stray exceptions as source-located
+`ParserError`s, which would re-hide an `InternalError` the moment it escaped C++. Every broad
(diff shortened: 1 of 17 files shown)


Hzfengsy:
Also took the nitpick from the same review — `BUG_CLASS_EXCEPTIONS` has two members but only `InternalError` was covered at the parser layer, so `test_assertion_error_from_op_is_not_wrapped` now exercises the `AssertionError` arm. Landed in 0f5a2c89 alongside the two inline fixes (no thread to reply on for that one).

Full suite after all three: 9,940 passed, 3 skipped, 0 failed.

lyfne123 (approved):

PR #2422, opened 2026-08-18
perf(runtime): name fork-inherited host memory instead of staging it

lterrac:
## What

`copy_to` / `copy_from` stage every device copy through a simpler-owned POSIX-shm Buffer plus a full host→host `memmove` of the payload. `alloc_tensor` is `malloc` + `copy_to`, so a resident weight set is copied once on the host before any byte reaches a chip — 346.6 GB of it for DeepSeek V4 Flash W8A8.

The staging is deliberate and worth keeping: it is what lets an ordinary post-fork tensor be a legal endpoint (`_require_copy_host_tensor`). But a range registered through `inherited_host_tensors` does not need it — that memory predates the fork, so every child already holds it at the same virtual address and `simpler.buffer.wrap_fork_inherited` can name it in place. So this is a fast path, not a change of contract:

| range | naming | staged instead |
|---|---|---|
| fork-inherited, `MAP_SHARED` | `FORK_SHM`, `READWRITE`, either direction | — |
| fork-inherited, `MAP_PRIVATE` | never | always — see below |
| anything else | — | always, exactly as today |

**Correction after review (thanks @codex-reviewer).** The first revision of this PR also named `MAP_PRIVATE` ranges as `FORK_COW` for uploads. That was wrong, and not just conservatively so: copy-on-write leaves the child reading its pre-fork snapshot, so a parent that registers a tensor through `inherited_host_tensors` and then writes to it would have uploaded the *old* bytes, silently. Staging is not a fallback for private ranges — it is the correct behaviour, because its `memmove` runs in the parent and therefore reads the current contents. The fast path is now restricted to `MAP_SHARED`, and `FORK_COW` is not used at all.

The backend kind is a classification the caller must make rather than something inferred from the access mode, so it is recorded at registration time, where the tensors are still in hand, as `(start, end, is_shared)` spans that a bare address is resolved against per copy. Ranges are wrapped individually rather than offset into a whole-tensor Buffer because a Buffer carries no offset: a shard's address is interior to its stacked tensor, so per-range wrapping is what keeps one shard's copy from moving the whole stack. The spans are dropped in `release_inherited_host_tensor_refs` and in `close`, so a copy after the parent releases its references stages instead of naming memory nobody vouches for.

> [!WARNING]
> **The numbers below do not reproduce on the current revision.** Restricting the fast path to `MAP_SHARED` also switched off the DeepSeek case: the prepacked sidecar is `MAP_SHARED` at the OS level but `torch.is_shared()` reports `False` for an `mmap` + `from_numpy` storage, so those ranges are now staged rather than named. `is_shared()` is the wrong predicate, and the fix needs an API decision — see [this comment](https://github.com/hw-native-sys/pypto/pull/2422#issuecomment-5342087994). Treat the table as what the mechanism is worth once the predicate is right, not as what this diff currently does.

## Measured

Current `main` (`49ed7797` + Simpler `1f27a157`), DeepSeek V4 Flash W8A8, 8 × 910B2, 346.6 GB resident, warm kernel cache, `pypto-serving` `71fd275`, `pypto-lib` `f0d352e`, PTOAS v0.57. #2292 is applied on both sides, so the shard upload is concurrent either way and this change is the only variable; the variant is swapped in site-packages between runs, so the kernel cache is warm and identical.

| | resident upload | `model loaded` |
|---|---|---|
| `main` + #2292 | 102.8s | 123.3s |
| + this change | **31.7s** | **39.1s** |
| | 3.24× | 3.15× |

A cold run of the same configuration puts the upload window (`kernel-cache STORED` → `resident main weights uploaded`) at 19.3s; the 31.7s above was taken while other jobs shared the node, so it is the conservative number of the two.

**Please read the absolutes as setup-dependent, not as a property of this change.** `pypto-serving`#134 measured the same class of host-copy cost at ~492s on one node and ~62s on another, with page-cache state, mount and driver all mattering. What transfers is the mechanism: one full copy of the payload, removed rather than parallelised. On a setup where the copy is cheap the win is smaller, but it should not be negative anywhere.

## Not fixed here

DeepSeek V4 decode stalls after the first token on `main` — see #2354. It reproduces with and without this change, and with and without #2292, so it is independent of both; it is mentioned only because it is visible in every run above (`req 0: TRUNCATED 1/32`). Startup, upload and prefill are healthy.

## Testing

`pre-commit` clean (including `pyright`). Exercised end to end by the runs above: the fast path is what carries all 346.6 GB, and staging still carries every post-fork copy on the same runs.

`TestNamedInheritedHostRanges` in `tests/ut/runtime/test_distributed_worker.py` covers the span-resolution boundary: a contained shared range is named (`FORK_SHM` + `READWRITE`, no staging buffer created), a private range is staged **and the staged bytes are the parent's current contents** — that one is the regression guard for the correction above — plus partial overlap, the D2H direction, staging resuming after `release_inherited_host_tensor_refs()`, and distinct buffer identities per call.

The seconds above were measured on the first revision, which named those ranges. I had reasoned that the current revision would be unaffected because the sidecar is `MAP_SHARED`; that reasoning was wrong, for the `torch.is_shared()` reason in the warning above. It is a good illustration of why the caveat was worth stating: the re-measurement I offered is what would have caught it, and it still needs to happen once the predicate is fixed.

### DIFF
--- a/python/pypto/runtime/distributed_runner.py
+++ b/python/pypto/runtime/distributed_runner.py
@@ -1525,6 +1525,18 @@ class DistributedWorker(Worker):
         )
         self._buffer_owner_id: bytes | None = None
         self._buffer_id_seq = 0
+        # One identity per host *range*, not per copy. Both properties this buys are load
+        # bearing: a consumer's ImportRegistry only drops an entry when the owner releases the
+        # Buffer, which a named copy never does, so minting per copy would leave one permanent
+        # ImportedBuffer per copy in every chip child — unbounded for a per-step D2H read-back.
+        # And re-copying the same range must reuse its identity, because one identity may name
+        # only one backing: `materialize` refuses a second descriptor for an identity it has
+        # already handed out.
+        self._named_identities: dict[tuple[int, int], tuple[bytes, int]] = {}
+        # Minting is not atomic (`+=` is load/add/store, and the lazy owner mint is
+        # check-then-act) while `alloc_stacked_tensor` runs one thread per chip through this
+        # path, so the cache and the counter are guarded together.
+        self._named_identity_mu = threading.Lock()
         self._persistent = bool(persistent)
         self._reset_persistent_windows = reset_persistent_windows
         self._persistent_error: BaseException | None = None
@@ -2179,18 +2191,36 @@ class DistributedWorker(Worker):
             return 0
         return int(self._w.committed_device_memory(worker_id))
 
-    def _next_buffer_identity(self) -> tuple[bytes, int]:
-        """Mint the ``(owner_instance_id, buffer_id)`` pair a named host range needs."""
+    def _buffer_identity_for(self, host_ptr: int, nbytes: int) -> tuple[bytes, int]:
+        """Return the stable ``(owner_instance_id, buffer_id)`` naming this host range.
+
+        Keyed on the range rather than counted per call, so a range copied N times keeps one
+        identity: the consumer's ``ImportRegistry`` refuses a second descriptor for an identity
+        it has already materialized, and it only drops an entry when the owner releases the
+        Buffer — which the named path never does, since it hands out the caller's own mapping.
+        Per-copy identities would therefore both collide on a re-copy and grow a child's
+        registry without bound. Distinct sub-ranges of one registered tensor still get distinct
+        identities, which is what a sharded upload needs.
+
+        Held under a lock because ``alloc_stacked_tensor`` drives this from one thread per chip.
+        """
         from simpler.buffer import (  # noqa: PLC0415  # pyright: ignore[reportMissingImports]
             mint_owner_instance_id,
         )
 
-        owner = self._buffer_owner_id
-        if owner is None:
-            owner = mint_owner_instance_id()
-            self._buffer_owner_id = owner
-        self._buffer_id_seq += 1
-        return owner, self._buffer_id_seq
+        key = (int(host_ptr), int(nbytes))
+        with self._named_identity_mu:
+            cached = self._named_identities.get(key)
+            if cached is not None:
+                return cached
+            owner = self._buffer_owner_id
+            if owner is None:
+                owner = mint_owner_instance_id()
+                self._buffer_owner_id = owner
+            self._buffer_id_seq += 1
+            identity = (owner, self._buffer_id_seq)
+            self._named_identities[key] = identity
+            return identity
 
     def _named_host_buffer(self, host_ptr: int, nbytes: int) -> Any:
         """Name a fork-inherited MAP_SHARED host range in place, or ``None`` to stage it.
@@ -2222,7 +2252,7 @@ class DistributedWorker(Worker):
         for start, stop, is_shared in self._inherited_host_spans:
             if host_ptr < start or end > stop or not is_shared:
                 continue
-            owner, buffer_id = self._next_buffer_identity()
+            owner, buffer_id = self._buffer_identity_for(host_ptr, nbytes)
             return wrap_fork_inherited(
                 host_ptr,
                 nbytes,
@@ -2394,6 +2424,8 @@ class DistributedWorker(Worker):
         # No later copy may name these ranges: the parent has dropped its references, so a
         # copy after this point stages rather than wrapping memory nobody vouches for.
         self._inherited_host_spans = ()
+        with self._named_identity_mu:
+            self._named_identities.clear()
 
     def copy_stacked_from(self, stacked: StackedDeviceTensor, host: torch.Tensor) -> None:
         """Read every shard of *stacked* back to *host* (D2H) — the read-back
@@ -2708,6 +2740,7 @@ class DistributedWorker(Worker):
             finally:
                 self._inherited_host_tensors = ()
                 self._inherited_host_spans = ()
+                self._named_identities.clear()
                 self._persistent_domains_by_program.clear()
                 if self._close_complete:
                     self._device_buffers.clear()
--- a/python/pypto/runtime/distributed_runner.py
+++ b/python/pypto/runtime/distributed_runner.py
@@ -1510,12 +1510,11 @@ class DistributedWorker(Worker):
         self._inherited_host_tensors = inherited
         # `copy_to` / `copy_from` stage through a simpler-owned shm Buffer so an ordinary
         # post-fork tensor is a legal endpoint. That relaxation costs one full copy of the
-        # payload, which a fork-inherited range does not need: it is present in every child
-        # at the same virtual address and can be named in place. Record the classification
-        # here, where the tensors are still in hand — MAP_SHARED (`FORK_SHM`) may serve as
-        # an output, MAP_PRIVATE (`FORK_COW`) is read-only because a child's write would
-        # split the page into a copy the parent never sees — as (start, end, is_shared)
-        # spans that a bare address can be resolved against later.
+        # payload, which a fork-inherited MAP_SHARED range does not need: parent and child
+        # see the same pages, so it can be named in place. Whether a range is MAP_SHARED is
+        # the caller's to know and cannot be inferred from an address, so record it here,
+        # where the tensors are still in hand, as (start, end, is_shared) spans. A
+        # MAP_PRIVATE range is deliberately left to staging — see `_named_host_buffer`.
         self._inherited_host_spans: tuple[tuple[int, int, bool], ...] = tuple(
             (
                 tensor.data_ptr(),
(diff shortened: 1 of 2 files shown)


python/pypto/runtime/distributed_runner.py
```diff
@@ -2162,12 +2180,61 @@ def committed_device_memory(self, worker_id: int = 0) -> int:
...
+                # so fall through rather than refusing the copy.
+                return None
+            owner, buffer_id = self._next_buffer_identity()
+            return wrap_fork_inherited(
+                host_ptr,
+                nbytes,
+                owner,
+                buffer_id,
+                access=AccessMode.READWRITE if is_shared else AccessMode.READ,
+                backend_kind=BackendKind.FORK_SHM if is_shared else BackendKind.FORK_COW,
```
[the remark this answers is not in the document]
lterrac: You're right, and this is the case the change got wrong. Copy-on-write freezes the child's view at fork, while the staging `memmove` runs in the parent and therefore reads the current contents — so `torch.zeros(...)` registered through `inherited_host_tensors` and then `fill_(1)` would have uploaded zeros, silently. Not a corner I had covered: nothing in the API promises post-fork immutability for those entries, and I was not proposing to add that promise.

Fixed in ad89d408 by restricting the fast path to `MAP_SHARED` ranges, where parent and child see the same pages and naming changes nothing observable. `FORK_COW` is no longer used on this path at all, and `_named_host_buffer` lost its `writing` parameter with it — a private range now stages unconditionally, in both directions.

The measured win survives, which is what made the narrower rule easy: DeepSeek V4's 346.6 GB of resident weights arrive through a `MAP_SHARED` prepacked sidecar (`pypto-serving`#134 mapped it that way precisely so the upload would not copy), so it stays on the named path.

There is now a regression guard for exactly your scenario — `test_copy_to_stages_a_private_inherited_range` mutates the tensor after registration and asserts the staged bytes equal the parent's current contents.

lterrac:
@coderabbitai thanks — one taken, one declined, both in ad89d408.

**Span-boundary and released-reference tests: added.** `TestNamedInheritedHostRanges` in `tests/ut/runtime/test_distributed_worker.py` covers a fully contained shared range (named, `FORK_SHM` + `READWRITE`, no staging buffer created), a private inherited range (stages — see below), a range starting inside a span but running past its end (stages, because a Buffer must not vouch for more than the mapping it names), D2H on a shared range, staging resuming after `release_inherited_host_tensor_refs()`, and distinct buffer ids across two live names. 167 tests in that file pass.

The private-range case changed meaning while I was writing those tests, from Codex's P1 on the same lines: naming a `MAP_PRIVATE` range would have uploaded the child's pre-fork snapshot, so the fast path is now restricted to `MAP_SHARED` and there is no COW branch left to test.

**Per-file `PLC0415` ignore: declining, with numbers.** `python/pypto/runtime/distributed_runner.py` already carries **21** inline `# noqa: PLC0415` markers — mine are 2 of them — so inline suppression is this file's established convention, not an omission. Meanwhile `[tool.ruff.lint.per-file-ignores]` is used only for `E501`/`F841` on DSL test modules. Adding a file-wide ignore would silence the rule across 2,700 lines, losing the signal on a future accidental lazy import, and being consistent about it would mean deleting 21 markers this PR does not own. If maintainers would rather standardise on the per-file form, that is worth doing as its own change across the runtime package.

lterrac:
**Correction: as it stands this PR no longer buys the number in its own description, and the fix needs an API decision. Details below — please read before reviewing.**

Restricting the fast path to `MAP_SHARED` was the right call for the staleness bug, but it also switched off the case the PR exists for. Measured just now:

```
mmap(MAP_SHARED) + np.frombuffer + torch.from_numpy : is_shared() = False
torch.zeros(...).share_memory_()                     : is_shared() = True
torch.from_file(path, shared=True, size=...)         : is_shared() = True
plain tensor                                         : is_shared() = False
```

DeepSeek's 346.6 GB arrive through `pypto-serving`'s prepacked sidecar, which maps the file with `mmap.mmap(fd, 0, ACCESS_READ)` — genuinely `MAP_SHARED` at the OS level — and wraps it via `np.frombuffer` + `torch.from_numpy`. `torch.is_shared()` reports **False** for that storage, so with this revision those ranges are staged, not named, and the 102.8s → 31.7s in the description does not reproduce. The earlier revision did name them, which is why the measurement was both fast and correct: the mapping really is shared, so naming was safe — my classifier simply could not tell it apart from a private one.

So `is_shared()` is the wrong predicate. And on reflection the right predicate is not "shared vs private" either: what makes naming safe is that **nobody writes the range after the fork**. A `MAP_SHARED` range is safe unconditionally; an immutable `MAP_PRIVATE` range is equally safe, and a mutated one is unsafe in either mapping. Only the caller knows which it registered — and it does know: `pypto-serving`'s own helper is called `_inherited_host_weights` and documents its return as *"immutable main and MTP weights that must be visible at worker fork"*.

That suggests making the promise explicit rather than inferred, and I would rather agree the shape with you than push a guess:

1. a separate argument — `nameable_host_tensors=[...]` alongside `inherited_host_tensors`, documented as "the caller guarantees these are not written after `prepare()`; they are named in place instead of copied"; or
2. keep one argument and let entries carry the promise, e.g. `inherited_host_tensors=[(tensor, HostBackend.SHARED), ...]`; or
3. no new API: treat registration through `inherited_host_tensors` as already implying immutability — that is arguably what the parameter has always meant for uploads — and name every registered range, with `MAP_SHARED` detected where possible. This is the smallest diff and restores the number, but it makes the existing contract load-bearing in a way it currently is not enforced to be.

My preference is (1): it is explicit, it defaults to today's behaviour, and it keeps "the caller classifies the backend" — which is the rule the Buffer API already states — instead of inferring it. Happy to implement whichever you prefer.

Until that is settled I have left the code in its correct-but-slow state rather than restore the speedup on a predicate I now know is wrong. The description's numbers are marked accordingly.

python/pypto/runtime/distributed_runner.py
```diff
@@ -2162,12 +2179,70 @@ def committed_device_memory(self, worker_id: int = 0) -> int:
...
+        """Mint the ``(owner_instance_id, buffer_id)`` pair a named host range needs."""
+        from simpler.buffer import (  # noqa: PLC0415  # pyright: ignore[reportMissingImports]
+            mint_owner_instance_id,
+        )
+
+        owner = self._buffer_owner_id
+        if owner is None:
+            owner = mint_owner_instance_id()
+            self._buffer_owner_id = owner
+        self._buffer_id_seq += 1
```
YunjiQin: **Identity minting races the concurrent shard upload.**

`self._buffer_id_seq += 1` is not atomic in CPython (LOAD_ATTR / ADD / STORE_ATTR are separately interruptible), and the lazy `_buffer_owner_id` mint above is check-then-act.

This path is concurrent in exactly the configuration the PR measures. `alloc_stacked_tensor` uploads shards through `ThreadPoolExecutor(max_workers=len(ids))`, and each `_upload_shard` runs `alloc_tensor(init=host[i].contiguous())` → `copy_to` → `_named_host_buffer`. For a contiguous host tensor `host[i].contiguous()` is a view, so its address is interior to a registered span and the fast path engages — 8 chips means 8 threads minting identities against this counter. #2292 is applied on both sides of the measurement, so this is not a hypothetical arrangement.

The failure is a hard one rather than a silent one: two `Buffer`s sharing an identity but carrying different bodies make the child raise `ImportRegistry: identity ... is already materialized from a different descriptor` (`simpler/buffer.py:605-616`), i.e. the upload fails.

Identity allocation — both the sequence increment and the first-time owner mint — needs to be atomic with respect to concurrent `copy_to` / `copy_from`. `test_named_ranges_get_distinct_buffer_ids` asserts precisely the invariant that breaks here, but it is single-threaded and cannot observe it; a concurrent case is needed to guard this.

python/pypto/runtime/distributed_runner.py
```diff
@@ -2162,12 +2179,70 @@ def committed_device_memory(self, worker_id: int = 0) -> int:
...
+
     def copy_to(self, dst_dev_ptr: int, src_host_ptr: int, nbytes: int, *, worker_id: int = 0) -> None:
         """H2D copy: ``nbytes`` from host *src_host_ptr* to device *dst_dev_ptr*."""
         self._require_open("copy_to")
         dst = self._device_buffer(dst_dev_ptr, worker_id, "copy_to")
         if not isinstance(nbytes, int) or nbytes <= 0:
             raise ValueError(f"nbytes must be a positive int, got {nbytes!r}")
+        src = self._named_host_buffer(int(src_host_ptr), int(nbytes))
+        if src is not None:
+            self._w.copy_to(dst, src)
```
YunjiQin, line 2274: **A named Buffer is never released, so each child's `ImportRegistry` grows without bound.**

The staging path releases its Buffer in a `finally`; this path returns straight after the copy, and every named copy mints a fresh `buffer_id`. On the consumer side `ImportRegistry._by_identity` only drops an entry in `unregister()`, and the sole broadcaster of that is the owner's `release_buffer()` — `Worker.release_buffer` documents this explicitly ("tell every descendant to drop its own cached import"), and `unregister`'s own docstring calls it out as what keeps a long-lived endpoint from holding every backing it ever saw (`simpler/buffer.py:703-713`).

So each named copy leaves one permanent `ImportedBuffer` (descriptor included) in **every** chip child. During the resident upload that is one per tensor. The D2H direction is the sharper case: `copy_stacked_from` into an inherited shared host tensor — reading a resident KV cache back each step is routine in serving — leaks entries per step for the life of the process. Same applies to the `copy_from` early return at line 2261.

The number of live entries in a child's registry needs to be bounded by the number of distinct named ranges, not by the number of copies performed.

Worth noting that `test_named_ranges_get_distinct_buffer_ids` currently pins the behaviour that produces the growth. The invariant that actually matters is that *concurrently live* distinct ranges do not collide on one identity; the test should be restated in those terms.

YunjiQin (commented):
The mechanism is sound and the `MAP_PRIVATE` correction is the right call — `ImportRegistry` really does resolve `FORK_SHM` as `base = int.from_bytes(desc.body)` with no mapping and no copy, so the zero-copy premise holds, and per-range wrapping is the correct answer to `Buffer` carrying no offset.

Two issues I'd want resolved before this lands. Both are on the named path only; staging is unaffected.

python/pypto/runtime/distributed_runner.py
lterrac: Confirmed and fixed in b6789a9a. Both of your findings turned out to be the same mistake — minting an identity per *copy* rather than per *range* — so they are fixed together.

On this one specifically: you are right that the mint is not atomic and that the measured configuration is exactly the concurrent one. Worth adding that our own measurement went through this path with 8 threads and did *not* fail, so it was luck rather than evidence of safety — which is a good argument for the concurrent test you asked for, since the single-threaded one could never have shown it.

Identities are now cached on `(host_ptr, nbytes)` under a lock, so the mint and the cache lookup are atomic together. `test_concurrent_copies_of_distinct_ranges_never_share_an_identity` drives 64 distinct ranges through 8 threads and asserts the mapping stays one-to-one in both directions — a duplicate id makes the child refuse the import, a missing one breaks reuse.

python/pypto/runtime/distributed_runner.py
lterrac: Confirmed, and this is the half that made me change the design rather than add a release. Fixed in b6789a9a.

Releasing the named Buffer would have been wrong: the backing is the caller's own fork-inherited mapping, not something this Worker allocated, so there is nothing for it to unlink — and dropping the child's view after every copy would defeat the reuse the named path exists for. Keying the identity on `(host_ptr, nbytes)` gets the bound you asked for instead: a range copied N times keeps one name, so live entries are bounded by the number of distinct named ranges, and your `copy_stacked_from`-per-step case now reuses one identity per shard for the life of the process instead of adding one per step. The cache is dropped with the spans it names, in `release_inherited_host_tensor_refs` and in `close`.

You were also right that `test_named_ranges_get_distinct_buffer_ids` pinned the behaviour that produced the growth. It is gone, restated as the two invariants that matter: one range keeps one identity across copies, and distinct sub-ranges do not collide — the latter being what a sharded upload needs.

YunjiQin (approved):

PR #2424, opened 2026-08-18
fix(language): keep the C++ `Check failed:` tail out of DSL error headers

Hzfengsy:
## Summary

`FatalLogger::~FatalLogger` appends `\nCheck failed: <expr> at <file>:<line>` to every check message, and `CHECK` throws `pypto::ValueError` — so every backend op-validation failure lands in `_dispatch_op`'s `except (TypeError, ValueError)` branch, which stringified the exception raw. The `except Exception` branch four lines below already sanitized; this one was simply missed.

`_dispatch_op` is the dispatcher for every DSL namespace (`pl.tensor`, `pl.tile`, `pl.system`, `pl.array`, `pl.prefetch`, `pld.*`, unified `pl.*`), fed by 1,358 `CHECK`/`CHECK_SPAN` statements under `src/ir/op/` — which is what makes this the dominant DSL error path.

**Before** — `pl.tile.matmul` on FP16 × FP32 operands:

```text
Error: pl.tile operation 'matmul': The operator tile.matmul requires identical lhs and rhs data types, but got fp16 and fp32
Check failed: lhs_type->dtype_ == rhs_type->dtype_ at /abs/path/.../src/ir/op/tile_ops/matmul.cpp:82 at kernel.py:9:37
  --> kernel.py:9:37
   |
 9 |     c: pl.Tile[[64, 64], pl.FP32] = pl.tile.matmul(a, b)
   |                                     ^^^^^^^^^^^^^^
```

A C++ boolean expression and an absolute build-machine path are the second thing the user reads — rendered *inside* the bold red header, since `renderer.py` splices the whole multi-line message into `f"Error: {message}"`, with two run-together `at file:line` clauses (C++ path first).

**After:**

```text
Error: pl.tile operation 'matmul': The operator tile.matmul requires identical lhs and rhs data types, but got fp16 and fp32
  --> kernel.py:9:37
   |
 9 |     c: pl.Tile[[64, 64], pl.FP32] = pl.tile.matmul(a, b)
   |                                     ^^^^^^^^^^^^^^
```

## What changed

- **`ast_parser.py`** — the `(TypeError, ValueError)` branch of `_dispatch_op` routes through the existing `concise_error_message()`. The `pl.func_attr` duplicate-key handler gets the same treatment: it surfaces a `CHECK` in `IRBuilder::AddFunctionAttrs` identically.
- **`exceptions.py`** — `concise_error_message()` no longer returns an empty string. See below.
- **`matmul_mx.cpp`** — the two `CHECK`s with no `<<` message get real user-facing text.
- **docs** — a subsection under *User-facing checks*, `en` + `zh`.

The C++ detail is hidden, not destroyed — `PTO_BACKTRACE=1` still prints it through the chained `__cause__`, and a test pins that.

## Why the helper needed hardening

Stripping alone would have regressed a check carrying no `<<` payload. `FatalLogger` emits its newline unconditionally, so for `CHECK(cond);` the tail starts at offset 1 and truncating at that newline leaves `""` — an empty bold `Error:` header, strictly worse than the noise it replaced. `concise_error_message()` now falls back to a named message, and the only two such checks repo-wide (the `tile.matmul_mx` scale operands, reached when K is symbolic) get real text. Those two are **split in two checks** because the rank message cannot name the actual rank while the null guard shares the predicate.

## One test expectation changed

`test_strips_check_failed_at_start` asserted the fallback string `"Internal backend check failed"`. Two independent reasons that is a test fix rather than a test hack:

1. **The wording was wrong.** `CHECK` is the *user*-error macro (`.claude/rules/error-checking.md`); telling someone with an invalid kernel that they hit an "Internal" failure sends them to file a compiler bug. The helper also runs on genuine `InternalError` messages, so the replacement is class-neutral.
2. **Its fixture pinned a shape production never emits.** The input `"Check failed: dim > 0 at src/foo.cpp:42"` has no leading `\n`, but `FatalLogger` always writes one first — so the branch it exercised is dead in production. Every real message-less check takes the other path, which is exactly the empty-string bug. A new test adds the fixture that does occur.

## Testing

- `tests/ut/` — **9,912 passed, 3 skipped, 0 failed**; `ctest` 1/1.
- 7 new tests. Each pairs a negative assertion (`"Check failed" not in msg`) with a positive one (the op's own text survives; the C++ detail survives on `__cause__`), so an over-eager strip cannot pass them.
- Verified by hand end to end, both with and without `PTO_BACKTRACE=1`.

## Interaction with #2419

No conflict — verified with a trial merge (`git merge-tree`, clean). The two are complementary and neither subsumes the other: #2419 adds a `BUG_CLASS_EXCEPTIONS` passthrough *above* this branch so bug-class exceptions escape untouched, and makes a C++ `TypeError` arrive as a real `TypeError` instead of a flattened `ValueError` — it still carries the `Check failed:` tail, and this PR is what strips it. #2419 also keeps appending the IR span *after* the tail, so the strip behaves identically post-merge. Whichever lands second merges cleanly; no rebase needed.

Note for reviewers: after the fix, a `CHECK_SPAN` diagnostic still ends with an inline ` [<abs path>:<line>:<col>]`, because `FatalLogger` writes the span *before* the newline the strip keys on. Left alone deliberately — `decorator.py` raises with no `span=`, where that inline span is the user's only location, so a blanket strip would regress it. It belongs call-site side as a follow-up.

## Compatibility

No interface change. A valid kernel compiles exactly as before; only the text of an already-failing diagnostic changes.

### DIFF
--- a/docs/en/dev/02-error-handling.md
+++ b/docs/en/dev/02-error-handling.md
@@ -95,6 +95,17 @@ CHECK_SPAN(shape.size() == 2, span) << "tensor.matmul: only 2D inputs are suppor
 
 The `span` argument follows the same safety rule as `INTERNAL_CHECK_SPAN`: it is evaluated only on failure, but unconditionally evaluated there. The span source must therefore be safe to dereference at the failure point (typically a local `Span` variable or a sibling IR node known non-null).
 
+#### The `Check failed:` tail is stripped before it reaches DSL users
+
+`FatalLogger::~FatalLogger` appends `\nCheck failed: <expr> at <file>:<line>` to **every** check message, `CHECK` included. That tail names a C++ expression and an absolute build-machine path — useful when debugging PyPTO, noise for someone whose kernel is simply invalid. Since the renderer splices the whole message into its bold `Error:` header, an unstripped tail lands between the header and the `-->` arrow pointing at the user's own source.
+
+The DSL parser therefore routes a user-facing backend exception through `concise_error_message()` (`python/pypto/language/parser/diagnostics/exceptions.py`) before wrapping it in a `ParserError`. The raw text stays reachable via `PTO_BACKTRACE=1`, which prints the Python traceback carrying the original exception as `__cause__`.
+
+Two consequences for op authors:
+
+- **Always give `CHECK` a `<<` message.** Once the tail is stripped, a bare `CHECK(cond);` has nothing left to say, and the user gets a generic "backend check reported no message" placeholder instead of an actionable error.
+- **Do not hand-roll `throw pypto::ValueError(...)` to dodge the tail.** That workaround predates the parser-side strip and is no longer needed for DSL-reachable checks.
+
 ### Internal invariant checks — `INTERNAL_CHECK_SPAN`
 
 Throw `InternalError` when an internal invariant is violated. Always attach the IR node's `Span` so the error message includes the user's source location:
--- a/docs/zh/dev/02-error-handling.md
+++ b/docs/zh/dev/02-error-handling.md
@@ -92,6 +92,17 @@ CHECK_SPAN(shape.size() == 2, span) << "tensor.matmul: only 2D inputs are suppor
 
 `span` 参数遵循与 `INTERNAL_CHECK_SPAN` 相同的安全规则:它仅在失败时求值,但在失败路径中是无条件求值的。因此 span 来源必须在失败点可以安全求值(典型如局部 `Span` 变量或已确认非空的兄弟 IR 节点)。
 
+#### `Check failed:` 尾部在抵达 DSL 用户前会被剥离
+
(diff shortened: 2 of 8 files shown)


Hzfengsy:
**Addressed** — `exceptions.py:147-184`, *"Track whether the check tail was stripped before using the fallback."*

Valid, and reproduced before fixing: `GetFullMessage()` always appends a traceback block (or the "no stack trace" note), so a bare `pypto::ValueError("")` under `PTO_BACKTRACE=1` strips to empty exactly like a message-less `CHECK` does. Gating on *"the raw message was non-empty"* therefore labelled it a silent backend check — naming a check that never ran, and telling the user to enable a flag they already had on.

Fixed in f9887987 by tracking whether the `Check failed:` tail was actually removed and gating on that instead, plus a regression test covering both the traceback and the no-symbols shape. Full unit suite: 9,913 passed, 0 failed.

lyfne123 (approved):

PR #2425, opened 2026-08-18
[experiment, do not merge] restore the pre-#2378 cross-core ring depth on A5

luohuan19:
Throwaway branch. **Do not merge, do not review** — its only purpose is to run `mixed-kernel-tests-a5` once on real 950 silicon. It will be closed and the branch deleted as soon as that job reports.

## What this probes

`test_qwen3_decode_scope3_mixed[a5]` passes on board at `main@71020585` and stalls 3/3 at `main >= 49ed7797` — scheduler timeout, `S1:running-stalled`, always inside the `oproj_block` mixed kernel, at a different task index and core each run (37/core 2, 32/core 1, 28/core 2).

Comparing the A5 device kernels compiled from those two bases, the only semantic difference is #2378: `oproj_block`'s C2V ring went 8 → 2 slots, `mlp_block`'s bidirectional rings 4 → 2. Every kernel without a cross-core pipe is byte-identical, and the orchestration differs only by the runtime type renames from #2273.

This commit puts the ring depth back on PTOAS's own `dir_mask` derivation (8 unidirectional / 4 bidirectional), reproducing the last-known-good device configuration exactly — the generated `TPipe<>` template arguments and every `.pto` match `main@71020585`, apart from `slot_num` now being stated explicitly instead of left implicit.

- **`mixed-kernel-tests-a5` green** → #2378 is the trigger; the real fix is tracked separately.
- **`mixed-kernel-tests-a5` red** → the search moves to the Simpler / PTO-ISA bump in #2273.

## Expected reds

The unit tests that pin the 2-slot default are deliberately untouched, so `unit-tests` and `codegen-tests` will fail here. Neither gates `mixed-kernel-tests-a5` (`needs: [changes, toolchain, pre-commit, clang-tidy]`).

### DIFF
--- a/.github/actions/setup-ci-job/action.yml
+++ b/.github/actions/setup-ci-job/action.yml
@@ -80,6 +80,13 @@ inputs:
       CANN_ROOT is not required in the runner config.
     required: false
     default: 'false'
+  npu-check-timeout:
+    description: >-
+      Maximum seconds the initial npu-smi task-submit probe may wait for the
+      requested device. A5 callers override the 5-minute default because that
+      pool can stay occupied by longer on-board suites.
+    required: false
+    default: '300'
   install-toolchain:
     description: >-
       When 'true' (the default) download ptoas, clone the pinned pto-isa, and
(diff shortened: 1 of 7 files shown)


luohuan19:
Experiment done. Result: **negative — #2378 is not the cause.**

With the automatic cross-core ring depth restored to PTOAS's own dir_mask derivation (8 unidirectional / 4 bidirectional, byte-identical device config to the last-known-good `main@71020585`), `test_qwen3_decode_scope3_mixed[a5]` still stalls with the same signature:

```
sub_class=S1:running-stalled (detail=1) completed=20/563 running=1 ready=0
waiting=542 orch_done=1 stuck_task_id=8589934610 stuck_core=2
```

Task 18, again inside the oproj_block loop, again at a different index than the three earlier runs (37 / 32 / 28). Closing; the bisect continues on a separate branch.

PR #2426, opened 2026-08-18
[experiment, do not merge] A5 mixed-kernel stall: bisect step at #2273's parent

luohuan19:
Throwaway branch. **Do not merge, do not review** — it exists only to run `mixed-kernel-tests-a5` once on real 950 silicon. Closed and deleted as soon as that job reports.

## Where the bisect stands

`test_qwen3_decode_scope3_mixed[a5]` passes on board at `main@71020585` and stalls at `main >= 49ed7797` — `sched_error_code=100`, `S1:running-stalled`, always inside the `oproj_block` mixed kernel, at a different task index and core every run (37 / 32 / 28 / 18).

**#2378 is already ruled out** (see the closed experiment PR #2425): restoring the cross-core ring depth to 8 unidirectional / 4 bidirectional reproduces the last-known-good device configuration byte-for-byte and the stall persists.

With the ring depth neutralised, the generated A5 device code between the good and bad bases differs in exactly three ways — every `.pto` is identical, and the orchestration differs only by the runtime type renames:

| delta | origin | device-visible |
|---|---|---|
| `Tensor` -> `ChipTensor` rename | #2273 | no (naming only) |
| `#if defined(__CPU_SIM)` shim removed | #2273 | no (not compiled on board) |
| `set_atomic_none()` injected at `kernel_entry` | #2382 | **yes** |

So the remaining suspects are #2382 (`0dc98f32`) and #2273 (`8662deb9`, which bumps Simpler `3165cc89` -> `1f27a157` and PTO-ISA `83d01313` -> `f51c92f6`).

## What this branch tests

`33ff8b49` — the commit immediately before #2273 — plus the three CI-only commits that define the job (cherry-picked clean; they touch only `.github/workflows/ci.yml`, `tests/st/`, and docs). #2382 and #2378 are both present here; only the Simpler / PTO-ISA bump and the ten commits after it are excluded. The `runtime` pin is back at `3165cc89`.

- **`mixed-kernel-tests-a5` green** -> the Simpler / PTO-ISA bump in #2273 is the cause; #2382 and #2378 are both cleared.
- **`mixed-kernel-tests-a5` red** -> the cause is older than #2273, which makes #2382 the prime suspect since it is the only remaining device-code delta.

Everything else on this branch is plain `main` history, so the other jobs should behave normally.

### DIFF
--- a/.github/actions/setup-ci-job/action.yml
+++ b/.github/actions/setup-ci-job/action.yml
@@ -80,6 +80,13 @@ inputs:
       CANN_ROOT is not required in the runner config.
     required: false
     default: 'false'
+  npu-check-timeout:
+    description: >-
+      Maximum seconds the initial npu-smi task-submit probe may wait for the
+      requested device. A5 callers override the 5-minute default because that
+      pool can stay occupied by longer on-board suites.
+    required: false
+    default: '300'
   install-toolchain:
     description: >-
       When 'true' (the default) download ptoas, clone the pinned pto-isa, and
@@ -269,7 +276,9 @@ runs:
     - name: Check NPU
       if: ${{ inputs.needs-device == 'true' }}
       shell: bash
-      run: npu-smi info
+      run: |
+        task-submit --device "$DEVICE_ID" --timeout "${{ inputs.npu-check-timeout }}" --max-time 120 \
+          --run "cd $GITHUB_WORKSPACE/${{ inputs.checkout-path }} && npu-smi info"
 
     - name: Scope ccache to pto-isa version
       # Namespace ccache by the pinned pto-isa commit so an ISA bump forces a real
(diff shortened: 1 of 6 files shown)


luohuan19:
Experiment done. Result: **negative — #2273 is not the cause either.**

At `33ff8b49` (immediately before #2273, `runtime` pin back at `3165cc89`) the stall reproduces unchanged:

```
sub_class=S1:running-stalled (detail=1) completed=32/563 running=1 ready=0
waiting=530 orch_done=1 stuck_task_id=8589934622 stuck_core=2
```

Task 30, core 2, inside the oproj_block loop — same shape as every other run, at yet another index (37 / 32 / 28 / 18 / 30).

Bisect state: the regression is in `71020585..33ff8b49` (24 commits). #2378 and #2273 are both cleared. Closing; the next step runs on a separate branch.

PR #2430, opened 2026-08-19
[experiment, do not merge] A5 mixed-kernel stall: test #2382's set_atomic_none() guard

luohuan19:
Throwaway branch. **Do not merge, do not review** — one `mixed-kernel-tests-a5` run, then closed and deleted.

## Bisect state

`test_qwen3_decode_scope3_mixed[a5]` stalls on board with `sched_error_code=100`, `S1:running-stalled`, always inside the `oproj_block` mixed kernel, at a different task index and core every run (37 / 32 / 28 / 18 / 30 — 5 for 5).

| candidate | verdict | evidence |
|---|---|---|
| #2378 cross-core ring 8/4 → 2 | **ruled out** (#2425) | restoring the ring depth reproduces the known-good device config byte-for-byte; still stalls |
| #2273 Simpler / PTO-ISA bump | **ruled out** (#2426) | `33ff8b49` stalls with `runtime` pinned back at `3165cc89` |

Remaining window: `71020585..33ff8b49`, 24 commits.

## Why this candidate

Comparing the A5 artifacts of the good and bad bases with the ring depth neutralised, every `.pto` is identical and the orchestration differs only by the runtime type renames. The **only** remaining device-code delta is the `set_atomic_none()` builtin that #2382 injects at the top of every `kernel_entry` — on both lanes of a mixed kernel, before the cross-core `TPipe` is constructed.

This branch removes that injection and nothing else. The regenerated `oproj_block.pto` is unchanged, and each kernel `.cpp` differs from the failing build by exactly those 5 lines.

- **green** → the regression is #2382.
- **red** → the bisect moves to the midpoint of the remaining window (`471b0102`).

`test_pto_codegen.py` covers the guard and is deliberately untouched, so `codegen-tests` / `unit-tests` are expected to fail here. Neither gates `mixed-kernel-tests-a5`.

PR #2431, opened 2026-08-19
[experiment, do not merge] A5 mixed-kernel stall: bisect midpoint 471b0102

luohuan19:
Throwaway branch. **Do not merge, do not review** — one `mixed-kernel-tests-a5` run, then closed and deleted.

## Bisect state

`test_qwen3_decode_scope3_mixed[a5]` stalls on board with `sched_error_code=100`, `S1:running-stalled`, always inside the `oproj_block` mixed kernel, at a different task index every run (37 / 32 / 28 / 18 / 30 / 24) and on a varying core (2 / 1 / 2 / 2 / 2 / 3). Good at `main@71020585`; bad at `main@33ff8b49` and everything after.

| candidate | verdict | evidence |
|---|---|---|
| #2378 cross-core ring 8/4 → 2 | **ruled out** (#2425) | restoring the ring depth reproduces the known-good device config byte-for-byte; still stalls |
| #2273 Simpler / PTO-ISA bump | **ruled out** (#2426) | `33ff8b49` stalls with `runtime` pinned back at `3165cc89` |
| #2382 `set_atomic_none()` guard | **ruled out** (#2430) | removing the injection still stalls |

#2382 was the *only* device-code delta across the whole window once the ring depth was neutralised — every `.pto` is identical between the good and bad bases and the orchestration differs only by runtime type renames. With it cleared, the regression is very likely not in the compiled artifacts at all, which points at the host / runtime side of the remaining window.

## What this branch tests

`471b0102` — the midpoint of the remaining 24-commit window `71020585..33ff8b49` — plus the three CI-only commits that define the job (cherry-picked clean; they touch only `.github/workflows/ci.yml`, `tests/st/`, and docs). The `runtime` pin is `3165cc89`.

- **green** → the regression is in `b22cee1e..33ff8b49`
- **red** → it is in `8e27c013..471b0102` (which contains #2351 `pl.split_aiv` placement, #2371 L0C double buffering, #2350 replay-from-build_output)

Everything else here is plain `main` history, so the other jobs should behave normally.

### DIFF
--- a/.github/actions/setup-ci-job/action.yml
+++ b/.github/actions/setup-ci-job/action.yml
@@ -80,6 +80,13 @@ inputs:
       CANN_ROOT is not required in the runner config.
     required: false
     default: 'false'
+  npu-check-timeout:
+    description: >-
+      Maximum seconds the initial npu-smi task-submit probe may wait for the
+      requested device. A5 callers override the 5-minute default because that
+      pool can stay occupied by longer on-board suites.
+    required: false
+    default: '300'
   install-toolchain:
     description: >-
       When 'true' (the default) download ptoas, clone the pinned pto-isa, and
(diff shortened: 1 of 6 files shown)


luohuan19:
Superseded: this branch was based on 471b0102 and cherry-picked the three CI commits, which conflicts with main (ci.yml and docs drifted since that base). GitHub could not compute the merge ref, so no workflow ran at all — the checks list only ever showed CodeRabbit.

Re-running the same bisect point from a conflict-free branch that adds a standalone A5-only workflow instead of cherry-picking.

PR #2434, opened 2026-08-19
[experiment, do not merge] A5 mixed-kernel stall: bisect midpoint 471b0102 (A5 job only)

luohuan19:
Throwaway branch. **Do not merge, do not review** — one on-board A5 run, then closed and deleted.

This branch runs **exactly one job**: `a5-experiment / a5-qwen3-mixed`. It adds a standalone `a5-exp.yml` with no lint / unit-test / docs gate in front of the A5 job, and removes the `pull_request` trigger from `ci.yml` so the full matrix never runs here and never competes for the shared NPU pool. (`toolchain` is kept only as the source of the ptoas version and pto-isa commit that `setup-ci-job` needs — GitHub-hosted, ~13s, no in-house slot.)

It replaces #2431, which cherry-picked the three CI commits onto this same base; `ci.yml` and the docs have drifted since `471b0102`, so that branch was `CONFLICTING` and GitHub never computed a merge ref — no workflow ran at all.

## Bisect state

`test_qwen3_decode_scope3_mixed[a5]` stalls on board with `sched_error_code=100`, `S1:running-stalled`, always inside the `oproj_block` mixed kernel, at a different task index every run (37 / 32 / 28 / 18 / 30 / 24) and on a varying core (2 / 1 / 2 / 2 / 2 / 3). Good at `main@71020585`; bad at `main@33ff8b49` and everything after.

| candidate | verdict | evidence |
|---|---|---|
| #2378 cross-core ring 8/4 → 2 | **ruled out** (#2425) | restoring the ring depth reproduces the known-good device config byte-for-byte; still stalls |
| #2273 Simpler / PTO-ISA bump | **ruled out** (#2426) | `33ff8b49` stalls with `runtime` pinned back at `3165cc89` |
| #2382 `set_atomic_none()` guard | **ruled out** (#2430) | removing the injection still stalls |

#2382 was the *only* device-code delta across the whole window once the ring depth was neutralised — every `.pto` is identical between the good and bad bases and the orchestration differs only by runtime type renames. With it cleared, the regression is very likely not in the compiled artifacts at all.

## What this tests

`471b0102`, the midpoint of the remaining 24-commit window `71020585..33ff8b49`. The `runtime` pin is `3165cc89`.

- **green** → the regression is in `b22cee1e..33ff8b49`
- **red** → it is in `8e27c013..471b0102` (which contains #2351 `pl.split_aiv` placement, #2371 L0C double buffering, #2350 replay-from-build_output)

### DIFF
--- a/.github/actions/setup-ci-job/action.yml
+++ b/.github/actions/setup-ci-job/action.yml
@@ -269,7 +269,12 @@ runs:
     - name: Check NPU
       if: ${{ inputs.needs-device == 'true' }}
       shell: bash
-      run: npu-smi info
+      # EXPERIMENT BRANCH: the npu-a5 pool cannot init dcmi outside a card
+      # session, so a bare `npu-smi info` fails with -8005. Borrow the card the
+      # job was assigned and run the probe inside it.
+      run: |
+        task-submit --device "$DEVICE_ID" --timeout 3600 --max-time 120 \
+          --run "cd $GITHUB_WORKSPACE/${{ inputs.checkout-path }} && npu-smi info"
 
     - name: Scope ccache to pto-isa version
       # Namespace ccache by the pinned pto-isa commit so an ISA bump forces a real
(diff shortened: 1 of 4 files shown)


luohuan19:
Result: **RED at `471b0102`** — the regression is in the older half.

```
sub_class=S1:running-stalled (detail=1) completed=37/563 ...
1 failed, 3 deselected in 80.50s
```

Remaining window: `71020585..471b0102` (12 commits).

Two setup fixes had to be carried on this branch because they are not on main yet and the npu-a5 pool needs both — a bare `npu-smi info` cannot init dcmi outside a card session (`-8005`), and CANN >= 9.2.0 ships an unrelated package of its own named `pypto` that shadows the freshly built one unless the venv's site-packages is put ahead of it on PYTHONPATH. Closing; the next probe runs on its own branch.

Issue #2411, opened 2026-08-18
[Bug] #2351 stalls DeepSeek V4 MTP-1 packed prefill at EP8


high-cloud:
## Background

DeepSeek V4 Flash MTP-1 serving stalls in the first packed-prefill request after upgrading PyPTO from `1eb0a861` to `e2079529` (the merge commit for #2351, `feat(ir): make pl.split_aiv authoritative for AIV placement`). The immediate parent passes the same accuracy test with the same Simpler, PTO ISA, PTOAS, serving, pypto-lib, model, and hardware.

This is distinct from #2354: that issue stalls on the second persistent decode run after a Simpler/Buffer-ABI pin advance. This regression stalls on the first packed-prefill run, and the good/bad PyPTO revisions pin identical runtime and toolchain dependencies.

## Environment

| Component | Value |
|---|---|
| First bad PyPTO | `e2079529db77e2a726169f18d9c6e161ef51ec1b` (#2351) |
| Immediate good PyPTO | `1eb0a86139c3d13c0d2075f5402c28f788e8e0da` |
| Simpler | `3165cc89b6ea6b58a0bc01cbec2d5f72f2029c35` (identical in both) |
| PTO ISA | `83d01313d9bfc247c4b7c8bcf969d1019f0d106f` (identical in both) |
| PTOAS | v0.57 (identical in both) |
| pypto-serving | `01fd781ee817deff5aff768a401c413b59b3b055` |
| pypto-lib | `76dfefebc7fc694e065cb3b0d13223b2a28ca33a` |
| Platform | 8 x Ascend 910B2, CANN 9.0.0 |

## Reproduction

Use the PyPTO revision's own pinned Simpler checkout and PTO ISA pin. From the serving checkout, with the DeepSeek V4 Flash W8A8 model available:

```bash
export PYPTO_DSV4_MODEL_DIR=/path/to/dsv4-flash-w8a8
export PYPTO_DSV4_RECIPES_COMPLETION=1
export PYPTO_RUNTIME_LOG=error
export SERVING_WORKER_STEP_TIMEOUT=1800
export PTO2_RING_DEP_POOL=131072
export PTO2_RING_TASK_WINDOW=131072
export PTO2_RING_HEAP=2147483648
export SIMPLER_OP_EXECUTE_TIMEOUT_US=400000000
export SIMPLER_STREAM_SYNC_TIMEOUT_MS=440000
[...]

PR #2435, opened 2026-08-19
[experiment, do not merge] A5 stall probe at e2079529 (#2351)

luohuan19:
Throwaway single-shot probe. **Do not merge, do not review.** Schedules exactly one job (`a5-qwen3-mixed`); `ci.yml` and `docs.yml` have their `pull_request` triggers removed on this branch so nothing else is scheduled and nothing competes for the shared NPU pool.

## Bisect state

`test_qwen3_decode_scope3_mixed[a5]` stalls on board — `sched_error_code=100`, `S1:running-stalled`, always inside the `oproj_block` mixed kernel, at a different task index every run (37 / 32 / 28 / 18 / 30 / 24 / 37) and on a varying core. Good at `main@71020585`.

| probe | result |
|---|---|
| #2378 ring depth restored to 8/4 | still stalls → ruled out |
| `33ff8b49` (before #2273 Simpler/PTO-ISA bump) | still stalls → ruled out |
| `set_atomic_none()` injection removed (#2382) | still stalls → ruled out |
| `471b0102` | still stalls |

Remaining window: `71020585..471b0102`, 12 commits.

## What this tests

`e2079529` — #2351 `feat(ir): make pl.split_aiv authoritative for AIV placement` — the 4th of those 12. Picked ahead of the blind midpoint because #2411 already reports #2351 stalling DeepSeek V4 MTP-1 packed prefill, which is the same shape of regression.

- **red** → the culprit is in `8e27c013..e2079529` (4 commits; #2351 the prime suspect)
- **green** → #2351 is cleared and the culprit is in `cde1046c..471b0102` (8 commits)

This branch carries two `setup-ci-job` fixes that are not on main yet and that the npu-a5 pool needs: the NPU probe must run inside a `task-submit` card session (a bare `npu-smi info` fails with `dcmi ... -8005`), and the venv site-packages must be ahead of CANN's on `PYTHONPATH` (CANN >= 9.2.0 ships an unrelated package of its own named `pypto` that otherwise shadows the freshly built one). Neither touches compiler or runtime behaviour.

### DIFF
--- a/.github/actions/setup-ci-job/action.yml
+++ b/.github/actions/setup-ci-job/action.yml
@@ -269,7 +269,9 @@ runs:
     - name: Check NPU
       if: ${{ inputs.needs-device == 'true' }}
       shell: bash
-      run: npu-smi info
+      run: |
+        task-submit --device "$DEVICE_ID" --timeout 3600 --max-time 120 \
+          --run "cd $GITHUB_WORKSPACE/${{ inputs.checkout-path }} && npu-smi info"
 
     - name: Scope ccache to pto-isa version
       # Namespace ccache by the pinned pto-isa commit so an ISA bump forces a real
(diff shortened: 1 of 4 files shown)


luohuan19:
Result: **RED at `e2079529`.**

```
sub_class=S1:running-stalled (detail=1) completed=45/563 running=1 ready=0
waiting=517 orch_done=1 stuck_task_id=8589934635 stuck_core=2
```

Remaining window is down to four commits, `71020585..e2079529`:

| | commit | |
|---|---|---|
| 1 | `8e27c013` | fix(ir): check innermost dim on GM->cube tile loads (#2356) |
| 2 | `f0449bf5` | fix(ir): attribute synthesized tile ops to their own source spans (#2358) |
| 3 | `1eb0a861` | feat(distributed): add pld.tensor.remote_store for computed values (#2357) |
| 4 | `e2079529` | feat(ir): make pl.split_aiv authoritative for AIV placement (#2351) |

Next probe is `1eb0a861` rather than the blind midpoint: green there pins the regression on #2351 outright in a single round, and #2351 is already reported stalling a different model in #2411. Closing.

PR #2436, opened 2026-08-19
[experiment, do not merge] A5 stall probe at 1eb0a861 (#2351's parent)

luohuan19:
Throwaway single-shot probe. **Do not merge, do not review.** Schedules exactly one job (`a5-qwen3-mixed`); `ci.yml` and `docs.yml` have their `pull_request` triggers removed on this branch so nothing else runs and nothing competes for the shared NPU pool.

## Bisect state

`test_qwen3_decode_scope3_mixed[a5]` stalls on board — `sched_error_code=100`, `S1:running-stalled`, always inside the `oproj_block` mixed kernel, at a different task index every run and on a varying core. Good at `main@71020585`.

| probe | result |
|---|---|
| #2378 ring depth restored to 8/4 | still stalls → ruled out |
| `33ff8b49` (before #2273 Simpler/PTO-ISA bump) | still stalls → ruled out |
| `set_atomic_none()` removed (#2382) | still stalls → ruled out |
| `471b0102` | red |
| `e2079529` (#2351) | red |

Window is down to four commits, `71020585..e2079529`:

| | commit | |
|---|---|---|
| 1 | `8e27c013` | fix(ir): check innermost dim on GM->cube tile loads (#2356) |
| 2 | `f0449bf5` | fix(ir): attribute synthesized tile ops to their own source spans (#2358) |
| 3 | `1eb0a861` | feat(distributed): add pld.tensor.remote_store for computed values (#2357) |
| 4 | `e2079529` | feat(ir): make pl.split_aiv authoritative for AIV placement (#2351) |

## What this tests

`1eb0a861` — #2351's immediate parent — instead of the blind midpoint:

- **green** → the regression is **exactly #2351**, pinned in one round. #2411 already reports #2351 stalling DeepSeek V4 MTP-1 packed prefill, which is the same shape of regression.
- **red** → the culprit is one of the three older commits, none of which is an obvious fit (#2357 is distributed-only and this is a single-card kernel; #2358 only moves source spans).

This branch carries two `setup-ci-job` fixes that are not on main yet and that the npu-a5 pool needs: the NPU probe must run inside a `task-submit` card session (a bare `npu-smi info` fails with `dcmi ... -8005`), and the venv site-packages must be ahead of CANN's on `PYTHONPATH` (CANN >= 9.2.0 ships an unrelated package of its own named `pypto` that otherwise shadows the freshly built one). Neither touches compiler or runtime behaviour.

### DIFF
--- a/.github/actions/setup-ci-job/action.yml
+++ b/.github/actions/setup-ci-job/action.yml
@@ -269,7 +269,9 @@ runs:
     - name: Check NPU
       if: ${{ inputs.needs-device == 'true' }}
       shell: bash
-      run: npu-smi info
+      run: |
+        task-submit --device "$DEVICE_ID" --timeout 3600 --max-time 120 \
+          --run "cd $GITHUB_WORKSPACE/${{ inputs.checkout-path }} && npu-smi info"
 
     - name: Scope ccache to pto-isa version
       # Namespace ccache by the pinned pto-isa commit so an ISA bump forces a real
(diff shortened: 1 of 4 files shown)


luohuan19:
Result: **RED at `1eb0a861`** — #2351 is cleared.

```
sub_class=S1:running-stalled (detail=1) completed=37/563 running=1 ready=0
waiting=525 orch_done=1 stuck_task_id=8589934627 stuck_core=1
```

Three commits left in `71020585..1eb0a861`: #2356, #2358, #2357.

Before narrowing further, the next probe re-tests `71020585` itself. The "good" data point is a CI run from 2026-08-13 on a differently shaped branch; every probe since then recompiles an old commit but runs against *today's* runner environment. If `71020585` no longer passes, the regression is environmental and none of the commit-level verdicts mean anything. Closing.

PR #2437, opened 2026-08-19
[experiment, do not merge] A5 stall: control probe at the known-good base 71020585

luohuan19:
Throwaway single-shot probe. **Do not merge, do not review.** Schedules exactly one job (`a5-qwen3-mixed`).

## Why re-test a commit that already passed

`test_qwen3_decode_scope3_mixed[a5]` passed on board at `main@71020585` in a CI run on **2026-08-13**. Every probe since has recompiled an old commit but executed against **today** runner environment — CANN, driver, firmware, the ptoas cache and ccache have all had six days to drift, and the branch shape is different too (single job, two setup patches).

The whole bisect rests on that six-day-old green still being reproducible. It has never been checked. This probe checks it.

- **green** → the premise holds; the window really is `71020585..1eb0a861` and two more rounds pin it.
- **red** → the regression is environmental, not a pypto commit, and every commit-level verdict below is void.

## Bisect state so far

Stall signature is always the same: `sched_error_code=100`, `S1:running-stalled`, inside the `oproj_block` mixed kernel, at a different task index and core every run.

| probe | result |
|---|---|
| #2378 ring depth restored to 8/4 | red → ruled out |
| `33ff8b49` (before #2273 Simpler / PTO-ISA bump) | red → ruled out |
| `set_atomic_none()` removed (#2382) | red → ruled out |
| `471b0102` | red |
| `e2079529` (#2351) | red |
| `1eb0a861` (#2351 parent) | red → #2351 cleared |

Remaining: `8e27c013` (#2356), `f0449bf5` (#2358), `1eb0a861` (#2357).

This branch carries two `setup-ci-job` fixes not yet on main that the npu-a5 pool needs: the NPU probe must run inside a `task-submit` card session (a bare `npu-smi info` fails with `dcmi ... -8005`), and the venv site-packages must precede CANN's on `PYTHONPATH` (CANN >= 9.2.0 ships an unrelated package of its own named `pypto`). Neither touches compiler or runtime behaviour — but note they are themselves a difference from the 2026-08-13 green run, which is part of what this probe controls for.

### DIFF
--- a/.github/actions/setup-ci-job/action.yml
+++ b/.github/actions/setup-ci-job/action.yml
@@ -269,7 +269,9 @@ runs:
     - name: Check NPU
       if: ${{ inputs.needs-device == 'true' }}
       shell: bash
-      run: npu-smi info
+      run: |
+        task-submit --device "$DEVICE_ID" --timeout 3600 --max-time 120 \
+          --run "cd $GITHUB_WORKSPACE/${{ inputs.checkout-path }} && npu-smi info"
 
     - name: Scope ccache to pto-isa version
       # Namespace ccache by the pinned pto-isa commit so an ISA bump forces a real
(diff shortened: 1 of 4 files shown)


luohuan19:
Result: **RED at `71020585`** — the known-good commit no longer passes.

```
sub_class=S1:running-stalled (detail=1) completed=41/563 running=1 ready=0
waiting=521 orch_done=1 stuck_task_id=8589934631 stuck_core=1
```

Same runner instance and card as the 2026-08-13 green run (`npu-a5-3`, device 0), so this is not a per-runner or per-card difference.

**Every commit-level verdict from this bisect is void** — they were all red because there is no good/bad boundary in the source history to find. This control should have run before the first probe.

Next: reconstruct the 2026-08-13 green data point exactly. That run executed the full `cross_core` file (10 cases) before qwen3 in the same pytest session; the probes since have run qwen3 alone. Closing.

PR #2438, opened 2026-08-19
[experiment, do not merge] A5 stall: reproduce the 2026-08-13 green run shape

luohuan19:
Throwaway single-shot probe. **Do not merge, do not review.**

## The bisect was chasing a boundary that does not exist

`test_qwen3_decode_scope3_mixed[a5]` passed on board at `main@71020585` on 2026-08-13. Re-tested today, **that same commit stalls** — same runner instance and card (`npu-a5-3`, device 0), identical signature:

```
sub_class=S1:running-stalled (detail=1) completed=41/563 running=1 ready=0
waiting=521 orch_done=1 stuck_task_id=8589934631 stuck_core=1
```

So the six earlier probes (#2425, #2426, #2430, #2434, #2435, #2436) were all red simply because everything is red. **Their commit-level verdicts are void**, including the ones that appeared to rule out #2378, #2273, #2382 and #2351. This control should have run first.

## What this probe changes

One thing: the pytest invocation. The 2026-08-13 run executed the full `cross_core` file (10 cases) before qwen3 in a single session; every probe since has run qwen3 alone. Board state left by preceding work is a plausible reason the same binary behaves differently.

- **qwen3 green here** → the stall depends on what runs before it, not on any source change; the a5 job as written is order-sensitive.
- **qwen3 red here** → the 2026-08-13 green is not reproducible in any form, and the difference is environmental drift on the npu-a5 host.

Two `cross_core` cases are expected to fail: they are xfail-ed on a later commit this branch deliberately does not carry. Only the qwen3 result matters.

### DIFF
--- a/.github/actions/setup-ci-job/action.yml
+++ b/.github/actions/setup-ci-job/action.yml
@@ -269,7 +269,9 @@ runs:
     - name: Check NPU
       if: ${{ inputs.needs-device == 'true' }}
       shell: bash
-      run: npu-smi info
+      run: |
+        task-submit --device "$DEVICE_ID" --timeout 3600 --max-time 120 \
+          --run "cd $GITHUB_WORKSPACE/${{ inputs.checkout-path }} && npu-smi info"
 
     - name: Scope ccache to pto-isa version
       # Namespace ccache by the pinned pto-isa commit so an ISA bump forces a real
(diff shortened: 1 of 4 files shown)


luohuan19:
Closing — investigation stopped. The qwen3 case is being dropped from the per-PR A5 job instead; it stalls on board for every commit tried, including the one that passed on 2026-08-13, so it cannot serve as a gate until the on-board stall is understood.

PR #2456, opened 2026-08-20
refactor(ir): derive outlined scope param directions from registry effects

lyfne123:
> **Draft — stacked on #2455.** This branch carries every commit of the series up to
> and including this one, so the diff against `main` is cumulative; the change to review
> here is the **last commit only**. It goes out of draft once #2455 merges and this is
> rebased.

`ScopeOutliner::InferParamDirections` recognised exactly two write operators —
`tile.store` and `tensor.assemble` — matched by name in three separate places.
A scope whose only write to a captured tensor went through any other operator
left that tensor looking untouched: the parameter stayed `In`, the caller got no
dependency on the write, and `ConvertTensorToTileOps` (which does model those
operators) disagreed with this pass about the same call.

It now asks the registry which arguments an operator writes, so `tensor.write`,
`tensor.expand_clone`, `pld.system.notify`, `pld.tile.put` / `get`,
`pld.tensor.*`, `tile.mscatter` and every future writer reach the analysis the
same way `tile.store` always has.

The kwarg carve-outs come with it and stop being special cases. An atomic store
or assemble declares `ReadWrite` rather than `Write`, so its destination stays
on the read path and an accumulator keeps `InOut` — the rule this pass used to
spell out for two operators now holds for all of them, which is what makes an
`AtomicAdd` notify keep its signal `InOut` while a `Set` notify does not.

**`pld.system.notify`'s signal parameter now derives `Out`, not `In`.** That
operator writes the peer's slot; the missing write is what `49ed7797` fixed in
`ConvertTensorToTileOps` after it deadlocked the communication card, and the
outliner simply never learned it. `test_deferred_wait_does_not_order_later_notify_behind_waiter`
pinned the old reading, stating in its own docstring that notify "has no memory
write specification". Its subject — that no spurious waiter -> notifier edge
appears — is unaffected and still asserted; only the direction expectation and
that sentence change. Verified: with the new direction the notifier still
carries neither `manual_dep_edges` nor `compiler_manual_dep_edges`, and the
consumer still has exactly one dep.

A scope writing a captured tensor through a previously unmodelled operator now
yields `Out` (or `InOut` when the body also reads it) instead of `In`. Where the
caller then reads the pre-call variable, the existing `InOutUseDiscipline`
verifier now rejects it — correctly: the write was always there, only invisible.

- `include/pypto/ir/transforms/utils/scope_outline_utils.h`: adds
  `CallWriteTargets(call)`, the one place that answers "which variables does this
  call write", registry-driven and kwarg-resolved. `ParamReadCollector`'s
  `DestinationSlot` becomes `DestinationSlots` over it, so any operand an
  operator declares it purely overwrites is skipped on the read path and any
  `ReadWrite` operand stays on it. `AssembleDestUpgrader` — which named one
  operator — becomes `WrittenParamUpgrader` over the same helper. `AsVarLike`
  replaces `As<Var>`, so a loop-carried destination (an `IterArg`) is no longer
  invisible to the write scan (`.claude/rules/ir-kind-traits.md`). The Step-2
  `CallFinder` gains a `Submit` overload: the base visitor does not forward
  `Submit` to the `Call` handler, so a `pl.submit` inside an outlined scope
  contributed no callee direction at all.

  `StoreTargetCollector` deliberately stays `tile.store`-only and is documented
  as such. It drives the *export* machinery — a store target becomes an extra
  outlined output and `StoreEvalToAssignMutator` binds a result Var for it — and
  that exists for a write whose result the body does not already thread. An
  SSA-pure writer such as `tensor.assemble` returns the updated tensor and the
  caller binds it, so exporting it too adds a redundant output. Widening it was
  tried and rejected on that evidence.

- `tests/ut/ir/transforms/test_outline_incore_scopes.py`: three tests for
  operators the pass could not see — a `tensor.write`-only capture deriving
  `Out`, an `expand_clone` target deriving `Out` while its source stays `In`,
  and a read-then-write capture staying `InOut`. The notify test's expectation
  and docstring are corrected as described above, and one stale docstring
  reference to the removed `AssembleDestUpgrader` is updated.

- `docs/en/dev/passes/08-outline_incore_scopes.md`,
  `docs/zh/dev/passes/08-outline_incore_scopes.md`: state that which operators
  write is a registry fact rather than a list in this pass, and fold the atomic
  carve-out into the `Write` / `ReadWrite` distinction.

- `cmake --build build --parallel 32`: exit 0
- `python -m pytest tests/ut/ -n 16 -q`: exit 0 — 10008 passed, 8 skipped
  (10005 before; the 3 new tests are the difference)
- The three new tests were checked negatively against the previous commit: each
  derives `In` there.
- `tests/lint/check_headers.py`, `check_english_only.py`,
  `check_docs_en_zh_parity.py`, `check_op_name_literals.py`,
  `check_no_broad_raises.py`: exit 0 each
- `clang-format --dry-run --Werror` on the changed header: exit 0
- `ruff check` / `ruff format` on the changed test file: exit 0


lyfne123:
Folded into #2455 rather than reviewed separately.

The two are the same refactor applied to the two passes that derive parameter
directions — `ConvertTensorToTileOps` here, `ScopeOutliner` there — and the
argument that carries the change only lands with both in view: the two passes
did not merely duplicate a write table, they *disagreed*. `#2391` taught
`ConvertTensorToTileOps` that `pld.system.notify` writes its target; the
outliner never learned it, so the two gave different answers about the same
call. Split across two PRs, that is a footnote in each; together it is the
reason the refactor exists.

Every commit here is preserved in #2455 — the combined commit's tree is
byte-identical to this branch's head. Nothing is dropped.

PR #2485, opened 2026-08-21
ci: cache the build products instead of rebuilding them in every job

luohuan19:
## Summary

CI was compiling the same PyPTO tree independently in every fan-out job. On the measured green self-hosted run, eight jobs each spent 1m56-3m09 building PyPTO even with ccache; on the GitHub-hosted side, `unit-tests` and `system-tests-a5sim` separately spent 8m51 and 6m32 compiling the extension. ccache helps across source changes, but it cannot reuse CMake configuration, linking, or wheel packaging across jobs building the same tree.

This change caches the finished `pypto` and `simpler` wheels instead. Self-hosted jobs coordinate through host-local, content-addressed wheel slots, while GitHub-hosted jobs consume one Actions Cache-backed prebuild. CI also uses `-g1` instead of `-g2` for `RelWithDebInfo`, retaining symbolized backtraces while substantially reducing compile, link, package, and transfer cost; local builds keep the existing `-g2` default.

The caches remain optimizations rather than test gates. A missing or unusable cache falls back to the previous in-place build, and a failed prebuild cannot silently skip required tests.

### Self-hosted wheel cache in `setup-ci-job`

Each self-hosted job resolves slots under `$CI_CACHE_ROOT/wheels`, installs a matching wheel when present, and can build and publish one on a miss. `pypto` slots are keyed by the checkout tree, interpreter ABI/architecture, toolchain identity, debug level, and an explicit cache epoch. `simpler` slots additionally include the runtime tree, pinned PTO-ISA commit, the runtime set the host can actually build (`sim` or `sim-onboard`), and the resolved CANN installation identity.

A per-slot `flock` serializes jobs on the same filesystem. The winner builds into a temporary directory, adds `.ok` only after a wheel exists, and atomically publishes it with `mv -T`; waiters then install the completed slot. Hits stamp the slot's last-use time, and job completion prunes old slots, abandoned temporary directories, and stale lock files. The cache stays on the self-hosted machine and sends no wheel traffic through the in-house proxy.

The existing PTO-ISA ccache namespace remains in place, so an ISA pin change forces real recompilation instead of risking the ABI-mismatched objects seen in #1139.

### Self-hosted `prebuild-wheels`

`prebuild-wheels` runs `setup-ci-job` as soon as `pre-commit` succeeds, without waiting for `clang-tidy`, so it can warm the host cache before the gated fan-out begins. Consumers deliberately do not `needs:` this job: the cache is host-local, and whichever job reaches a cold key first can safely become the builder. This avoids turning an advisory warm-up job into a global scheduling gate when the CPU pool is saturated or a consumer lands on another server.

The job runs one matrix leg per architecture. The cache it fills is the host's own disk, and the self-hosted fleet is no longer a single machine: the `cpu` and `npu` pools each span an aarch64 box (`cpu-N` / `npu-N`) and an x86_64 one (`infra024-cpu-x64-N` / `infra024-npu-a2a3-N`), and #2525 removed the `arm64` pin from every consumer so a job lands on whichever host is free. Over eight recent runs, 43% of CPU-pool jobs and half the NPU jobs went to the x86_64 machine; a single-architecture prebuild would leave that machine's cache cold. `arm64` and `x64` are the runner application's own default labels and the only means in `runs-on` of distinguishing the two machines. Each machine's CPU and NPU runners share one `CI_CACHE_ROOT`, so warming from the CPU pool serves both pools, and the wheel key already carries the architecture, so the two legs write disjoint slots. `fail-fast: false` keeps one unreachable host from cancelling the other leg.

### Smaller CI debug artifacts

`PYPTO_DEBUG_INFO_LEVEL` is now a validated CMake cache variable for `RelWithDebInfo` builds. It defaults to `2`, preserving local developer behavior, while CI passes `1` only around the PyPTO build and includes that level in the wheel key. The English and Chinese installation guides document the trade-off and invocation.

Measured on this tree with GCC 15.2:

| | `-g2` (default) | `-g1` (CI) |
| --- | ---: | ---: |
| extension | 304.1 MiB | 62.5 MiB |
| wheel | 111.9 MiB | 18.9 MiB |
| `.debug_info` | 206 MiB | 22 MiB |
| `.debug_loclists` | 28.6 MiB | 0 |
| `.debug_line` | 13.3 MiB | 6.2 MiB |
| compile, 240 TUs at `-j32` | 85s | 55s |
| link | 8s | 4s |
| wheel packaging | 16s | 3s |

The retained line tables produce byte-identical libbacktrace output for the measured binding, `INTERNAL_CHECK_SPAN`, and `op_registry.cpp` throw paths. `tests/ut/core/test_error.py` keeps that claim under CI coverage.

### GitHub-hosted `prebuild-wheels-github`

`prebuild-wheels-github` builds both wheels once in `ghcr.io/hw-native-sys/pypto/github-ci:latest` (Ubuntu 22.04, GCC 15), publishes the wheelhouse through Actions Cache, and exposes its content-addressed key to `unit-tests` and `system-tests-a5sim`. Building in the older container keeps the wheel compatible with the Ubuntu 24.04 unit-test runner; consumers import the installed packages immediately to fail fast on an ABI mismatch. The job is named for whose runners it serves rather than for an architecture: every GitHub-hosted runner is x86_64, but so is one of the self-hosted machines, which `prebuild-wheels` covers with a matrix leg. What separates the two jobs is whether the host keeps its cache between runs.

The two consumers take a real `needs:` edge because GitHub-hosted runners share no disk, but their job conditions do not require the prebuild to succeed. If no wheel key or cache entry is available, they build in place and still run their tests. This preserves the original lint/toolchain gates and prevents a failed helper from turning required tests into misleading `skipped` checks.

The two former per-consumer ccaches collapse into one prebuild ccache. It is restored for every run, capped at 1 GiB, and saved only on `main` to avoid spending the repository cache quota on PR-private entries. The CI image installs ccache permanently, while the workflow tolerates an older image by attempting installation and then building without ccache if it remains unavailable. The prebuild also declares the package-read permission needed to pull the container, configures Git safe-directory ownership, and selects Bash explicitly.

## Effect

| Path | Before | After |
| --- | --- | --- |
| self-hosted per-job PyPTO build | 1m56-3m09 | about 20s on a wheel hit |
| self-hosted PyPTO builds per run | 8 | 2, one per machine, plus a cold A5 host when applicable |
| self-hosted `simpler` builds per run | 8 | 4, one per machine and runtime set, plus a cold A5 host when applicable |
| GitHub-hosted PyPTO builds per run | 2 | 1 shared prebuild |
| GitHub-hosted consumer behavior | each compiles from source | restore and install the shared wheelhouse |
| in-house proxy traffic for self-hosted wheels | 0 | 0 |

On the measured GitHub-hosted baseline run, `unit-tests` was the workflow's final job because it rebuilt an extension already built by A5Sim. With a warm prebuild ccache, the commit-range measurement expects unit testing to begin around 11:17 instead of 11:26:21 and the workflow to finish around 11:21 instead of 11:30:41.

## Verification

- All eight required CI checks succeeded (`pre-commit`, `clang-tidy`, `codegen-tests`, `unit-tests`, `system-tests`, `system-tests-a5sim`, `dist-system-tests`, and `pypto-lib-model`) on the previous head.
- The branch has since been rebased onto `ec5d20c1` (#2525) and the `prebuild-wheels` matrix added, so that run does not cover the current head; CI is re-running.

### DIFF
--- a/.github/actions/setup-ci-job/action.yml
+++ b/.github/actions/setup-ci-job/action.yml
@@ -800,6 +800,12 @@ runs:
         # is a guard against those invariants breaking, not a path CI takes.
         toolchain_prefix="${PYPTO_TOOLCHAIN:-${CONDA_PREFIX:-}}"
         toolchain_id="$(basename "${toolchain_prefix:-none}")"
+        # CI never attaches a debugger to what it builds, so it takes the
+        # cheaper of the two levels CMakeLists.txt offers: 111 MiB of wheel
+        # becomes 18 MiB, paid back once in the build and eight more times in
+        # the installs. It is in the key because it changes the artifact -- the
+        # same rule the arch and the toolchain are here for.
+        debug_level=1
         # Bump to invalidate every slot on every host at once — the escape hatch
         # for a change that alters what a slot holds without altering any input
         # named above.
@@ -872,7 +878,7 @@ runs:
           echo "::warning::this uv cannot build wheels — wheel cache disabled"
           cache_ready=false
         fi
-        pypto_key="pypto-${pypto_tree}-${abi_tag}-${toolchain_id}-b${build_inputs_id}-e${cache_epoch}"
+        pypto_key="pypto-${pypto_tree}-${abi_tag}-${toolchain_id}-g${debug_level}-b${build_inputs_id}-e${cache_epoch}"
         simpler_key="simpler-${simpler_tree}-${PTO_ISA_COMMIT:-none}-${abi_tag}-${toolchain_id}-${simpler_rt}-b${build_inputs_id}-e${cache_epoch}"
 
         # 0 = slot is complete, 1 = cache unusable (build in place instead),
@@ -964,9 +970,12 @@ runs:
         # LRU prune, before the builds rather than after them: run last, it could
         # never free space for the job that had just run out of it. install_slot
         # stamps every hit, so mtime is last use, not
-        # creation — a slot that main keeps re-testing survives. At ~123 MB a
-        # pypto slot and ~30 runs a day, two days sits well inside the headroom
-        # on both hosts (808 G free on the arm64 pool, 340 G on a5). The second
+        # creation — a slot that main keeps re-testing survives. At ~20 MB a
+        # pypto slot (PYPTO_DEBUG_INFO_LEVEL above is what makes it that
+        # rather than ~123 MB) and ~30 runs a day, two days sits well inside the
+        # headroom measured on the aarch64 cpu/npu box (808 G free) and the a5
+        # host (340 G). infra024 was not measured and does not need to be: the
+        # prune, not the headroom, is what bounds this on any host. The second
         # sweep collects temp dirs from a build that was killed mid-flight.
         if [ "$cache_ready" = true ]; then
           find "$wheel_cache" -mindepth 1 -maxdepth 1 -type d -mtime +2 -exec rm -rf {} + 2>/dev/null || true
@@ -980,6 +989,11 @@ runs:
         fi
 
         pypto_slot="$wheel_cache/$pypto_key"
+        # Exported around this build alone, not for the step: simpler's CMake
+        # has no such option, and handing it one would earn a "Manually-
+        # specified variables were not used by the project" warning in every
+        # runtime build for no gain.
+        export SKBUILD_CMAKE_DEFINE="PYPTO_DEBUG_INFO_LEVEL=$debug_level"
         rc=0; ensure_slot "$pypto_slot" . || rc=$?
         case "$rc" in
           0) echo "wheel cache: pypto <- $pypto_key"
@@ -991,6 +1005,7 @@ runs:
           *) echo "::warning::wheel cache unusable — installing pypto in place"
              py_install --no-build-isolation '.[dev]' ;;
         esac
+        unset SKBUILD_CMAKE_DEFINE
 
         # simpler (./runtime) is only needed on the codegen -> device path; the
         # --codegen-only job skips it (install-toolchain=false). Its slot is
--- a/.github/actions/setup-ci-job/action.yml
+++ b/.github/actions/setup-ci-job/action.yml
@@ -659,6 +659,7 @@ runs:
       working-directory: ${{ inputs.checkout-path }}
       env:
         INSTALL_TOOLCHAIN: ${{ inputs.install-toolchain }}
+        PTO_ISA_COMMIT: ${{ inputs.pto-isa-commit }}
       run: |
         source activate.sh
         rm -rf ./build ./_skbuild ./runtime/build ./runtime/_skbuild
@@ -674,6 +675,14 @@ runs:
             uv pip install "$@"
           else
             pip install "$@"
+          fi
+        }
+        # Same split, for producing a wheel instead of installing one.
+        py_build_wheel() {  # <out-dir> <source-dir>
+          if [ "${PYPTO_USE_UV:-false}" = "true" ]; then
+            uv build --wheel --no-build-isolation --out-dir "$1" "$2"
+          else
+            pip wheel --no-build-isolation --no-deps --wheel-dir "$1" "$2"
           fi
         }
         if [ "${PYPTO_USE_UV:-false}" != "true" ]; then
@@ -725,14 +734,13 @@ runs:
           echo "note: download.pytorch.org unreachable — falling back to PyPI"
           py_install "$TORCH_PIN"
         fi
-        py_install --no-build-isolation '.[dev]'
         # One line naming the build inputs this artifact was produced from. The
         # 2026-08-24 outage was hard to place precisely because nothing reported
         # them: an upstream release and a code change looked identical.
         # importlib.metadata rather than `pip list`: the bundle toolchain's venv
         # is built by `uv venv` and has no pip in it at all, which is exactly the
         # set of jobs that outage hit.
-        python - <<'PY'
+        build_inputs="$(python - <<'PY'
         from importlib.metadata import PackageNotFoundError, version
         names = ("cmake", "nanobind", "ninja", "scikit-build-core")
         def resolve(n):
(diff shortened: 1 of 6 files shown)


luohuan19:
Thanks @coderabbitai — both points addressed, plus one diagnosis that turned out not to belong to this PR.

**The `unknown` toolchain fallback — fixed**, folded into `ci: cache the build products instead of rebuilding them in every job` (`ace330fa`, the finding's own commit rather than a repair commit on top). The critique was right on its own terms: that fallback kept `abi_tag` while dropping the toolchain dimension, which is precisely the silently-mismatched binary the paragraph above it exists to prevent, and it was the one unresolvable input in the block that did *not* turn the cache off.

```bash
toolchain_prefix="${PYPTO_TOOLCHAIN:-${CONDA_PREFIX:-}}"
toolchain_id="$(basename "${toolchain_prefix:-none}")"
```

...with `-z "$toolchain_prefix"` added to the existing cache-off guard. Verified by running the shipped lines themselves, not a copy: both empty -> cache off, `PYPTO_TOOLCHAIN` set -> `1.2.3` and cache on, `CONDA_PREFIX` set -> `py310` and cache on. Worth recording that this is a guard against a broken invariant rather than a path CI takes — the bundle bootstrap already hard-fails on an empty `PYPTO_TOOLCHAIN` before the venv is built, and the conda path always carries `CONDA_PREFIX`.

**"Cache reuse may not work across the CPU and NPU runner pools without shared storage"** — checked, and it holds today, so no code change. Each machine's `cpu` and `npu` runners are instances on one host sharing a single `CI_CACHE_ROOT`, so warming from the CPU pool reaches both pools on that machine. That assumption is now stated explicitly in the `prebuild-wheels` comment rather than left implicit. If it ever stopped being true the consequence would be a cold cache for the NPU jobs, i.e. the in-place build this PR replaces — not a wrong artifact, since the key is content-addressed either way.

Related, and the reason the head moved: the self-hosted fleet is no longer one machine. The `cpu` and `npu` pools each span an aarch64 box (`cpu-N` / `npu-N`) and an x86_64 one (`infra024-cpu-x64-N` / `infra024-npu-a2a3-N`), and #2525 dropped the `arm64` pin from every consumer. `prebuild-wheels` now runs one matrix leg per architecture so neither machine is left cold; both legs are green on this head (arm64 1m28s, x64 1m52s). The GitHub-hosted job was renamed `prebuild-wheels-x86` -> `prebuild-wheels-hosted` -> `prebuild-wheels-github`, since the axis that produces its different plumbing is whose runners it serves, not an architecture.

**On the `examples-tests` failure** — not caused by this PR, and deliberately not "fixed" here. The one failing block is `docs/en/user/performance/01-task-granularity.md:161`, which compares `exp(a + b)` against `torch.exp(A + B)` at `rtol=1e-4, atol=1e-4` over inputs drawn by an unseeded `torch.randn` at line 50. On an ordinary draw `a + b` reaches ~5, where `exp` is ~148 and a 1e-4 absolute tolerance is below what the simulator's hardware `exp` delivers, so the block passes or fails on the luck of the draw. The identical block fails the same way on the unrelated branch `feat/examples-distributed-zoo` (job 98141600465: 39/16384 mismatched, max abs diff 2.6e-2, versus 22/16384 and 3.3e-3 here) — different magnitude each run, which points at the input rather than the kernel. Loosening someone else's assertion to turn this PR green would be the wrong fix; it needs a seed and a realistic tolerance, in its own change.

luohuan19:
Correcting my earlier comment on the `examples-tests` failure: I said the doc example's input came from an **unseeded** `torch.randn`, and that it was therefore a luck-of-the-draw flake. That was wrong — [`01-task-granularity.md:49`](https://github.com/hw-native-sys/pypto/blob/main/docs/en/user/performance/01-task-granularity.md#L49) is `torch.manual_seed(0)`, two `randn` draws follow, and nothing else on the page consumes the RNG. I missed the line. The input is identical on every run and every host, and the real picture is worse than what I described.

**Two runs of the same tree on the same host produce different outputs.** The heads `c55416cd` and `ace330fa` differ only in `.github/actions/setup-ci-job/action.yml` — pure CI config, incapable of moving a number:

```
job 98192028861  head c55416cd  infra024-cpu-x64-1  Mismatched elements: 22 / 16384
job 98210395078  head ace330fa  infra024-cpu-x64-4  Mismatched elements: 12 / 16384
both: Greatest absolute difference: 0.0033016204833984375 at index (108, 19)
      Greatest relative difference: 0.0001465580571675673 at index (101, 26)
```

Same input, same binary, different result — while the worst offender stays bit-identical. Identical input and identical binary must not produce different device output; that property is what every golden-compare test rests on.

**It is the job shape, not the architecture.** `docs-examples` runs the *identical* `run_doc_examples.py --pages docs/en/user --check-parity -p a2a3sim` invocation on the same hosts and has passed every time, 5/5 of them on x86_64. What differs is position within the job: there the doc blocks are the first simulator work, here they run after the distributed ladder (`-d 0,1,2,3`) and the teaching examples. So prior simulator activity in the same job is contaminating the later run — which also explains why the count of wrong elements moves while the extremes do not.

aarch64 does not show it. 12/12 green in CI, and running the kernel eight times in one local process gives a single bit-identical output matching torch exactly — for the fused `merged_chain` and the GM-round-tripping `two_tasks_via_gm` alike, which even share an output hash.

**Worked around in `26f47107`, deliberately narrowly:** `examples-tests` is pinned back to `arm64`, and `docs-examples` is left unpinned, because the evidence says the clean-job case is fine and pinning it would hide the contrast that localises the bug. The pin carries all of the above as a comment and should be removed once the contamination is fixed — #2525 removed every other arch pin on purpose and this must not become permanent.

Worth flagging beyond this PR: since #2525 made `[self-hosted, linux, cpu]` span both architectures, roughly 43% of cpu-pool jobs land on the x86_64 host, so this has been failing a required check about that often for **every** pull request. It surfaced here only because this branch pushed often enough to hit it twice in a row.

I did not touch the doc's tolerance. Loosening it would turn CI green while burying the nondeterminism, which is the part actually worth fixing.

lyfne123 (approved):

PR #2523, opened 2026-08-25
Upgrade PTOAS to v0.60 and pin runtime to 0659e29

yanghaoran29:
## Summary

Rebased onto latest `main` (includes #2530). This PR upgrades PyPTO to **PTOAS v0.60** and pins the **runtime** submodule to simpler **0659e29** (merged #2016; pto-isa **be5ccb76**).

### Toolchain coupling (not independently revertible)

Rolling back only `toolchain/versions.env` will break codegen/planning. The following are bound to v0.60 together:

- A2/A3 **level-3 explicit tmp** forms for `pto.tci`, narrowing `pto.tcvt`, `pto.tsort32`
- **Static valid_shape** verification on explicit-tmp codegen paths (ci/tcvt/col_sum/sort32 static-view bridges)
- **TSEL tmp geometry** UINT32 `[1, 16]` on A2/A3 (A5 retains ABI operand but does not read tmp)
- Soft **syncall** ABI / **tfillpad** rename and related codegen adjustments in this PR

Revert toolchain and codegen changes as one unit.

### InitMemRef (A2/A3 level-3 scratch)

Under `MemoryPlanner.PYPTO` or `DSA_RP` on A2/A3 only:

| Op | Behavior |
| --- | --- |
| `tile.ci` | Append FP32 scratch when tmp missing (width 192/448 by dst dtype) |
| Narrowing `tile.cast` | Append i8 scratch sized by PTOAS v0.60 `makeTCvtTmpType` (FP32→INT16, FP16→INT16/INT8/UINT8) |
| `tile.sort32` | Append scratch when required by valid width/dtype |
| `tile.sel` / `tile.sels` / `tile.prelu` | **Preserve** explicit caller tmp; InitMemRef does not replace |
| FP16→INT4 | **Excluded** from scratch set (native vconv path, no level-3 tcvt tmp) |

PTOAS planner and A5 unchanged for compiler-owned scratch insertion.

### MemoryReuse

- `tile.ci`: `.forbid_output_alias(2)` — output must not alias compiler scratch
- `tile.sel`: registry `forbid_output_alias(0,3)` only (mask + tmp); **dead lhs/rhs may be reused** on A2/A3 and A5
- Reverted incorrect A2/A3-only lhs/rhs forbid in `ForbidAliasCollector`

### A2/A3 vs A5 — PTOAS constraints (summary)

See inline block comments at every arch branch in this PR. Highlights:

| Area | A2/A3 (level3) | A5 |
| --- | --- | --- |
| InitMemRef scratch | Materializes ci / narrowing cast / sort32 tmp | Unchanged (PTOAS planner) |
| `tile.ci` | 3-arg + FP32 tmp; forbid dst alias tmp | 2-arg, no scratch insert |
| Narrowing `tile.cast` | Explicit i8 tmp + static views | 1-arg tcvt |
| `tile.sort32` | Explicit tmp + static views when required | 2-arg typical |
| `tile.sel` tmp | UINT32 [1,16] read as cmpmask scratch | ABI operand, unread |
| `tile.sels` tmp | Written for scalar; no overlap mask/src | Unread; may alias |
| `tile.prelu` tmp | UINT8 scratch, static view, active input | ABI unread; dst may alias tmp |
| `tile.col_sum` binary | tmp + static view + isBinary | No static-view bridge |

---

- Drop Flatten/Cast legalization scratch duplication
- Simplify `tensor.scatter` flat-index lowering (UINT32 `[1,16]` sel tmp in scatter paths)
- Docs: `32-init_memref.md`, `34-memory_reuse.md` aligned with above

## Review responses (YunjiQin)

- **Blocking docs:** revert lhs/rhs MemoryReuse logic; docs already correct at mask/tmp only
- **FP16→INT4:** excluded with comment + negative UT
- **UINT32[1,16] on A5:** verified unread ABI; docstrings clarified
- **Non-blocking:** named constants, tcvt branch UT, ci/tcvt/col_sum codegen UT, sort32 `*, tmp=`, ci alias UT strengthened
- **Deferred:** legalize_tile_cast refactor revert, DRY EmitInsOuts, sort32 FP16 ST

## Test plan

- [x] `tests/ut/ir/transforms/test_init_memref.py` — scratch matrix, FP16→INT4 exclusion, tcvt head/tail/rows>255
- [x] `tests/ut/ir/transforms/test_memory_reuse.py::TestForbidOutputAlias` — ci/sel/sels/prelu alias guards
- [x] `tests/ut/codegen/test_pto_codegen_ops.py` — tsels/tprelu/tsort32 + ci/tcvt/col_sum static-view
- [x] pre-commit (local, touched files)
- [x] pypto-lib model CI (local a2a3): prefill/decode csa·hca·swa + moe
- [ ] upstream CI

## Dependencies

- Runtime: simpler **0659e29** on top of #2530 API surface
- PTOAS: **v0.60** via `toolchain/versions.env`


src/ir/op/tile_ops/memory.cpp
```diff
@@ -1523,8 +1537,15 @@ REGISTER_OP("tile.ci")
...
     .add_argument("shape", "Destination shape (TupleType of ConstInt)")
+    .add_argument("tmp", "Optional A2/A3 scratch tile (FP32 Vec)")
     .set_attr<DataType>("dtype")
     .set_attr<bool>("descending")
+    .set_input_memory(2, MemorySpace::Vec)
+    // The A2/A3 PTOAS level3 TCI form takes tmp as an explicit scratch input
+    // and may still read it while producing dst, so MemoryReuse cannot recycle
+    // tmp's allocation for the output. A5 normally uses the tmp-free form, and
+    // InitMemRef never synthesizes this operand for A5.
+    .forbid_output_alias(2)
```
YunjiQin, line 1548: Confirming this resolves the alias hazard I flagged on the previous revision, and the comment states the reason accurately.

For the record on why it works: `tile.ci` is inplace-safe by default (`is_inplace_safe_{true}`), so `ForbidAliasCollector` takes the `else` branch and reads `ForbidOutputAliasArgs()`; `forbid_arg` guards `i < call->args_.size()`, so the A5 2-arg form is unaffected.

src/ir/transforms/init_memref.cpp
```diff
@@ -149,6 +152,144 @@ std::optional<uint64_t> StaticSliceViewSpanBytes(const CallPtr& call, const Shap
...
+
+int64_t CeilDivI64(int64_t num, int64_t den) { return (num + den - 1) / den; }
+
+/// Non-saturating A2/A3 pto.tcvt forms whose ISA implementation needs a tmp.
+/// PTOAS materializes these before PlanMemory at level2. PyPTO and DSA-RP emit
+/// fixed addresses at level3, so the tmp must become an ordinary tile before
+/// this pass initializes MemRefs.
+bool TcvtNeedsLevel3Scratch(DataType src, DataType dst) {
+  if (src == DataType::FP32 && dst == DataType::INT16) return true;
+  return src == DataType::FP16 && (dst == DataType::INT16 || dst == DataType::INT8 || dst == DataType::UINT8);
```
YunjiQin, line 181: **Question:** is `FP16 -> INT4` missing here?

The A2/A3 tcvt adjacency table lists it as a native narrowing edge:

```cpp
// src/backend/910B/backend_910b_handler.cpp
add(DataType::FP16, {DataType::INT4});
```

If PTOAS v0.60's `makeTCvtTmpType` also requires a tmp for `f16 -> i4`, that path silently emits the 1-arg form on A2/A3 and gets rejected by PTOAS. If it does not require one, no change needed — but it would be worth a sentence in the comment saying `INT4` was considered and excluded, so the next reader does not have to re-derive it from the adjacency table.

(The other narrowing edges look correctly out of scope: `FP32 -> FP16/BF16` are float resizes, and `INT32 -> INT16` / `INT64 -> INT32` are integer narrows.)

src/ir/transforms/init_memref.cpp
```diff
@@ -149,6 +152,144 @@ std::optional<uint64_t> StaticSliceViewSpanBytes(const CallPtr& call, const Shap
...
+      << "InitMemRef: A2/A3 non-saturating narrowing tile.cast requires static source shape "
+         "and valid_shape to size its level3 scratch tile";
+
+  const int64_t rows = rows_ci->value_;
+  const int64_t cols = cols_ci->value_;
+  const int64_t src_cols = src_cols_ci->value_;
+  int64_t bytes = 0;
+  if (src_tile->dtype_ == DataType::FP32) {
+    if (rows > 0 && cols > 0) {
+      const int64_t head = int64_t{4} * 64 * std::min<int64_t>(cols / 64, 255);
```
YunjiQin: This block replicates PTOAS v0.60's `makeTCvtTmpType` capacity math. I verified it against the three UT parameter sets and it is arithmetically correct (1024 / 64 / 160 all match), but the coupling is fragile in a way that fails dangerously: if PTOAS changes this formula, an under-sized scratch is an out-of-bounds write, not a compile error.

Two cheap mitigations:

1. Name the constants — `255` is the repeat-count ceiling, `64` / `8` / `32` are block / element / alignment granularities. Right now they read as unexplained literals.
2. Put the PTOAS version *and* function name in the comment (the version is there, the function name is not), so the next `versions.env` bump has something greppable.

The existing UT parameters all land in the `tail` branch — `cols % 64 == 0` (head branch) and `rows > 255` (the clamp) are unexercised. Worth two more parameter rows.

src/ir/transforms/init_memref.cpp
```diff
@@ -149,6 +152,144 @@ std::optional<uint64_t> StaticSliceViewSpanBytes(const CallPtr& call, const Shap
...
+    new_assign->value_ = std::move(new_call);
+    return SeqStmts::Flatten({std::move(scratch_assign), std::move(new_assign)}, op->span_);
+  }
+
+ private:
+  static std::optional<PtoScratchSpec> GetScratchSpec(const CallPtr& call, const Span& span) {
+    if (IsOp(call, "tile.ci") && call->args_.size() == 2) {
+      auto result_type = As<TileType>(call->GetType());
+      INTERNAL_CHECK_SPAN(result_type, span) << "tile.ci result must be TileType before InitMemRef";
+      const int64_t cols = result_type->dtype_.GetBit() == 32 ? 192 : 448;
```
YunjiQin: Same concern as the tcvt capacity constants: `192` and `448` are bare magic numbers. The comment explains *which* one is selected (32-bit vs 16-bit dtype) but not where either value comes from.

A one-line derivation or a PTOAS reference would make this maintainable across the next version bump.

src/ir/transforms/init_memref.cpp
```diff
@@ -825,7 +967,19 @@ FunctionPtr TransformInitMemRef(const FunctionPtr& func) {
...
   }
 
-  // Step 2: Resolve author-declared allocations (`pl.Tile[..., pl.MemRef("name"),
+  // PTOAS level2 owns implicit-tmp materialization as part of PlanMemory. PyPTO
+  // and DSA-RP instead emit fixed addresses and invoke level3, so their A2/A3
+  // scratch tiles must enter the ordinary MemRef pipeline here. This is the last
+  // tile-level boundary: cast legalization and shape/layout rewriting are done,
+  // while every synthesized tile can still participate in reuse and placement.
+  const MemoryPlanner planner = ctx ? ctx->GetMemoryPlanner() : MemoryPlanner::PyPTO;
+  if (handler != nullptr && handler->GetPtoTargetArch() == "a2a3" &&
```
YunjiQin: **Gate inconsistency between the producing and consuming sides.**

Here the materializer is gated on `GetPtoTargetArch() == "a2a3"`, but the consumers are gated on `!= "a5"`:

- `src/backend/common/pto_ops_elementwise.cpp` — `MakeTcvtCodegenPTO`
- `src/backend/common/pto_ops_elementwise.cpp` — `tile.col_sum`

For any third architecture the two disagree: no scratch is materialized, yet a static `valid_shape` is still required. Harmless with exactly two arches today, but it is the kind of latent mismatch that surfaces as a confusing user-facing error when a third one is added.

Suggest picking one form. Cleanest option: put the predicate on `BackendHandler` (e.g. `RequiresLevel3TmpScratch()`), which also addresses the hard-coded arch branch in `memory_reuse_pass.cpp` and satisfies the `pass-context-config` rule that passes not branch on architecture directly.

src/backend/common/pto_ops_elementwise.cpp
```diff
@@ -282,6 +287,40 @@ static std::string MakePrecisionCodegenPTO(const std::string& pto_op_name, size_
...
 }
 
+// The level3 explicit-tmp form verifies tcvt scratch against src capacity and
+// dst valid_shape. alloc_tile types keep v_row=?, v_col=?, so bridge to
+// static-valid views the same way tprelu / tcolsum do.
+static std::string MakeTcvtCodegenPTO(const CallPtr& op, codegen::CodegenBase& codegen_base) {
+  auto& codegen = AsPto(codegen_base);
+  INTERNAL_CHECK_SPAN(op->args_.size() == 1 || op->args_.size() == 2, op->span_)
+      << "tile.cast requires 1 or 2 arguments (src[, tmp]), but got " << op->args_.size();
+  if (op->args_.size() == 2 && codegen.GetBackendHandler()->GetPtoTargetArch() != "a5") {
```
YunjiQin: This `!= "a5"` is the consuming side of the gate mismatch noted on `init_memref.cpp` (materialization uses `== "a2a3"`). Same for the `tile.col_sum` branch below.

Separately: `RequireStaticValidShapeForPtoas` uses `CHECK_SPAN`, i.e. `pypto::ValueError`. Per `error-checking.md` that is the right call *only* if this is a documented user-facing limitation rather than a compiler invariant. The existing `tprelu` tests assert `pytest.raises(ValueError)`, so the classification looks intentional — flagging it only so the choice is deliberate and stays consistent if these guards are extended.

src/ir/transforms/op_conversion_registry.cpp
```diff
@@ -2461,9 +2461,9 @@ void OpConversionRegistry::RegisterCmpOps() {
     auto zero_var = make_full(0.0, "cmp_zero");
 
     std::vector<ExprPtr> tmp_shape_dims = {std::make_shared<ConstInt>(1, DataType::INDEX, span),
-                                           std::make_shared<ConstInt>(32, DataType::INDEX, span)};
+                                           std::make_shared<ConstInt>(16, DataType::INDEX, span)};
     auto tmp_shape_tuple = std::make_shared<MakeTuple>(tmp_shape_dims, span);
-    std::vector<std::pair<std::string, std::any>> tmp_kw = {{"dtype", DataType::UINT8},
+    std::vector<std::pair<std::string, std::any>> tmp_kw = {{"dtype", DataType::UINT32},
```
YunjiQin: **Question: does A5 accept `UINT32[1, 16]`?**

This lowering — and the two `scatter` ones above — are target-independent, so the tmp geometry change from `UINT8[1, 32]` (32 B) to `UINT32[1, 16]` (64 B) reaches A5 as well. But everything else in the PR scopes the new geometry to A2/A3:

- the registry description: `"Scratch tile required by TSEL (TileType UINT32 [1, 16] on A2/A3)"`
- both Python docstrings: `"(UINT32 [1, 16] on A2/A3)"`
- `tests/st/runtime/ops/test_sels.py` still branches: `tmp_dtype or (dtype if platform == "a2a3" else DataType.UINT8)`

If A5's TSEL validates tmp dtype/extent, these three conversions will now produce invalid A5 kernels. If A5 accepts both, no change needed — but then the "on A2/A3" qualifier in the descriptions is misleading and should be dropped.

src/ir/transforms/memory_reuse_pass.cpp
```diff
@@ -1714,6 +1714,11 @@ class ForbidAliasCollector : public IRVisitor {
           for (size_t i = 0; i < forbidden_arg_count; ++i) forbid_arg(i);
         } else {
           for (size_t i : entry.ForbidOutputAliasArgs()) forbid_arg(i);
+          // A2/A3 TSEL also reads lhs/rhs while writing dst; A5 may reuse them.
+          if (IsOp(call, "tile.sel") && !IsA5Target()) {
```
YunjiQin: `.claude/rules/pass-context-config.md` asks that per-backend behaviour live behind a `BackendHandler` virtual rather than an architecture branch inside a pass. `IsA5Target()` is an existing helper in this file (used by `IsA5Prelu`), so this is locally consistent — but it is now the third such branch, and whether TSEL may overwrite a dead value operand is an ISA property, not a MemoryReuse policy.

Suggest `BackendHandler::AllowsSelValueOperandAlias()` (or a more general `TselReadsValueOperandsWhileWriting()`). That would also give the arch gate noted on `init_memref.cpp` a single home.

Non-blocking — the behaviour is correct as written and well covered by the new A2/A3-vs-A5 test pair.

src/ir/transforms/legalize_tile_cast_pass.cpp
```diff
@@ -273,24 +277,34 @@ class LegalizeTileCastMutator : public IRMutator {
...
-    final_assign->value_ = MakeCast(cur, chain.back(), mode, op->span_);
-    stmts.push_back(std::move(final_assign));
+    AppendCastHop(stmts, cur, chain.back(), mode, op, op);
 
     if (stmts.size() == 1) return stmts.front();
     return std::make_shared<SeqStmts>(std::move(stmts), op->span_);
   }
 
  private:
+  void AppendCastHop(std::vector<StmtPtr>& stmts, ExprPtr& cur, DataType hop_dst, int mode,
```
YunjiQin: This extraction has lost its motivation now that scratch materialization moved to InitMemRef, and it reads worse than the code it replaced:

- `final_assign` is a flag parameter selecting between two disjoint bodies.
- The final call passes the same object twice: `AppendCastHop(stmts, cur, chain.back(), mode, op, op)`.
- `cur` is updated on one path and not the other, which the signature does not convey.
- `src_tile` is bound solely to feed one `INTERNAL_CHECK` and is otherwise unused.

Reverting to the previous inline loop + explicit final assignment would shrink the diff and remove a shape the next reader has to decode. Non-blocking.

src/backend/common/pto_ops_internal.h
```diff
@@ -79,6 +81,20 @@ std::string EmitFlatOffsetSSAFromValues(const std::vector<std::string>& indices,
...
+bool HasStaticValidShape(const std::shared_ptr<const ir::TileType>& tile_type);
+void RequireStaticValidShapeForPtoas(const std::shared_ptr<const ir::TileType>& tile_type,
+                                     std::string_view op_name, std::string_view role, const ir::Span& span);
+std::string GetTileViewTypeAnnotation(const ir::ExprPtr& expr, codegen::PTOCodegen& codegen);
+std::string EnsureTileViewSsa(const ir::ExprPtr& expr,
+                              const std::shared_ptr<const ir::TileType>& view_tile_type,
+                              codegen::PTOCodegen& codegen, const std::string& temp_prefix);
+std::string EnsureStaticViewTileSsa(const ir::ExprPtr& expr, codegen::PTOCodegen& codegen,
+                                    const std::string& temp_prefix);
+void EmitInsOutsWithViewTypes(codegen::PTOCodegen& codegen, std::string_view pto_op_name,
```
YunjiQin, line 93: `EmitInsOutsWithViewTypes` and `EmitInsOuts` are ~80% identical; the only real difference is where the result SSA and type come from (explicit arguments vs `GetCurrentResultTarget()` / `GetCurrentResultTileBufTypeString()`).

Per the DRY rule, consider either giving `EmitInsOuts` an optional result override, or factoring a shared `BuildInsOutsClause` that both call. As written, a future fix to the clause format has to be applied twice.

Minor related point: `EnsureTileViewSsa` emits an untyped `pto.treshape` when `GetViewTileBufTypeStringFromTileType` returns empty (a memref-less TileType). Unreachable today since every call site runs after InitMemRef, but since `EnsureStaticViewTileSsa` already `INTERNAL_CHECK`s the TileType, checking `memref_.has_value()` there too would make the invariant explicit.

python/pypto/language/op/tile_ops.py
```diff
@@ -2809,23 +2822,25 @@ def sels(mask: Tile, src: Tile, tmp: Tile, scalar: int | float | Expr | Scalar)
     return Tile(expr=call_expr)
 
 
-def sort32(src: Tile, idx: Tile) -> Tile:
+def sort32(src: Tile, idx: Tile, tmp: Tile | None = None) -> Tile:
```
YunjiQin: `tmp` should be keyword-only here for consistency with the two sibling ops added in this PR:

```python
def ci(..., descending: bool = False, *, tmp: Tile | None = None) -> Tile
def cast(tile, target_type, mode="round", *, tmp: Tile | None = None) -> Tile
def sort32(src: Tile, idx: Tile, tmp: Tile | None = None) -> Tile   # <- missing `*`
```

The `ir.op` layer already has it keyword-only (`def sort32(src, idx, span=None, *, tmp=None)`), so only the language layer diverges. As written, `pl.tile.sort32(a, b, c)` positionally accepts an operand documented as "Normally compiler-generated".

tests/ut/codegen/test_pto_codegen_ops.py
```diff
@@ -2903,6 +2906,40 @@ def kernel(
         assert tmrgsort_lines, "No pto.tmrgsort line found"
         assert "i32" in tmrgsort_lines[0], f"block_len type annotation should be i32: {tmrgsort_lines[0]}"
 
+    @pytest.mark.parametrize(
+        ("src_dtype", "output_cols"),
+        [(pl.FP32, 128), (pl.FP16, 256)],
+    )
+    def test_sort32_dynamic_valid_width_emits_level3_scratch(self, src_dtype, output_cols):
```
YunjiQin, line 2913: This test is a good model for the pattern — but the same treatment is missing for the other three codegen branches this PR adds:

- `MakeCiCodegenPTO`, 3-argument form (`ci_tmp_view` / `ci_dst_view`)
- `MakeTcvtCodegenPTO`, 2-argument form (`tcvt_src_view` / `tcvt_tmp_view` / `tcvt_dst_view`)
- `tile.col_sum`, static-view branch (`colsum_*_view`)

These are exactly the paths most likely to drift on the next PTOAS bump, and none of them is currently exercised: the file's other changes cover `tsels`, `tprelu`, `tsort32` and `syncall` only. Three tests shaped like this one (assert the `%*_view` operands appear and that no `v_row=?` / `v_col=?` survives) would close the gap.

Separately, `sort32` FP16 now produces a 4x-wide output — a behaviour change — but `tests/st/runtime/ops/test_sort.py` is FP32-only. The type inference is unit-tested; an FP16 ST would be worth adding since the previous 2x shape under-allocated.

tests/ut/ir/transforms/test_memory_reuse.py
```diff
@@ -5037,17 +5037,55 @@ def test_ascend950_allows_load_buffer_reuse(self):
...
-    via OpRegistryEntry::forbid_output_alias(); MemoryReuse honours the marker
-    even when shape/dtype would otherwise permit the reuse.
+    These constraints live in the op registry so MemoryReuse can distinguish
+    operands whose lifetimes end at an op from operands safe for its output to
+    overwrite. The target-specific value-operand exception for TSEL remains in
+    MemoryReuse so A5 may reuse a dead lhs/rhs buffer.
     """
 
+    def test_ci_output_does_not_alias_compiler_scratch(self):
+        """A2/A3 TCI reads its compiler-owned scratch while writing dst."""
```
YunjiQin: Please confirm this test actually fails with `.forbid_output_alias(2)` reverted — otherwise it asserts a fact that happens to hold rather than guarding the fix.

The kernel here is minimal (a single `tile.ci`, no other allocations), whereas the sibling `test_sel_output_does_not_alias_mask_or_tmp` deliberately stages a dead 1024 B FP32 buffer to force the reuse decision. It is plausible that the scratch is the only reuse candidate and the test is genuinely load-bearing — just worth verifying rather than assuming.

YunjiQin (commented):
## Review — PTOAS v0.60 upgrade + A2/A3 level-3 scratch ownership

Reviewed at `c21688309`. The direction is right: InitMemRef is the correct — and only — insertion point for compiler-owned scratch (cast legalization and shape/layout rewriting are done, MemRef collection has not started), and the code comments make that reasoning explicit. `src/ir/op/sort_utils.h` correctly de-duplicates the tile/tensor sort32 width derivation.

### Resolved in this revision

`tile.ci` now carries `.forbid_output_alias(2)` plus a MemoryReuse regression test. The fix is correct: `tile.ci` is inplace-safe by default, so `ForbidAliasCollector` takes the `else` branch and reads `ForbidOutputAliasArgs()`; `forbid_arg` guards `i < args_.size()`, so the 2-arg A5 form is unaffected.

### Blocking

**Docs are out of sync.** `docs/en/dev/passes/34-memory_reuse.md` (L94, L278) and `docs/zh/dev/passes/34-memory_reuse.md` (L90, L265) still describe `tile.sel` as `forbid_output_alias(0)` (mask) + `(3)` (tmp) only. This PR adds an A2/A3-gated lhs/rhs restriction in MemoryReuse, which those four places must reflect. Small change, but required by the project documentation rule.

### Open questions (inline)

Two findings are questions rather than assertions — if PTOAS v0.60 behaviour has already been checked, no code change is needed. Both would fail silently at runtime if the assumption is wrong:

- `FP16 -> INT4` is a native A2/A3 narrowing edge but is not in `TcvtNeedsLevel3Scratch`.
- The `tile.sel` tmp change to `UINT32[1, 16]` lands in three target-independent lowerings, so A5 gets it too.

### Risk note

The toolchain pin is not independently revertible: the soft `syncall` ABI, `tfillpad` rename, TSEL tmp geometry and the level-3 tmp forms are all bound to v0.60. Rolling back `toolchain/versions.env` requires rolling back those codegen changes together. Worth stating in the PR description.

### Test coverage

Strong on the IR side — the InitMemRef materialization matrix (arch x planner x explicit-tmp preservation) and the sort32 dtype/symbolic-valid cases are thorough, and the MemoryReuse A2/A3-forbid vs A5-allow pair is a good contrast. The gap is on the codegen side; see the inline comment on `tests/ut/codegen/test_pto_codegen_ops.py`.

Remaining inline comments are non-blocking suggestions (naming constants, gate consistency, a refactor that lost its motivation, DRY, one signature inconsistency).

src/ir/op/tile_ops/memory.cpp
yanghaoran29: Acknowledged — no further change.

yanghaoran29:
Thanks for the thorough review @YunjiQin — addressed below.

**Blocking (tile.sel docs):** I agree the docs/code were out of sync, but the fix is the opposite direction: the A2/A3 lhs/rhs forbid in `MemoryReuse` was incorrect and is **reverted** in `b3a2f4073`. `34-memory_reuse.md` already documents the right contract (`forbid_output_alias(0)` + `(3)` only); registry + UT now match (dead lhs/rhs may be reused on both A2/A3 and A5 when lifetimes allow).

**Open questions:**
- **FP16→INT4:** Checked pto-isa `TCvt.hpp` — not in `kIsNarrowingCvt`; FP16→S4 uses direct `vconv_f162s4*` without tmp. Added an exclusion comment in `TcvtNeedsLevel3Scratch` + negative InitMemRef UT.
- **UINT32[1,16] on A5:** PTOAS a5 `tsel.py` does not read `tmp` (`_ = tmp`); ptodsl omits tmp on a5 via `dst` placeholder. UINT32[1,16] is safe as an unread ABI operand. Docstrings clarify "required on A2/A3; unread on A5".

**Risk note:** PR description updated to state toolchain v0.60 is not independently revertible.

**Non-blocking:** Named tcvt/ci constants + PTOAS refs; tcvt head/tail/rows>255 UT; ci/tcvt/col_sum codegen static-view UT; `sort32(..., *, tmp=)` keyword-only; strengthened ci alias UT; **`RequiresLevel3TmpScratch()` + op_conversion arch-gated TSEL tmp** (not global UINT32 on A5). Deferred: legalize_tile_cast revert, DRY `EmitInsOuts`, sort32 FP16 ST.

src/ir/transforms/init_memref.cpp
yanghaoran29: Good catch. Verified against pto-isa: FP16→INT4 is **not** a non-saturating narrowing tcvt tmp path (`kIsNarrowingCvt` excludes int4; `GenCastCallFp16ToS4` has no tmp). Added comment + `test_fp16_to_int4_has_no_level3_scratch`.

src/ir/transforms/init_memref.cpp
yanghaoran29: Agreed. Named constants + `makeTCvtTmpType` in comment; added UT rows for head-only (`cols % 64 == 0`) and rows>255 tail clamp.

src/ir/transforms/init_memref.cpp
yanghaoran29: Added `kA2A3CiScratchColsInt32` / `Int16` with PTOAS TCI tmp reference in comment.

src/ir/transforms/init_memref.cpp
yanghaoran29: Added `BackendHandler::RequiresLevel3TmpScratch()` (910B=true, 950=false); init + tcvt/col_sum codegen now share one predicate. PassContext-first handler lookup.

src/backend/common/pto_ops_elementwise.cpp
yanghaoran29: Same fix: `RequiresLevel3TmpScratch()` replaces `!= "a5"` here; aligned with InitMemRef materializer gate.

src/ir/transforms/op_conversion_registry.cpp
yanghaoran29: Scatter/cmp lowering now **arch-gated** via `RequiresLevel3TmpScratch()` (UINT32[1,16] on 910B only; A5 keeps prior UINT8[1,32]). A5 PTOAS ignores unread tmp. Docstrings updated.

src/ir/transforms/memory_reuse_pass.cpp
yanghaoran29: **Reverted** per YunjiQin #2106 principle — alias rules belong in OpRegistry, not pass arch hacks. Docs at mask/tmp only.

src/ir/transforms/legalize_tile_cast_pass.cpp
yanghaoran29: Fair — motivation weakened after InitMemRef move. **Deferred** this PR to limit diff; can revert inline loop in a small follow-up.

src/backend/common/pto_ops_internal.h
yanghaoran29: Agreed in principle. **Deferred** — static-view bridge is new; DRY refactor is separate cleanup.

python/pypto/language/op/tile_ops.py
yanghaoran29: Fixed: `def sort32(src, idx, *, tmp=None)`.

tests/ut/ir/transforms/test_memory_reuse.py
yanghaoran29: Verified: reverting `.forbid_output_alias(2)` lets dst alias tmp in minimal kernel. Strengthened test doc/assertions in push.

tests/ut/codegen/test_pto_codegen_ops.py
yanghaoran29: Added three tests mirroring sort32 static-view pattern. sort32 FP16 ST deferred (onboard follow-up).

src/ir/transforms/legalize_tile_cast_pass.cpp
yanghaoran29: Done in latest push: reverted `AppendCastHop` to the previous inline hop loop; only the header comment remains (scratch is materialized in InitMemRef). Behavior unchanged.

YunjiQin (approved):

PR #2567, opened 2026-08-30
refactor(ir): let TensorType canonicalize the cross-core boundary view

Hzfengsy:
## What changed

`DeduceSplitReshapeTensor` — the type deducer for `tensor.aiv_shard` / `tensor.aic_gather` — branched on whether the reshaped `valid_shape` equalled the reshaped shape, attaching a `TensorView` only in the partial case. The branch is dead. It is removed, along with the comment that justified it, and the now-unused `tile_view_semantics.h` include.

```cpp
-  // ... TileType's constructor drops a tile_view whose valid_shape matches the
-  // shape (the implicit view), but TensorType performs no such canonicalization.
-  if (tile_view_semantics::ShapeExprListsEquivalent(reshaped.valid, reshaped.shape)) {
-    return std::make_shared<TensorType>(std::move(reshaped.shape), dtype, std::nullopt);
-  }
   TensorView tensor_view({}, TensorLayout::ND, std::move(reshaped.valid));
   return std::make_shared<TensorType>(std::move(reshaped.shape), dtype, std::nullopt,
                                       std::make_optional(std::move(tensor_view)));
```

## Why

The comment's premise was true when written and stopped being true twelve days later:

| Date | Commit | |
|---|---|---|
| 2026-05-03 | `f21c2dd4` (#1266) | `TileType` gains `CanonicalizeTileViewInPlace` — **tile side only** |
| 2026-07-02 | `82083eb7` (#1923) | `tensor.aiv_shard` / `aic_gather` added. The author correctly observes `TensorType` has no mirror, hand-rolls the guard, and records the asymmetry in a comment |
| 2026-07-14 | `3079ba55` (#2031) | `TensorType` gains `CanonicalizeTensorViewInPlace` — asymmetry closed. Does not touch `cross_core.cpp` |

`TensorType`'s constructor now calls `CanonicalizeTensorViewInPlace`, which clears a `valid_shape` equal to the shape and then, finding all four fields at their defaults, resets the view to `nullopt`. And the two predicates are the same function: `ShapeExprListsEquivalent` and `AreExprVectorsEqual` are both a size check plus element-wise `AreExprsEqual`. The guard's condition *is* the constructor's condition, so it could never select a different outcome — both arms built an identical `TensorType`.

The tell is right next door: the tile deducer directly above has no such guard, because on the tile side the constructor had done this since #1266.

The stale comment was the actual cost. It asserts an invariant of the type system that no longer holds, in the file a reader would trust on the subject, and teaches the next author to hand-roll the same thing.

## Behaviour

None. At the DSL level, a fully-valid operand (`pl.aiv_shard(acc)` where `acc` comes straight from `pl.matmul`) took the guarded arm; a narrowed one (`pl.slice(..., valid_shape=[v, K])` feeding the matmul) took the other. Both now go through one path and land on the same type. The tensor-level view is transient regardless: `ConvertTensorToTileOps` re-runs the *tile* deducer via `OpRegistry::Create`, so this `tensor_view_` never reaches codegen.

## Verification

Deduced types compared before vs. after, guard present vs. removed — identical:

| operand | result |
|---|---|
| `Tensor[[16,128]]` | `[8,128]`, `view=None` |
| `Tensor[[16,128]]` + explicit full view | `[8,128]`, `view=None` |
| `Tensor[[16,128]]` valid `[16,100]` | `[8,128]`, `valid_shape=[8,100]` |

- Cross-core / split_aiv / tensor-view-semantics suites: **489 passed**
- Full `tests/ut/`: **10597 passed**, 3 skipped, 3 xfailed
- Lint: `check_english_only`, `check_headers`, `check_op_name_literals`, `check_emitter_check_classification` all pass; full pre-commit hook set green
- No docs referenced the removed claim

## For the reviewer

The load-bearing question is whether the guard's predicate and the constructor's are genuinely the same. `ShapeExprListsEquivalent` ([tile_view_semantics.h:84](https://github.com/hw-native-sys/pypto/blob/main/include/pypto/ir/tile_view_semantics.h#L84)) and `AreExprVectorsEqual` ([expr.cpp:116](https://github.com/hw-native-sys/pypto/blob/main/src/ir/expr.cpp#L116)) each do a size check then element-wise `AreExprsEqual` — no difference in strength, so canonicalization clears exactly the cases the guard caught.

### DIFF
--- a/src/ir/op/tile_ops/cross_core.cpp
+++ b/src/ir/op/tile_ops/cross_core.cpp
@@ -27,7 +27,6 @@
 #include "pypto/ir/scalar_expr.h"
 #include "pypto/ir/span.h"
 #include "pypto/ir/stmt.h"
-#include "pypto/ir/tile_view_semantics.h"
 #include "pypto/ir/transforms/printer.h"
 #include "pypto/ir/type.h"
 #include "pypto/ir/type_inference.h"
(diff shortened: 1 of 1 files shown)


Hzfengsy:
Addressed the outside-diff finding on `src/ir/op/tile_ops/cross_core.cpp:305-307` in 17c2f44c — the `split=0` comment no longer points at the deleted halving-path guard; it now states why a `TensorType` operand is returned unchanged (no memory space / tile layout to strip, view already canonicalized at construction). Replying here rather than on a thread since the finding was posted outside the diff and has no inline thread.

lyfne123 (approved):

PR #2590, opened 2026-08-31
fix(runtime): forward ring config through execute_compiled

### LINKED ISSUE #2586: [Feature] execute_compiled cannot pass ring sizing (ring_heap) to CallConfig.runtime_env
## Background

**pypto-lib-side problem.** `models/deepseek_v4_pro/prefill_attention_csa.py`,
`prefill_attention_hca.py` and `prefill_attention_swa.py` fail on device with:

```
[ERROR] validate_runtime_impl: [runtime_maker.cpp:945] simpler runtime failed:
        orch_error_code=2 sched_error_code=0 runtime_status=-2
[ERROR] error detail: orch_error_code=2 HEAP_RING_DEADLOCK -
        the task allocator could not reserve the heap bytes required for another task
```

These are single-chip (L2) kernels. Their per-task args and intermediates
exceed the runtime's compile-time default output heap
(`CHIP_HEAP_SIZE`, 256 MiB per ring —
simpler `src/a5/runtime/tensormap_and_ringbuffer/runtime/runtime_types.h:72`).
Measured on device: **512 MiB per ring is already enough for all three**.

The runtime exposes exactly the knob needed —
`CallConfig.runtime_env.ring_heap` — and pypto surfaces it as
`RunConfig.ring_heap`. **There is no way to reach it from an L2 dispatch
driven by `execute_compiled`.**

### Reproduction environment

| Component | Version |
|---|---|
| pypto | `750131ca` |
| simpler (pypto `runtime` gitlink) | `dbdd041e` |
| pto-isa (`runtime/pto_isa.pin`) | `be5ccb76` |
| ptoas (`toolchain/versions.env`) | `v0.60` |
| pypto-lib | `4e488fb` |
| Platform / runtime | `a5` (Ascend 950), `tensormap_and_ringbuffer` |

Repro: `DEEPSEEK_V4_VARIANT=pro python models/deepseek_v4_pro/prefill_attention_swa.py -p a5 -d <dev>`

---

## Summary

`execute_compiled` cannot carry per-task ring sizing
(`ring_task_window` / `ring_heap` / `ring_dep_pool`), so an L2 dispatch driven
by it always runs at the runtime's compile-time defaults. Please make
`execute_compiled` able to pass a `RunConfig` (or at minimum the ring fields)
through to `CallConfig.runtime_env`.

## Motivation / Use Case

pypto has **two parallel L2 dispatch surfaces**, and ring sizing exists on only
one of them.

**Surface A — the `work_dir` surface (`execute_compiled`).** No ring parameter:

```python
def execute_compiled(work_dir, args, *, platform, device_id, dfx=_DfxOpts(),
                     level=2, aicpu_thread_num=None,
                     analyze_auto_scopes_for_deps=False) -> None
```

It calls `compile_and_assemble(work_dir, platform)` and then
`execute_on_device(...)`, which builds a fresh `CallConfig()` and sets exactly
`aicpu_thread_num`, the five DFX flags and `output_prefix` — and nothing else:

```python
cfg = CallConfig()
if aicpu_thread_num is not None:
    cfg.aicpu_thread_num = aicpu_thread_num
cfg.enable_chip_swimlane = enable_chip_swimlane
cfg.enable_dump_args = enable_dump_args
cfg.enable_pmu = enable_pmu
cfg.enable_dep_gen = enable_dep_gen
cfg.enable_scope_stats = enable_scope_stats
if output_prefix:
    cfg.output_prefix = output_prefix
```

`cfg.runtime_env` occurs **zero times** in `python/pypto/runtime/device_runner.py`,
and `ring_heap` / `ring_task_window` / `ring_dep_pool` occur zero times in both
`device_runner.py` and `ir/compiled_program.py` (checked at `750131ca`).

**Surface B — the `RunConfig` / `CompiledProgram` surface.** `ring_heap` is a
`RunConfig` field and reaches the device through a single application point:

```
RunConfig(ring_heap=...)
  -> compiled.build_call_config(rc)
    -> _build_call_config(...)                    # runtime/runner.py:979
      -> _apply_ring_overrides(cfg, rc)           # called at runner.py:1019,
        -> cfg.runtime_env.ring_heap = ...        #   defined at runner.py:957
```

Only three entry points reach it: `ChipWorker.run(compiled, *args, config=rc)`,
`benchmark(compiled, args, config=rc)`, and the L3 path
(`distributed_runner._make_call_config`).

Note `compiled(*args, config=rc)` is **not** one of them. `_invoke_compiled`
(`ir/compiled_program.py:658`) already holds a `RunConfig` and hands it to
`execute_compiled` — but only four fields survive the call:

```python
execute_compiled(
    output_dir,
    coerced,
    platform=execution_platform,
    device_id=config.device_id,
    dfx=_DfxOpts.from_run_config(config),
    aicpu_thread_num=config.aicpu_thread_num,
)
```

The ring fields are dropped here silently. This is also why the fix below is
natural: the `RunConfig` is already in hand at the call site, it just has
nowhere to go.

**Why this only became a problem recently.** Ring sizing used to be the
`PTO2_RING_*` environment variables, which are process-global and therefore
worked on *both* surfaces. When simpler retired them and sizing moved onto
`CallConfig.runtime_env`, only Surface B was rewired. Surface A was left
behind, and with the env route gone an L2 `execute_compiled` dispatch now has
**no** way to size its rings.

**Why this matters to pypto-lib specifically.** `run_jit(runtime_cfg=...)` is
deliberately designed as a **pass-through to `execute_compiled`'s keyword
arguments** — no lib-owned field vocabulary, so pypto-lib does not have to
re-adapt every time the runtime surface changes. A knob that exists in
`RunConfig` but not on `execute_compiled` breaks that contract: the lib either
carries a custom field with custom routing, or the kernel simply cannot ask for
the knob. We would like to keep `runtime_cfg` a pure pass-through.

Measured today at the last Python frame before the mailbox
(`simpler.worker.Worker.run`), an L2 dispatch through `execute_compiled`
receives `ring_heap = [0, 0, 0, 0]`, i.e. "fall through to the compile-time
default", with no way for the caller to change it.

## Proposed API / Behavior

**Preferred — let `execute_compiled` take a `RunConfig`:**

```python
def execute_compiled(work_dir, args, *, platform, device_id, dfx=_DfxOpts(),
                     level=2, aicpu_thread_num=None,
                     analyze_auto_scopes_for_deps=False,
                     config: RunConfig | None = None) -> None
```

forwarded to `execute_on_device`, which applies it with the existing
`_apply_ring_overrides` — the same code path Surface B already uses, so both
surfaces stay in step by construction.

This is preferred over adding three individual keywords because it makes the
two surfaces converge permanently: **any future `RunConfig` knob then reaches
L2 with no pypto-lib change at all**, which is exactly the property
`runtime_cfg` was designed to have.

**Minimum acceptable** — three explicit keywords, forwarded the same way:

```python
execute_compiled(..., ring_task_window=None, ring_heap=None, ring_dep_pool=None)
```

**Secondary, same area.** `execute_on_device` already has a parameter literally
named `runtime_env`:

```python
def execute_on_device(..., runtime_env: dict[str, str] | None = None) -> None
```

but it is **OS environment variables** applied through `_temporary_env()`, not
`CallConfig.runtime_env`. The collision reads exactly like the surface being
asked for here, and since the `PTO2_RING_*` knobs were retired, setting it has
no effect on ring sizing at all. Renaming it (e.g. `os_env` / `env_overrides`)
would remove a real trap.

## Alternatives Considered

1. **Environment variables** — gone. `PTO2_RING_TASK_WINDOW` / `PTO2_RING_HEAP`
   / `PTO2_RING_DEP_POOL` were retired in simpler.
2. **`compiled(*args, config=RunConfig(ring_heap=...))`** — silently ineffective;
   `_invoke_compiled` drops the ring fields (`ir/compiled_program.py:658`).
3. **Route pypto-lib's L2 dispatch through `ChipWorker` instead** — works, and is
   what pypto-lib PR
   [#1089](https://github.com/hw-native-sys/pypto-lib/pull/1089) currently does
   as a stopgap. We consider it the wrong layer: it forces pypto-lib to own a
   ring-key vocabulary and a second dispatch route, which is precisely what
   `runtime_cfg`-as-pass-through was meant to avoid. It also loses
   `execute_compiled`'s dep-capture subprocess, so that route has to refuse
   swimlane / dep_gen runs. **We intend to revert it once `execute_compiled`
   carries the knob.**
4. **Raising the compile-time default** — a global change that costs every
   program device memory for the benefit of a few.

## Additional Context

Evidence that this is ordinary under-provisioning and nothing structural —
injected at `simpler.worker.Worker.run`, all outputs validated:

| Case | `ring_heap` (all four rings) | Result |
|---|---|---|
| `prefill_attention_swa` | default (`[0,0,0,0]` -> 256 MiB) | HEAP_RING_DEADLOCK |
| `prefill_attention_swa` | 512 MiB | PASS |
| `prefill_attention_swa` | 1 GiB | PASS |
| `prefill_attention_csa` | 512 MiB / 1 GiB | PASS |
| `prefill_attention_hca` | 512 MiB / 1 GiB | PASS |

One doubling of the default is enough. A stale comment in our tree claimed
4 GiB on all four rings still deadlocked; that measurement went through a route
which dropped the value, and does not reproduce.

Related: the ring sizing must also cover the **benchmark** dispatch, which is a
second, independent launch — sizing only the correctness run moves the deadlock
rather than removing it. `benchmark()` already accepts `config=`, so this is
only a note for whoever wires `execute_compiled`.

Little-oil:
## Summary

- let execute_compiled accept and forward an optional RunConfig
- apply ring_task_window, ring_heap, and ring_dep_pool to CallConfig.runtime_env before worker prewarm and dispatch
- preserve the same ring sizing in CompiledProgram, replay, and the dependency-capture subprocess
- document the L2 work-dir API in English and Chinese and add regression coverage

## Compatibility and precedence

Existing explicit execute_compiled arguments remain authoritative for platform, device_id, DFX, and aicpu_thread_num. The forwarded config supplies the previously unreachable ring-sizing fields, so existing callers retain their behavior.

## Testing

- native RelWithDebInfo worktree build completed successfully
- full unit suite: 10849 passed, 63 skipped, 1 xfailed
- targeted runtime/IR regression suite: 245 passed, 5 skipped
- pre-commit hooks, Ruff, formatting, docs parity, Markdown, and Pyright passed

A real device dispatch was not run because the user-site _task_interface binding in this workspace is stale relative to the pinned Simpler checkout; the host-side tests cover forwarding, CallConfig overlay behavior, serialization, and child-process reconstruction.

Fixes #2586

### DIFF
--- a/docs/en/dev/05-runtime-ring-sizing.md
+++ b/docs/en/dev/05-runtime-ring-sizing.md
@@ -4,7 +4,7 @@ PyPTO exposes Simpler's per-task ring sizing as three optional overrides on
 [`RunConfig`](../../../python/pypto/runtime/runner.py). They let you size the
 runtime's per-task ring resources for a single dispatch without touching the
 compiled artifact or any global state. This works on both the L2 single-chip
-path (`run()` / `ChipWorker.run()`) and the L3 distributed path
+path (`run()` / `execute_compiled()` / `ChipWorker.run()`) and the L3 distributed path
 (`DistributedWorker.run()` / the one-shot `compiled(...)`).
 
 The runtime keeps its task-launch resources in *ring buffers*. Each override
@@ -57,7 +57,7 @@ than raising an opaque `TypeError` from the power-of-two check.
 
 ## Usage
 
-### L2 single-chip (`run`)
+### L2 single-chip (`run` / `execute_compiled`)
 
 ```python
 from pypto.runtime import run, RunConfig
@@ -73,6 +73,26 @@ compiled = run(
     ),
 )
 ```
+
+The directory-driven entry point accepts the same per-dispatch config. Its
+existing explicit `platform`, `device_id`, DFX, and AICPU arguments keep their
+current precedence; `config` carries the ring overrides that previously had no
+route to `CallConfig.runtime_env`:
+
+```python
+from pypto.runtime import RunConfig, execute_compiled
+
+execute_compiled(
+    work_dir,
+    [a, b, c],
+    platform="a2a3",
+    device_id=0,
+    config=RunConfig(platform="a2a3", ring_heap=512 * 1024 * 1024),
+)
+```
+
+The same ring sizing is used for worker prewarm, the device dispatch, and the
+dependency-capture subprocess used by onboard swimlane collection.
 
 ### L3 distributed (`DistributedWorker` / `compiled(...)`)
 
--- a/docs/en/dev/05-runtime-ring-sizing.md
+++ b/docs/en/dev/05-runtime-ring-sizing.md
@@ -18,9 +18,15 @@ applied per dispatch on top of the program's `DistributedConfig` baseline
 
 | `RunConfig` field | `CallConfig.runtime_env` member | Controls | Constraint |
 | ----------------- | ------------------------------- | -------- | ---------- |
-| `ring_task_window: int \| None` | `ring_task_window` | Number of in-flight task slots in the task ring | power of 2, `>= 4` |
-| `ring_heap: int \| None` | `ring_heap` | Bytes of the per-ring task-output heap | power of 2, `>= 1024` |
-| `ring_dep_pool: int \| None` | `ring_dep_pool` | Dependency-edge pool capacity | `[4, INT32_MAX]` |
+| `ring_task_window: int \| list[int] \| tuple[int, ...] \| None` | `ring_task_window` | Number of in-flight task slots in the task ring | power of 2, `>= 4` |
+| `ring_heap: int \| list[int] \| tuple[int, ...] \| None` | `ring_heap` | Bytes of the per-ring task-output heap | power of 2, `>= 1024` |
+| `ring_dep_pool: int \| list[int] \| tuple[int, ...] \| None` | `ring_dep_pool` | Dependency-edge pool capacity | `[4, INT32_MAX]` |
+
+Each field accepts either a scalar, which is broadcast to every scope-depth
+ring, or a list/tuple of exactly four entries for rings 0 through 3. In the
+per-ring form, `0` leaves that ring at its runtime default; every nonzero entry
+must satisfy the constraint in the table. Scalar `0` is invalid—use `None` to
+leave the entire field unset.
 
 `None` (the default) leaves the field **unset** (`0` on `CallConfig`). PyPTO
 writes the value into `CallConfig.runtime_env` only when it is not `None`, so an
--- a/docs/zh/dev/05-runtime-ring-sizing.md
+++ b/docs/zh/dev/05-runtime-ring-sizing.md
@@ -16,9 +16,14 @@ Simpler 的 `CallConfig.runtime_env` 上的同名字段一一对应，并且按 
 
 | `RunConfig` 字段 | `CallConfig.runtime_env` 成员 | 作用 | 约束 |
 | ---------------- | ----------------------------- | ---- | ---- |
-| `ring_task_window: int \| None` | `ring_task_window` | task ring 中在途 task slot 的数量 | 2 的幂，`>= 4` |
-| `ring_heap: int \| None` | `ring_heap` | 每个 ring 的 task 输出堆字节数 | 2 的幂，`>= 1024` |
-| `ring_dep_pool: int \| None` | `ring_dep_pool` | 依赖边池容量 | `[4, INT32_MAX]` |
+| `ring_task_window: int \| list[int] \| tuple[int, ...] \| None` | `ring_task_window` | task ring 中在途 task slot 的数量 | 2 的幂，`>= 4` |
+| `ring_heap: int \| list[int] \| tuple[int, ...] \| None` | `ring_heap` | 每个 ring 的 task 输出堆字节数 | 2 的幂，`>= 1024` |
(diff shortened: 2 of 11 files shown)


Little-oil:
Addressed the actionable CodeRabbit documentation feedback in 71a35d8e:

- all three ring fields now document their actual scalar/list/tuple types
- the docs specify scalar broadcast, the exact four-entry per-ring form, the per-ring zero sentinel, and scalar-zero rejection
- English and Chinese contracts remain aligned

I also corrected ring_heap, although the review text suggested leaving it unchanged, because RunConfig validates the same per-ring list/tuple form for ring_heap as for the other two fields.

The docstring-coverage item is a non-blocking heuristic warning rather than a repository CI failure. Its denominator includes existing test functions and local fake helpers; adding docstrings to those would be unrelated churn. Repository doc checks, Markdown, pre-commit, Pyright, and all functional CI checks passed on the original revision.

Little-oil:
CI retry result on no-op commit e2acfc19: unit-tests passed, including the previously failing worker-memory test, and every project test job is green. The sole remaining failure is onboard-tests-a5 job 99482930507, which stopped before setup/tests during actions/checkout with `fatal: bad object refs/heads/main` on runner npu-a5-2. This is runner-local checkout state, not a PR regression. Please rerun only that job with `gh run rerun --job 99482930507 -R hw-native-sys/pypto` (WRITE permission required).

lyfne123 (approved):

Issue #2242, opened 2026-07-31
Ring allreduce: unaligned-data handling (dcci flush tail + 32B transfer alignment)


georgebisbas:
Two alignment gaps in the ring allreduce kernel template that the mesh allreduce kernel already handles but the ring kernel currently does not.

### 1. dcci flush loop misses the chunk tail

**File:** `python/pypto/runtime/builtins/collectives/allreduce_ring/templates/kernel.cpp.in`

The dcci flush loop in both the Reduce-Scatter and AllGather phases steps by 16 elements (`i += 16`), which only covers the full chunk when `chunk_elems % 16 == 0`. Compile-time validation only guarantees `numel % NR == 0`, so a tail cache-line spanning the chunk end goes unflushed.

**Counter-example:** `numel = 18, NR = 2` → `chunk_elems = 9` (36 bytes). Chunk 1 starts at byte 36 and spans two cache lines; the 16-element stride misses the second one.

### 2. 32-byte transfer alignment is missing

The mesh allreduce kernel explicitly rounds ragged spans up to `kTransferAlignmentBytes` = 32 bytes (in `collectives/allreduce/templates/kernel.cpp.in`), aligning `tile_elems` to avoid partial-vector load/store. The ring kernel feeds `tile_elems` raw into `ShapeDyn` / `StrideDyn` with no such rounding.

### 3. `kRecvTileDebugValue` rename

The constant name at the top of the kernel reads like leftover debug scaffolding. The `TASSIGN` call is legitimate (it binds `recv_tile` to its buffer as a placeholder, overwritten by the subsequent `TLOAD`), but the constant should be renamed to something like `kRecvTileInitFill` with a one-line comment.

### Suggested approach

- Extract a shared alignment helper used by both mesh and ring kernel templates.
- For the dcci flush: step by 16 elements but emit one additional flush for the final partial cache-line.
- Add an ST case where `chunk_elems % 16 != 0` (e.g. `numel=18, NR=2`).
- Rename `kRecvTileDebugValue` → `kRecvTileInitFill`.

### Context

Raised in review of PR #2094 by @YunjiQin.

georgebisbas:
Status check against `main` (`08d3b231`):

- Item 3 (`kRecvTileDebugValue` → `kRecvTileInitFill` rename) is already done — it landed with #2094 itself, so no separate action needed there.
[...]

PR #2603, opened 2026-09-01
feat(distributed): support ragged (non-divisible) lengths in HOST ring allreduce

georgebisbas:
## Summary

Closes the HOST-ring ragged / arbitrary-length leftover of issue #2242.

The host builtin ring allreduce previously required a statically-known src shape with `numel % NR == 0`. This PR relaxes both enforcement points and makes the kernel partition each rank's buffer into **NR balanced, potentially ragged chunks at runtime** — chunk `r` spans `[floor(numel·r/NR), floor(numel·(r+1)/NR))` — mirroring the InCore composite ring (#2161).

## Changes

- **`CheckRingChunkConstraints`** (`src/ir/transforms/lower_host_tensor_collectives_pass.cpp`): dropped the static-shape + `src_numel % NR == 0` CHECK_SPAN. Only the distributed-type internal check remains, plus a comment documenting the runtime ragged partition.
- **Ring builtin deducer** (`src/ir/op/distributed/collective.cpp`): dropped the compile-time static-shape + `numel % nr` validation block (the `kMaxSupportedRanks` check stays).
- **`allreduce_ring` kernel template** (`python/pypto/runtime/builtins/collectives/allreduce_ring/templates/kernel.cpp.in`): added `RingChunkStart()`; the soft guard no longer rejects non-divisible `numel`; `max_chunk_elems = ceil(numel/nranks)`; `tile_cols = min(max_chunk_elems, kTileCount)`; both RS/AG loops iterate ragged send spans, with each TPUT transfer narrowed to the chunk's exact extent via the staging tile's valid shape (`ColMaskInternal`).
- **ST** (`tests/st/distributed/test_l3_host_tensor_allreduce_ring.py`): new ragged test `{1, 17, 4097, 65537}` × P=2/P=4 (all non-divisible by both P=2 and P=4; 65537 > UB); stage-in/out chunked (`STAGE_CHUNK=8192`, 32-byte `[8,1]` tile for `SIZE=1`, stage cols rounded to 32-byte row alignment).
- **UT** (`tests/ut/ir/transforms/test_lower_host_tensor_collectives.py`): `test_host_allreduce_ring_rejects_nondivisible_numel` → `test_host_allreduce_ring_accepts_nondivisible_numel` (SIZE=27, P=4).

## Verification

| Gate | Result |
|---|---|
| UT `test_lower_host_tensor_collectives.py` | **51/51** |
| a2a3sim `test_l3_host_tensor_allreduce_ring.py` (uniform + ragged + signal-reuse) | **12/12** |
| NPU P=2 (devices 4-5) | **6/6** |
| NPU P=4 (devices 4-7) | **12/12** |
| pre-commit (all hooks on the changed files) | green |

The NPU runs are the ordering gate: TPUT push + `NeighborBarrier` + `pipe_barrier(PIPE_ALL)`/`dsb` ordering is not modelled in sim, and is proven on silicon for every ragged size × rank count.

## Scope notes

- **Sum-only (deliberate):** the HOST ring kernel is Sum-only by construction (`entry.cpp.in` `enum ReduceOp { kSum = 0 }`; `TPUT<AtomicType::AtomicAdd>` cannot Prod). Prod on the HOST ring requires the atomic-add → non-atomic-push algorithm change, tracked separately. The ragged ST mirrors the InCore ring ragged ST, which is also Sum-only.
- **No perf regression on the previously-supported path by construction:** for `numel % NR == 0` the floor boundaries coincide exactly with the old `chunk_elems` multiples, so `send_base`/`send_end`/`send_len`/`max_chunk_elems`/`tile_cols` are identical and the generated kernel is unchanged (the only removed code is the `numel % nranks != 0` guard branch, which never fired on the supported path).
- **Pre-existing pyright failure (unrelated):** `pre-commit run --all-files` reports 4 pyright errors in `python/pypto/runtime/distributed_runner.py` (`DataType`/`AccessMode`/`BackendKind` unknown import symbol) that reproduce identically on a pristine `origin/main` worktree @ `f1a2f35f` — an environment artifact of the isolated hook env (missing `_task_interface` C-extension the installed `simpler` re-exports from), not caused by this PR. pyright passes on all files changed here.

### DIFF
--- a/src/ir/op/distributed/collective.cpp
+++ b/src/ir/op/distributed/collective.cpp
@@ -167,25 +167,11 @@ TypePtr DeduceBuiltinTensorAllReduceRingType(const std::vector<ExprPtr>& args,
         << " for NR = " << sig_shape1_const->value_;
   }
 
-  // Compile-time validation: the host builtin ring kernel partitions src into
-  // NR contiguous compile-time chunks, so the src shape must be statically
-  // known — a dynamic extent could reach the kernel with a runtime numel that
-  // is not divisible by NR and silently return unreduced data.  A non-divisible
-  // numel would produce a trailing partial chunk the schedule cannot handle.
-  if (sig_shape1_const && sig_shape1_const->value_ > 0) {
-    int64_t src_numel = 1;
-    for (const auto& dim : src_type->shape_) {
-      auto extent = As<ConstInt>(dim);
-      CHECK(extent) << kOpName
-                    << " requires a statically-known src shape (dynamic host-ring extents are not "
-                       "supported; the ring schedule partitions src into NR compile-time chunks)";
-      src_numel *= extent->value_;
-    }
-    const int64_t nr = sig_shape1_const->value_;
-    CHECK(src_numel % nr == 0) << kOpName << " requires the src data size (product of shape = " << src_numel
-                               << ") to be an exact multiple of the rank count (" << nr
-                               << "); got a remainder of " << (src_numel % nr);
-  }
+  // The host builtin ring kernel partitions src into NR balanced, potentially
+  // ragged chunks at runtime (chunk r spans [floor(numel*r/NR), floor(numel*(r+1)/NR)))
+  // and narrows every TPUT transfer to the chunk's exact extent via the staging
+  // tile's valid_shape — no compile-time static-shape or divisibility
+  // requirement remains (issue #2242).
 
   auto op_value = GetRequiredKwarg<int>(kwargs, "op", kOpName);
   auto dtype = GetRequiredKwarg<DataType>(kwargs, "dtype", kOpName);
--- a/src/ir/transforms/lower_host_tensor_collectives_pass.cpp
+++ b/src/ir/transforms/lower_host_tensor_collectives_pass.cpp
@@ -178,38 +178,6 @@ void CheckStaticSignalCapacity(const CallPtr& call, const ExprPtr& signal_expr, 
       << ") must be at least the participating device count (" << required_slots << ")";
 }
 
-void CheckRingChunkConstraints(const CallPtr& call, const ExprPtr& src_expr, size_t world_size) {
-  auto src_type = As<DistributedTensorType>(src_expr->GetType());
-  INTERNAL_CHECK_SPAN(src_type, call->span_)
-      << "LowerHostTensorCollectives: ring allreduce src must be DistributedTensorType";
-  if (src_type->shape_.empty()) return;
-
-  // The ring kernel partitions src into NR contiguous compile-time chunks
-  // (chunk_elems = numel / NR), so the host-ring path requires a
-  // statically-known src shape.  A dynamic extent would let a runtime numel
-  // that is not divisible by NR reach the kernel, which silently returns
-  // unreduced data instead of failing — reject dynamic host-ring extents here.
-  int64_t src_numel = 1;
-  for (const auto& dim : src_type->shape_) {
-    auto extent = As<ConstInt>(dim);
-    CHECK_SPAN(extent, call->span_)
-        << "LowerHostTensorCollectives: ring allreduce requires a statically-known "
-           "src shape (dynamic host-ring extents are not supported; the ring "
-           "schedule partitions src into NR compile-time chunks)";
-    src_numel *= extent->value_;
-  }
-
-  int64_t nr = static_cast<int64_t>(world_size);
-
-  // Divisibility: the host builtin ring schedule partitions data into NR
-  // contiguous chunks of chunk_elems = numel // NR.  A non-divisible numel
-  // would produce a trailing partial chunk that the kernel cannot handle.
-  CHECK_SPAN(src_numel % nr == 0, call->span_)
-      << "LowerHostTensorCollectives: ring allreduce requires the per-rank data size (product of src shape = "
-      << src_numel << ") to be an exact multiple of the rank count (" << nr << "); got a remainder of "
-      << (src_numel % nr);
-}
-
 void CheckRingSignalCapacity(const CallPtr& call, const ExprPtr& signal_expr, size_t world_size) {
   auto signal_type = As<DistributedTensorType>(signal_expr->GetType());
   INTERNAL_CHECK_SPAN(signal_type, call->span_)
@@ -287,9 +255,6 @@ void CheckAllReduceSignalCapacity(const CallPtr& call, const ExprPtr& signal_exp
     CHECK_SPAN(world_size <= kMaxSupportedRanks, call->span_)
         << "LowerHostTensorCollectives: ring allreduce requires " << static_cast<int>(kMaxSupportedRanks)
         << " or fewer participating devices, got " << world_size;
-    INTERNAL_CHECK_SPAN(call->args_.size() >= 1, call->span_)
-        << "LowerHostTensorCollectives: ring allreduce requires a src arg";
-    CheckRingChunkConstraints(call, call->args_[0], world_size);
     return;
   }
   CheckAllReduceCoreCapacity(call, core_num);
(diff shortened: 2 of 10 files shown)


src/ir/op/distributed/collective.cpp
```diff
@@ -167,25 +167,11 @@ TypePtr DeduceBuiltinTensorAllReduceRingType(const std::vector<ExprPtr>& args,
...
-                    << " requires a statically-known src shape (dynamic host-ring extents are not "
-                       "supported; the ring schedule partitions src into NR compile-time chunks)";
-      src_numel *= extent->value_;
-    }
-    const int64_t nr = sig_shape1_const->value_;
-    CHECK(src_numel % nr == 0) << kOpName << " requires the src data size (product of shape = " << src_numel
-                               << ") to be an exact multiple of the rank count (" << nr
-                               << "); got a remainder of " << (src_numel % nr);
-  }
+  // The host builtin ring kernel partitions src into NR balanced, potentially
```
YunjiQin, line 170: The removed block enforced **two independent** constraints, and only one of them is what this PR is about:

1. `src_numel % nr == 0` — ragged. This is the PR's subject, and the runtime `RingChunkStart` geometry genuinely replaces it. :+1:
2. `CHECK(extent)` on every dim — **dynamic extents**. Nothing else in this PR addresses this one.

Everywhere else the PR is scoped to (1): the title, the summary, the kernel comment, the new ST parametrization `{1, 17, 4097, 65537}`, and the renamed UT are all static-shape ragged cases. But the docs now claim (2) as well:

> ...so `numel(src)` need not be divisible by `NR` and the host-ring `src` shape need not be statically known. ... **ragged and dynamic inputs are fully supported.**

I can't find a UT or ST that puts a dynamic extent through the host-ring path, so that sentence is unverified. These two `CHECK`s were the only guard on it — if anything downstream still assumes a static `numel` (window sizing, variant / `kernel_config` generation), a user now gets an obscure downstream failure instead of the clear compile-time error they used to get.

The kernel side is fine — `numel` is accumulated at runtime from `data_tensor->shapes[]` — so this is most likely just a coverage / doc-scope gap rather than a real defect. Two ways to close it:

**(a) Add coverage.** A deducer-level UT is cheap and lands exactly on the deleted check. `tests/ut/ir/test_distributed_ops.py:322` already has the mesh analogue to copy:

```python
def test_tensor_allreduce_accepts_dynamic_shape():
    n = ir.Var("n", ir.ScalarType(DataType.INT64), span)
    src = ir.Var("src", ir.DistributedTensorType([n], DataType.FP32), span)
    ...
```

A ring version (dynamic-extent `src` + `[7, 4]` INT32 signal, asserting the builtin type-deduces instead of raising) would pin (2) directly.

**(b) Or scope the docs back** to "ragged is supported; a dynamic host-ring `src` shape is still rejected", and keep the `CHECK(extent)` loop. That keeps this PR's blast radius equal to its title.

Either is fine by me — I'd just rather not have a doc-advertised capability with no test behind it.

src/ir/transforms/lower_host_tensor_collectives_pass.cpp
```diff
@@ -178,36 +178,16 @@ void CheckStaticSignalCapacity(const CallPtr& call, const ExprPtr& signal_expr,
       << ") must be at least the participating device count (" << required_slots << ")";
 }
 
-void CheckRingChunkConstraints(const CallPtr& call, const ExprPtr& src_expr, size_t world_size) {
+void CheckRingChunkConstraints(const CallPtr& call, const ExprPtr& src_expr) {
```
YunjiQin: After this change the function does exactly one thing — assert `src` is a `DistributedTensorType` — and that assertion is already made twice elsewhere:

- `MakeBuiltinAllReduce` (this file, line 315): same expression `call->args_[0]->GetType()`, near-identical message — and it runs on **both** the static and the dynamic device-set path, so it strictly dominates this one (`CheckRingChunkConstraints` sits behind `if (!world_size_known) return;` at line 285).
- `DeduceBuiltinTensorAllReduceRingType` (`src/ir/op/distributed/collective.cpp:146`): `CHECK(src_type) << " src must be a DistributedTensor"`.

So the body is a no-op wrapper whose only remaining value is the comment — and the name actively misleads: a future reader sees `CheckRingChunkConstraints` in the call chain and concludes the ring's chunk geometry still has compile-time guarding, which it no longer does.

Suggest deleting it and relocating the comment:

```diff
-void CheckRingChunkConstraints(const CallPtr& call, const ExprPtr& src_expr) {
-  auto src_type = As<DistributedTensorType>(src_expr->GetType());
-  INTERNAL_CHECK_SPAN(src_type, call->span_)
-      << "LowerHostTensorCollectives: ring allreduce src must be DistributedTensorType";
-  // The host builtin ring kernel partitions src into NR balanced ...
-}
```

and at the call site:

```diff
     if (!world_size_known) return;
     CheckRingSignalCapacity(call, signal_expr, world_size);
     CHECK_SPAN(world_size <= kMaxSupportedRanks, call->span_) << ...;
-    INTERNAL_CHECK_SPAN(call->args_.size() >= 1, call->span_)
-        << "LowerHostTensorCollectives: ring allreduce requires a src arg";
-    CheckRingChunkConstraints(call, call->args_[0]);
     return;
```

Dropping the `args_.size() >= 1` guard costs nothing: with `CheckRingChunkConstraints` gone it has no consumer, and the dynamic path already reaches `MakeBuiltinAllReduce`'s `call->args_[0]` without it today.

The ragged-geometry explanation is worth keeping — it reads well next to `const bool as_ring = ShouldLowerAllReduceAsRing(call);` in `MakeBuiltinAllReduce`, where the ring builtin is actually selected.

YunjiQin (commented):
Two review notes on the compile-time-check removals — the runtime ragged geometry itself looks right to me (the `floor(numel*r/NR)` boundaries match the InCore composite ring's FP32 convention in `lower_composite_ops_pass.cpp:1413`, `numel` is already accumulated at runtime from `data_tensor->shapes[]`, empty chunks at `numel < NR` are handled, and the divisible path is numerically identical to before). Both notes are about scope and dead code, not the algorithm.

src/ir/op/distributed/collective.cpp
georgebisbas: Fixed — added test_tensor_allreduce_ring_accepts_dynamic_shape in tests/ut/ir/test_distributed_ops.py (ring analogue of the mesh test_tensor_allreduce_accepts_dynamic_shape): a dynamic-extent src + [7, 4] INT32 ring signal now type-deduces instead of raising, pinning the doc claim that dynamic host-ring inputs are supported. Ragged non-divisible coverage stays as the lowering UT (SIZE=27, P=4) + the {1,17,4097,65537} x P=2/P=4 ST. Branch also rebased onto current main.

src/ir/transforms/lower_host_tensor_collectives_pass.cpp
georgebisbas: Fixed — deleted CheckRingChunkConstraints (its only remaining assert was already made by MakeBuiltinAllReduce, which strictly dominates this path, and by the ring builtin deducer), dropped the now-consumerless args_.size() >= 1 guard, and relocated the ragged-geometry comment next to the as_ring selection in MakeBuiltinAllReduce. Branch rebased onto current main.

src/ir/op/distributed/collective.cpp
YunjiQin: Pushed 67a72fe3 to this branch (maintainer edit) — the lowering-side companion test, since `builtin.tensor.allreduce_ring` is `set_internal_only(true)` and its deducer is only reachable via `LowerHostTensorCollectives` -> `MakeBuiltinAllReduce` -> `OpRegistry::CreateInternal`. `test_tensor_allreduce_ring_accepts_dynamic_shape` builds the user-level `pld.tensor.allreduce`, which runs `DeduceTensorAllReduceType` (`allreduce.cpp:79`) — that one never inspects `shape_` and never reads `mode`, so it passes identically on `origin/main` and cannot observe the check you removed. Keep it, it just pins a different thing.

The new `test_host_allreduce_ring_accepts_dynamic_src_extent` uses a `[pld.world_size(), SIZE]` ring src window (a `pld.system.world_size` Call, not a `ConstInt` dim — exactly what the deleted `CHECK(extent)` rejected). Verified non-vacuous: with that check temporarily restored it fails with `requires a statically-known src shape`, while the `test_distributed_ops.py` one still passes. Full `test_lower_host_tensor_collectives.py` + `test_distributed_ops.py` = 229 passed, pre-commit green on the changed file.

YunjiQin (approved):

PR #2607, opened 2026-09-01
refactor(ir)!: re-export the distributed artifact types, internalize the argument builders

lyfne123:
## Summary

Two entry-point cleanups from the plan of record in
`docs/en/dev/08-entry-points.md`. Both are about the same question — which names
on the artifact layer are meant to be typed by hand — and they answer it in
opposite directions, so they travel together.

`DistributedCompiledProgram` and `DistributedConfig` are public types that
`pypto.ir` did not export. Every caller therefore reached past the package into
`pypto.ir.distributed_compiled_program`: 91 import statements across tests,
examples and docs, spelling out a private-looking module path for two public
names.

`CompiledProgram.build_orch_args` and `build_call_config` are the mirror image.
They marshal user arguments into simpler's `TaskArgs` and `CallConfig`, and the
user manual listed them as part of the extraction surface — but the supported
route to them is `ChipWorker.run` / `register`, which calls both for you.

## Changes

- `python/pypto/ir/__init__.py`: export `DistributedCompiledProgram` and
  `DistributedConfig` alongside `CompiledProgram`. The defining module stays
  importable and keeps its four `pypto.runtime` importers, which take it
  directly to avoid an import cycle. The re-export is an alias of the same
  objects, so an `isinstance` check cannot depend on which spelling the caller
  used — there is a test for that.
- `python/pypto/ir/__init__.pyi`: the type stub shadows the runtime module for
  type checkers, so the two new names had to land there too. It had also drifted
  before this change — `CompiledProgram`, `DiagnosticPhase`, `DiagnosticCheck`
  and `DiagnosticCheckSet` were exported at runtime but absent from the stub,
  which is why `ir.CompiledProgram` did not type-check. All six are now listed.
- `python/pypto/ir/compiled_program.py`: rename both builders to
  `_build_orch_args` / `_build_call_config`, on `CompiledProgram` and on
  `_SubChipCallable`. The runner-side helper the latter delegates to is now
  imported under an alias, so the module-level function and the method it backs
  no longer share a name at the call site.
- `python/pypto/runtime/worker.py`, `python/pypto/jit/decorator.py`: follow the
  rename; the JIT `compile()` docstring stops advertising the builders.
- Tests and examples: 69 files migrate to `from pypto.ir import ...`.
  `tests/ut/ir/test_compiled_program.py` and `test_distributed_compiled_program.py`
  keep the module-path import — they also pull `_DISTRIBUTED_META_FILENAME` from
  it, and they are white-box tests of that module.
- `tests/st/runtime/framework_and_models/test_compiled_program.py`: the manual
  `simpler.Worker` helper keeps calling the builders directly, so the
  marshalling `ChipWorker.run` delegates to stays covered on its own; its
  docstring now says why it reaches for internals.
- Docs (`docs/{en,zh}/`, 12 files): drop the builders from the `CompiledProgram`
  contract table and from the JIT extraction surface, pointing at `ChipWorker`
  instead; list them under "What is internal" in the entry-points page, which
  also gains the `pypto.ir` import line and loses the paragraph saying the
  distributed types are *not* re-exported; migrate the remaining snippets.

## Migration

Nothing outside this repository imported either builder. `pypto-lib` imports
`DistributedConfig` / `DistributedCompiledProgram` from the module path, which
keeps working unchanged.

```python
# before
from pypto.ir.distributed_compiled_program import DistributedConfig
orch_args, coerced, style = compiled.build_orch_args(*args, worker=w)
cfg = compiled.build_call_config(run_config)

# after
from pypto.ir import DistributedConfig
worker.run(compiled, *args)          # marshals both for you
```

A caller driving a hand-constructed `simpler.worker.Worker` still gets
`chip_callable`, `runtime_name` and `runtime_config` from the public surface.

## Verification

Run in the worktree against its own build, at the pushed commit.

- `pytest tests/ut/ -n 16`: 10992 passed, 8 skipped, 1 xfailed
- `tests/lint/check_*.py` (12 scripts, incl. docs nav and en/zh parity): all pass
- `ruff check` / `ruff format --check` (pinned 0.14.8) on the changed Python
  files: clean
- `pyright python/pypto tests examples`: only two pre-existing
  `torch.float4_e2m1fn_x2` errors, from this machine's older torch rather than
  CI's pinned 2.8.0
- `markdownlint-cli2 v0.20.0` on the 12 changed pages: 0 errors
- Import-cycle probe: `pypto.ir`, `pypto.runtime`, and seven submodules import
  cleanly as the first import in a fresh interpreter, in either order

One unit test is deselected:
`test_unified_ops.py::TestUnifiedSlicePadValue::test_symlinked_import_path_still_names_the_caller`.
It fails identically on unmodified `main` in this environment — the editable
install's meta-path finder redirects `pypto` inside the subprocess the test
spawns — and is unrelated to this change.

Device tests were not run; this change touches no kernel or codegen output.

### DIFF
--- a/docs/en/dev/03-runtime-replay.md
+++ b/docs/en/dev/03-runtime-replay.md
@@ -218,7 +218,7 @@ from pypto.runtime import execute_distributed_compiled
 execute_distributed_compiled("build_output/<jit_dir>/", [a, b, c])
 
 # reusable object (override the persisted platform / devices if needed):
-from pypto.ir.distributed_compiled_program import DistributedCompiledProgram, DistributedConfig
+from pypto.ir import DistributedCompiledProgram, DistributedConfig
 prog = DistributedCompiledProgram.from_dir(
     "build_output/<jit_dir>/",
     platform="a2a3",
--- a/docs/en/dev/08-entry-points.md
+++ b/docs/en/dev/08-entry-points.md
@@ -57,6 +57,17 @@ Python builtin `compile`, so prefer `from pypto import ir` and call
 
 Its parameters are documented in [Compiling](../user/execution/00-compile.md).
 
+Both artifact types it can return, and the config that selects the distributed
+one, are exported from `pypto.ir`:
+
+```python
+from pypto.ir import CompiledProgram, DistributedCompiledProgram, DistributedConfig
+```
+
+The defining module, `pypto.ir.distributed_compiled_program`, stays importable —
+`pypto.runtime` reaches for it directly to avoid an import cycle — but user code
+and tests should take the three names from `pypto.ir`.
+
 ## Execute
 
 | Entry | Layer | Takes | Location |
@@ -107,6 +118,8 @@ These have no stability guarantee. They are not in any `__all__`, and code
 outside PyPTO should not import them:
 
 - `pypto.ir.compile` — `_ensure_orchestration_headers`
+- `pypto.ir.compiled_program` — `CompiledProgram._build_orch_args`,
+  `CompiledProgram._build_call_config`
 - `pypto.runtime.device_runner` — `compile_and_assemble`, `compile_single_kernel`,
   `compile_single_orchestration`, `execute_on_device`
 - `pypto.runtime.kernel_compiler` — `KernelCompiler`, `compile_incore`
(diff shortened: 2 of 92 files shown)


lyfne123:
Folded into #2609, which carries this commit plus the deprecation of the directory-driven execute entry points. Both were Stage B of the same plan and touch the same question — which artifact-layer names are public — so they are easier to review together than stacked. Closing here; nothing is lost, #2609's first commit is this branch's HEAD verbatim.

PR #2648, opened 2026-09-04
test(st): run the scheduling swimlane assertions

luohuan19:
Thirty-five assertions under `tests/st/runtime/scheduling` have never run. They
gate on `--enable-chip-swimlane`, and CI passes that flag to exactly one file --
`test_perf_swimlane.py` -- so every one of them skipped in the batched shard,
silently, since the day it was written:

    test_manual_scope_pipeline.py        13
    test_pl_at_deps_pipeline.py          11
    test_phase_fence_dep_compression.py  11

The mechanism they use is the one that already works in CI: the fixture runs its
case, then picks the freshest `build_output/*/dfx_outputs/chip_swimlane_records
.json`. Nothing about them is broken; they were simply never given the flag.

Selection is by a `swimlane` marker rather than by path, because these files
also hold the correctness tests for the same pipelines, and those already run in
the batched shard. Marking the class, not the method, matches how the record is
scoped -- one class-or-module fixture per swimlane program.

The batched shard now excludes the marker. Those 35 ran there only to skip, and
35 silent skips are exactly what hid this: a real skip had nowhere to stand out.
Shard A collects 185 -> 146, and the 39 that leave are the 35 plus
`test_perf_swimlane`'s 4, which its own flagged step runs.

The new step omits `--forked`, unlike the perf_swimlane step above it. These
hang their record off class- or module-scoped fixtures, so one process reuses it
across the class -- about 20 device runs instead of 35. `test_perf_swimlane` is
left exactly as it was; a change there would risk a step that works today.

Not verified locally: whether the 35 pass. This box's simulator fails before any
of them reaches an assertion --

    RuntimeError: sim context failed to bind host-log state:
                  module rejected the host-log state ABI

-- which is a stale sim runtime library, unrelated to this change and present on
a clean checkout too. What is verified is the selection: `-m swimlane` picks
exactly those 35 under `runtime/scheduling` (and 39 across `tests/st`), and the
shard-A exclusion removes exactly those 39 and nothing else. CI is what will say
whether the assertions hold; some may not, and a failure there is the finding
this change exists to surface.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1200,7 +1200,12 @@ jobs:
           # compile + golden card-free, device run BATCHED via task-submit. The
           # `not device_batch` (direct @pl.jit / config=test_config) tests for
           # ALL dirs run in the `system-tests-direct` job.
-          python -m pytest tests/st -v -m device_batch \
+          #
+          # `not swimlane` because those assert on a chip-swimlane record and
+          # skip without `--enable-chip-swimlane`, which this step does not pass.
+          # They ran here only to skip, and 35 silent skips buried the real ones;
+          # the flagged steps in `system-tests-direct` run them for real.
+          python -m pytest tests/st -v -m "device_batch and not swimlane" \
             --ignore=tests/st/distributed \
             --ignore=tests/st/codegen \
             --ignore=tests/st/runtime/ops \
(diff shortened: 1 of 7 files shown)


luohuan19:
Folded into https://github.com/hw-native-sys/pypto/pull/2654, which carries both commits from this branch unchanged plus the case-declaration work. Both touch the same scheduling test files, so keeping them apart only bought a merge conflict.

PR #2654, opened 2026-09-07
test(st): run the swimlane assertions, and put every inline-compiled case in the pool

luohuan19:
Supersedes and includes #2648, which is closed in favour of this branch: both touch the same scheduling test files, and holding one back to avoid the conflict served nothing.

Three things, in the order they were found: tests that had never run, cases the pre-compile pool could not see, and a guard so neither comes back.

## Tests that had never run

**35 swimlane assertions.** They are `@pytest.mark.swimlane`, and the only shard that could have selected them runs `--forked`, which the DFX profiling collection does not survive. Shard A narrows to `-m "device_batch and not swimlane"` (185 -> 146 collected) and a dedicated step runs the swimlane set unforked.

Turning them on showed 26 of them were also broken: all failed on `KeyError: 'tasks'`. The runtime's swimlane JSON is v2 now — raw `aicore_tasks` / `aicpu_tasks` plus `metadata` — and the unified `tasks` view every assertion reads is rebuilt by the runtime's own `swimlane_converter.read_perf_data`, never present in the file. All 9 read sites go through a new `harness.swimlane.read_swimlane`.

**One multi-card test.** `test_benchmark_st.py::test_benchmark_l3_surfaces_per_rank_timing` needs two cards, but multi-card was selected by directory — the two-card job runs `tests/st/distributed` and nothing else does. Living in `framework_and_models`, it was collected by a single-card step where it skipped itself from its own body.

Both are the same shape: the requirement lived where selection cannot see it, so the test was collected in the one place that could not satisfy it and reported itself as skipped forever. `@pytest.mark.multi_card(n)` states it where selection can act on it; the single-card shards deselect it and a new step in the two-card job picks it up. On board: 1 card skips with a reason, 2 cards is `1 passed` — its first execution.

## Cases the pool could not see: 76 -> 0

The pre-compile pool needs the case list before any test body runs. Collection reads it from the test source, so a case built from an expression it cannot evaluate is invisible and compiles on its own thread of control while the pool idles.

The first commit names them at the end of a run. The rest declare all of them, in two groups.

**Built from an expression collection cannot evaluate** — a factory call, arithmetic on a name, an index into a parametrize value:

| file | n |
| --- | --- |
| `test_phase_fence_dep_compression.py` correctness | 12 |
| `test_gemv.py` | 7 |
| `test_sels.py` | 4 |
| `test_prelu.py` | 2 |
| `test_mgather.py` | 1 |

**Run by a fixture, then located by globbing** — the DFX group: a module- or session-scoped fixture (or a `_new_swimlane_json` helper) called `test_runner.run()` and then searched a shared `build_output` for records that appeared during the call, taking the newest by mtime. Whichever landed last won, which is another case's record whenever two runs overlap.

| file | n |
| --- | --- |
| `test_phase_fence_dep_compression.py` swimlane | 11 |
| `test_manual_scope_pipeline.py` | 4 fixture pairs |
| `test_pl_at_deps_pipeline.py` | 3 fixture pairs |
| `test_perf_swimlane.py` | 1 fixture pair |
| `test_dump_tag.py` | 1 fixture pair |

These declare their case and read through `case_run.swimlane()` / `case_run.dfx(...)`, which resolve inside the case's own artifact directory — named by the cache key, so no other case can write into it. That is what retires the mtime race.

One addition to the case surface made it possible. `@st.cases` already made a case a collection-time value and `CaseRun.work_dir` already pointed at the artifact directory; what was missing is that N tests declaring one case each scheduled their own execution of an artifact they already shared. Device-pool execution is memoised on the compile cache's key, giving that path the property the task-submit path always had through `_case_to_batch`. A second caller whose tolerance differs is a name collision rather than a shared case — they already share `golden.py` — so it asserts instead of absorbing it.

## A guard, so it stays at zero

Collection fails when a test reaches `test_runner` with no case it could read. The check runs where the misses are already detected, at the end of collection, so it fires before anything executes and names every offending node id:

```
ERROR: 1 test(s) reach test_runner with no case collection could read.
  tests/st/runtime/ops/test_mgather.py::TestGuardProbe::test_undeclared
```

A summary line was not enough: the advisory version of this report is one commit earlier in this branch, and the count it was reporting on had already reached 76 while nobody read it.

Two escapes, deliberately different in scope. `@pytest.mark.inline_case(reason=...)` exempts one test, for a case that genuinely cannot be a collection-time value; the reason is required, because an unexplained exemption is indistinguishable from a test nobody got round to declaring. `--allow-inline-cases` turns the guard off for the session — useful mid-edit, useless as a CI setting. Nothing in `tests/st` carries the marker today.

## Six defects found on the way

**`test_sels` reads the platform at construction time.** `_tmp_dtype` defaults to the src/dst dtype on A2/A3 and to UINT8 elsewhere, and A2/A3 rejects a `tsels` whose tmp element type differs from src and dst. The matrix binds a platform only after the object exists, so moving the platform off the constructor silently built the A5 form and all four cases failed ptoas. The product is spelled out instead, each case pinning its own platform.

**A pinned case was never deselected in a single-platform run.** `pytest_collection_modifyitems` compared the pin against the item's platform *parameter*, but the matrix expands only when the CLI names more than one platform — so in a single-platform run, the shape CI uses, there was no parameter to disagree with and every pinned case survived. An A5-pinned case would have been handed an A2A3 card.

**Artifact lookup had a race.** Covered above: globbing a shared `build_output` for the newest file by mtime returns another case's record under any concurrency, and does so silently.

**The inventory summary was truncating.** It printed the first 20 ids and `... and N more`; read as a list it looks complete, and every count taken off it landed at twenty-something. The real total was 76. The full list now goes to a gitignored `undiscovered_cases.txt` named in the summary line.

**A DFX marker did not enforce its own contract.** `swimlane` has always meant "needs `--enable-chip-swimlane`, skips without it", but each fixture implemented that skip itself; a class that declares its case has no such fixture. `pytest_runtest_setup` applies it now, along with a new symmetric `dump_args` marker, and both cover `--codegen-only` as the same condition by another route — nothing executes, so nothing is written.

**`--forked` and the pre-compile pool do not compose.** A forked child cannot resolve the compile futures its parent created. Measured on board:

| step | forked | unforked |
| --- | --- | --- |
| `test_perf_swimlane.py` | killed at the 900s task cap | 4 passed, 11.3s |
| `test_dump_tag.py` | 8 passed, 19.0s | 8 passed, 3.1s |

Both steps drop `--forked` and gain `--precompile-workers`. The flag is not an optimisation for them: a declared case's DFX lookup resolves inside the work_dir the pool publishes, and conftest returns early without the flag, so these steps would fail rather than merely run serially.

## Verification

Failure counts are not comparable here — a declared case starts honouring `--codegen-only`, so the count moves for reasons unrelated to the change. FAILED node-id *sets* were diffed against main per platform instead.

Card-free, the whole of `tests/st`: a5 is 0 both ways; a2a3 is 78 before and 79 after, the one difference being `test_argmax_reduction::TestTileArgReduce::test_row` — which is also the difference between two runs of the *same* tree (78/79/78 over three runs), so it is this suite's own variance. Nothing stopped failing. Collecting with `--platform=a2a3,a5` yields the same item count, and each converted test collects the same per-platform items on all four platforms.

On an a2a3 card, running each changed CI step's exact post-change command:

```
Test swimlane output                              4 passed                    13.9s
Test scheduling swimlane assertions               20 passed, 15 skipped       3m40s
Test dump-args output                             8 passed, 3 deselected       3.4s
Test multi-card cases outside tests/st/distributed  1 passed (2 cards)         6.5s
```

The guard was driven by adding an undeclared case to a real test file and confirming collection fails with exit 4, then reverting it.


tests/st/conftest.py
```diff
@@ -1060,6 +1132,31 @@ def _collect_test_case_from_item(
             _cache_key(instance, instance.get_platform() or platform, session_memory_planner),
             instance,
         )
+        return True
```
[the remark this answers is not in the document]
luohuan19: Fixed in b3d978e8 — you are right, and the consequence was worse than a missed case.

The early return also made `_collect_test_case_from_item` report `True`, so `_discover_cases` never recorded the miss. The undeclared case was invisible to the undiscovered inventory and the `--strict-case-discovery` guard this PR adds, i.e. it bypassed the very mechanism meant to catch it.

Now tracks a flag and returns after the loop, restoring what the docstring has always promised and what the pre-branch code did.

A regression test builds two cases from one body and asserts both are filed; I checked it fails against the early return and passes with the fix. No ST body constructs two cases today, so the pool still sees 1265 — this closes a latent hole rather than recovering cases.

tests/st/conftest.py
```diff
@@ -1097,6 +1298,95 @@ def _collect_test_case_from_item(
...
+    guard still sees an undeclared swimlane test in a run with no
+    ``--enable-chip-swimlane``. Skipped items therefore collect into a throwaway
+    dict: the answer is computed, the case is not kept.
+    """
+    discarded: dict[str, PTOTestCase] = {}
+    for item in session.items:
+        # Session hook: without this it walks tests/ut bodies too. See _is_st_item.
+        if not _is_st_item(item):
+            continue
+        runs = marker_skip_reason(item) is None
```
YunjiQin: `runs` is decided from markers alone, but `_pool_submitter` now *executes* every registered case on a card before the item loop starts. A test that skips from a fixture or from its body is invisible here, so its case is compiled **and** run.

There is a live instance inside this PR's own diff. `TestPhaseFenceDepCompressionCorrectness` carries an autouse fixture that skips the whole class when `--enable-chip-swimlane` is on (`test_phase_fence_dep_compression.py:895`), and this PR converts that class's 12 correctness cases to `@st.cases`. So

```
pytest tests/st/runtime/scheduling --enable-chip-swimlane --precompile-workers=8
```

now puts those 12 cases on a card and then skips every test that would have asserted on them. Before this PR they were compiled but never executed. The new CI step avoids it only because `-m swimlane` deselects the class — a developer running the directory does not.

That is the waste this function's own docstring says it prevents ("the pool would compile it, and the device-pool submitter would run it, for a test that skips before asserting anything"); markers just are not the only way a test skips.

Related, one layer down: line 1364 tests `m.name == "skip"`, which does not match `skipif` — `iter_markers` reports that one as `"skipif"`. No `skipif` exists under `tests/st` today, so it is latent rather than live, but it is the same blind spot.

Cheapest fix is probably to give that autouse fixture's condition a marker `marker_skip_reason` can read, which is the move the rest of the PR already makes.

tests/st/conftest.py
```diff
@@ -688,6 +747,44 @@ def pytest_configure(config):
...
+        "markers",
+        "inline_case(reason): this test's case cannot be a collection-time value, so it "
+        "compiles inline instead of in the pre-compile pool. Exempts the test from the "
+        "collection guard, which otherwise fails the session. A reason is required — an "
+        "unexplained exemption is indistinguishable from a test nobody got round to "
+        "declaring, which is what let the count reach 76.",
+    )
+    config.addinivalue_line(
+        "markers",
+        "dump_args: the test asserts on an args-dump manifest, so it needs "
```
YunjiQin: This marker is registered here and enforced below (`marker_skip_reason` reads it at line 908, `pytest_runtest_setup` applies the skip), but nothing carries it. `@pytest.mark.dump_args` appears zero times in this diff and zero times on `main`, so the branch never fires: `test_dump_tag.py` gets `inline_case` only, and its fixtures keep doing their own skip.

Nothing is broken by it. But "a marker registered where selection can act on it, applied nowhere" is the exact shape this PR exists to remove — `swimlane`, in the same commit, does both halves (registered, then applied to 11 classes). The `dump_args` half of that symmetry stopped at registration.

Suggest either putting it on `TestDumpTagManifest` / `TestSubmitDumpsManifest`, or dropping the two lines until a test needs it.

YunjiQin (commented):
Two findings from a read of the harness changes. Both are about surface this PR adds; neither blocks the main result, which reads well — the three problems all trace back to one root (a requirement written where selection cannot see it) and the fix moves the requirement rather than patching around it.

tests/st/conftest.py
luohuan19: Fixed in 2b6bd98f. You are right, and the instance you found reproduces exactly as described:

```
pytest tests/st/runtime/scheduling --enable-chip-swimlane --precompile-workers=8
  before   Pipeline: 25 test case(s)   47 skipped
  after    Pipeline: 13 test case(s)   47 skipped
```

Same outcomes, twelve fewer cases on the card. `TestPhaseFenceDepCompressionCorrectness` now carries `@pytest.mark.without_swimlane(reason=...)` — the inverse of `swimlane` — and the autouse fixture is gone. That is the move you suggested and the one the rest of the branch makes: state the requirement where selection can read it, rather than patch around it.

On `skipif`: new `_will_not_run` folds `skip` and `skipif` together, and is now what collection asks for both decisions — registration included, since a case behind either marker would otherwise be put on a card for a test pytest never intends to call. The condition is deliberately not evaluated: treating a conditional skip as "may skip" costs at most a serial inline compile, while guessing the other way costs a device run. `marker_skip_reason` is unchanged, so `pytest_runtest_setup` behaves as before.

One correction to the report, for the record: `tests/st` does have two `skipif` today, both in `test_case_declaration.py` guarding on `_ptoas_available()`. Neither uses `test_runner`, so the blind spot was indeed latent rather than live — for a different reason than "none exists".

Three regression tests cover both halves.

tests/st/conftest.py
luohuan19: Dropped in 2b6bd98f — the registration and the `marker_skip_reason` branch both.

Of your two options I took the second. The only file that would carry it is `test_dump_tag.py`, whose case declarations this branch already reverted after two CI failures with two different symptoms, and re-opening that file for a marker nothing selects on is not worth it. If a test later needs the flag gated this way, adding it back is three lines and it will arrive with a user.

You are right that this is the exact shape the PR exists to remove; registered-and-unapplied is the same "declared where nobody reads it" failure in the other direction. With it gone the remaining condition replaces what had become a one-element loop, so `marker_skip_reason` reads a little straighter too.

YunjiQin (approved):

PR #2702, opened 2026-09-09
fix(ir): give every compile its own output directory

luohuan19:
`ir.compile` derived its default output directory from the program name and a
one-second timestamp, then created it with `exist_ok=True`:

```python
output_dir = os.path.join(base, f"{program.name}_{timestamp}")
os.makedirs(output_dir, exist_ok=True)
```

Two compiles of same-named programs inside one second therefore shared a
directory, and the second's kernels overwrote the first's. Nothing raised. The
caller kept a `DistributedCompiledProgram` whose `output_dir` pointed at another
program's artifacts, so dispatching it ran the wrong kernel — only a numeric
assertion could notice.

## How it surfaced

Preparing several programs before dispatching any of them. Three FP16 reduce
variants sharing the class name `HostTensorAllReduceArbitraryLength`:

```
max   -> build_output/HostTensorAllReduceArbitraryLength_20260908_232837
min   -> build_output/HostTensorAllReduceArbitraryLength_20260908_232837
prod  -> build_output/HostTensorAllReduceArbitraryLength_20260908_232838
```

The `min` dispatch returned `i * (100 + i)` — prod's answer.

Every caller in tree today compiles one program and runs it before compiling the
next, which is why this has not bitten: the overwrite lands after the artifacts
have been used. It bites the moment a batch is compiled up front.

## The fix

`tempfile.mkdtemp` under the same base, which is how `@pl.jit` has always
resolved its own output directory (`decorator.py`). An explicit `output_dir=` is
untouched.

The test asserts three compiles land in three directories rather than pinning the
naming scheme, so it survives the scheme changing again; it fails against the
timestamp form.

## Note

Split out of https://github.com/hw-native-sys/pypto/pull/2696 (hw-native-sys/pypto),
which keeps the distributed-ST changes that motivated finding this. This half is an
independent correctness fix and does not depend on them.

### DIFF
--- a/python/pypto/ir/compile.py
+++ b/python/pypto/ir/compile.py
@@ -11,8 +11,8 @@
 
 import logging
 import os
+import tempfile
 from contextlib import AbstractContextManager, nullcontext
-from datetime import datetime
 from typing import TYPE_CHECKING, Any, NamedTuple
 
 from pypto.backend import BackendType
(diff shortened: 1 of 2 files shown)


luohuan19:
Folding back into https://github.com/hw-native-sys/pypto/pull/2696 (hw-native-sys/pypto) — keeping the output-directory fix and the distributed-ST work it surfaced in one PR.

PR #2726, opened 2026-09-10
TEMP(do not merge): A/B control — serving-guard at ptoas v0.60

luohuan19:

Diagnostic for hw-native-sys/pypto#2686. That PR's `serving-guard` job faults the
card during pypto-serving's Qwen3-14B warmup prefill dispatch, reproducibly and
on two different dies (device 8 and device 9), always at the same point:

```
[warmup] prefill dispatch … (batch=16, tokens=4096)
rtStreamQuery (AICore) failed: 507901
Device[N] fault, ret=0x7110011 … reason=hdc disconnect
validate_runtime_impl: orch_error_code=2 HEAP_RING_DEADLOCK
```

This branch is #2686's head with **one** change: `toolchain/versions.env` back at
`PTOAS_VERSION=v0.60`. Both codegen commits are still here, deliberately — the
store-partition change alters the emitted `tile.store` window under either
assembler, so it has to stay in the control for the comparison to isolate the pin.

- `serving-guard` green here → the v0.61 bump is what breaks serving generation.
- `serving-guard` red here → the codegen commits are, and the pin is innocent.


luohuan19:

**What it established:** with #2686's two codegen commits present and only the pin
held at v0.60, `serving-guard` passed Qwen3 accuracy, Qwen3 serving and the
DeepSeek V4 MTP matrix. That is the half of the comparison that made the pin the
sole remaining variable against #2686.

**Caveat worth recording:** the first push here reverted `versions.env` to main's
exact content, which took the file out of the PR diff, so `toolchain_changed` was
false and `serving-guard` was *skipped* — the green on that push meant nothing.
The result above is from the second push, which restored the file to the diff with
a comment while leaving the pin at v0.60.

**Where it ended up:** the corruption turned out to be intermittent (#2735 red and
#2736 green on identical content), and a static comparison of every `.pto` serving
compiles — 2861 kernels, both assemblers — found them structurally identical. So
the pin does not change what serving runs. #2686 is now fully green.

PR #2728, opened 2026-09-10
TEMP(do not merge): A/B control — serving-guard on plain main

luohuan19:

Diagnostic for hw-native-sys/pypto#2686, and the arm that should have existed first:
it validates the `serving-guard` job itself.

This branch is `main` plus **only** that PR's `serving-guard` CI job. No codegen
commits, and the pin stays at `PTOAS_VERSION=v0.60` — the comment appended to
`toolchain/versions.env` is there solely to put the file in the PR diff, which is
what makes the `changes` job set `toolchain_changed=true`.

Why it is needed: pypto-serving's own `unit-tests` job proves these same Qwen3
guards pass, but it has only ever proved it on runner `npu-1`. `serving-guard`
runs on the `npu-xp` runner (`227-npu-a2a3-XP-1`) instead, so "pypto-serving runs
on that host at all" has never been shown. On #2686 it fails there reproducibly,
on two different dies, at the same point:

```
[warmup] prefill dispatch … (batch=16, tokens=4096)
rtStreamQuery (AICore) failed: 507901
Device[N] fault, ret=0x7110011 … reason=hdc disconnect
validate_runtime_impl: orch_error_code=2 HEAP_RING_DEADLOCK
```

- `serving-guard` green here → the host and the wiring are sound, and the fault
  comes from #2686's own content (#2726 then separates the pin from the codegen).
- `serving-guard` red here → the fault is the host or the job, and nothing in
  #2686 causes it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,7 @@ jobs:
     outputs:
       docs_only: ${{ steps.classify.outputs.docs_only }}
       docs_changed: ${{ steps.classify.outputs.docs_changed }}
+      toolchain_changed: ${{ steps.classify.outputs.toolchain_changed }}
     steps:
       - name: Classify changed paths
         id: classify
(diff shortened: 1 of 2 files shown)


luohuan19:

**What it established:** plain `main` at v0.60, with only #2686's `serving-guard`
job added, passed both Qwen3 guards on the `npu-xp` runner. That answered the
question this arm existed for — pypto-serving's own `unit-tests` job had only ever
proved those guards on runner `npu-1`, so "serving runs on this host at all" was
untested until here. The host and the job wiring are sound; the failures seen on
#2686 were not an artifact of standing the job up.

**Also surfaced here:** the DeepSeek V4 step failed on this arm at v0.60, and the
16-card DSpark guard failed on every arm — including pypto-serving's own
`unit-tests` on its main branch. DSpark has since been dropped from the job
(hw-native-sys/pypto@a2849d34) precisely because nothing on this side can turn it
green, and keeping it would leave the job permanently red.

**Where it ended up:** the corruption proved intermittent, and a static comparison
of all 2861 kernels serving compiles found v0.60 and v0.61 structurally identical.
#2686 is now fully green.

PR #2735, opened 2026-09-10
TEMP(do not merge): A/B control — serving-guard on main at ptoas v0.61

luohuan19:

Fourth arm of the diagnostic for hw-native-sys/pypto#2686. `main` + that PR's
`serving-guard` job + the pin bumped to `v0.61`, and **no codegen fixes**.

The three arms so far say the Qwen3-14B corruption tracks the pin:

| arm | pypto code | ptoas | `test_qwen3_output_matches_expected_tokens` |
| --- | ---------- | ----- | ------------------------------------------- |
| #2686 | store + remote_load fixes | **v0.61** | `'!!!!!!!!'` — token ids all `0` |
| #2726 | identical fixes | v0.60 | correct |
| #2728 | plain main | v0.60 | correct |

That is not yet enough to exonerate the codegen commits. v0.60's
`pto.partition_view` bounds check bails when the source dim is `kDynamic`; v0.61
folds constant shape operands into that dim and actually enforces it. So a *wrong*
store window emitted by the store fix would be **inert under v0.60 and live under
v0.61** — which fits the table above exactly as well as "v0.61 is broken on its own".

This arm separates them:

- garbage tokens here → v0.61 corrupts generation on its own; the codegen commits
  are innocent, and this is a PTOAS bug to file upstream.
- correct tokens here → the window the store fix emits is wrong, and v0.61 is
  merely the first assembler that honours it.
- compile failure here → serving's kernels do reach the rank>2 store path, so the
  fix is causally in play; rerun with it restored.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,7 @@ jobs:
     outputs:
       docs_only: ${{ steps.classify.outputs.docs_only }}
       docs_changed: ${{ steps.classify.outputs.docs_changed }}
+      toolchain_changed: ${{ steps.classify.outputs.toolchain_changed }}
     steps:
       - name: Classify changed paths
         id: classify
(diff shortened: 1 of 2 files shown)


luohuan19:

**What it established:** plain `main` at v0.61, carrying **none** of #2686's
codegen commits, reproduced the corruption — `test_qwen3_output_matches_expected_tokens`
returned token ids of all `0` (`'!!!!!!!!'`). That exonerated the store-partition
and remote-load fixes: they are not required for the failure to appear.

It also ruled out the alternative this arm was built for. v0.60's
`pto.partition_view` bounds check bails on a `kDynamic` source dim while v0.61
folds constants into it, so a *wrong* window emitted by the store fix would have
been inert at v0.60 and live at v0.61 — indistinguishable from "v0.61 is broken"
until this arm removed the fix and still failed.

**Where it ended up:** the conclusion I drew at the time — that the pin was
therefore the cause — was premature at n=1 per cell. #2736 is this branch plus 28
lines of CI and it passed the same guard, so the failure is **intermittent**. A
static comparison of every `.pto` serving compiles (2861 kernels, both assemblers,
op sequence + operand types + sync + dynamic-view constructor values) found them
structurally identical, with v0.61's only difference being view declarations no op
consumes. #2686 is now fully green.

PR #2736, opened 2026-09-10
TEMP(do not merge): probe v0.61 sync insertion + dump serving's .pto

luohuan19:
**Throwaway diagnostic branch — do not merge, do not review.**

Fifth arm for hw-native-sys/pypto#2686. The 2×2 is settled:

| arm | pypto code | ptoas | Qwen3 accuracy |
| --- | ---------- | ----- | -------------- |
| #2686 | store + remote_load fixes | **v0.61** | `'!!!!!!!!'` (token ids all `0`) |
| #2726 | identical fixes | v0.60 | correct |
| #2728 | plain main | v0.60 | correct |
| #2735 | plain main | **v0.61** | `'!!!!!!!!'` |

v0.61 corrupts Qwen3-14B generation on its own; the codegen commits are innocent.
The question is now *which* of the 166 commits in `v0.60..v0.61` does it.

**Most of them cannot.** pypto invokes ptoas with only `--enable-insert-sync` and
`--pto-level=level3` (plus `--pto-arch`), and `--pto-backend` defaults to `emitc` —
so every VPTO change in that range (soft post-update flipped on by default, A2/A3
gather lowering, the VPTO scheduler) is unreachable from this configuration.

Sync insertion *is* reachable, and v0.61 deleted the graph sync solver
(`64903ecc`) whose own help text described it as mutually exclusive with
`--enable-insert-sync`. A sync regression also fits both symptoms: garbage tokens,
and — before the runner was restarted — a card fault (`507901`, hdc disconnect).

Two independent changes here:

1. **`--enable-inject-barrier-all-sync`** appended to pypto's ptoas flags. It
   fences every memory-effecting pipe op with `PIPE_ALL`, which should mask any
   missing barrier. Correct tokens under it → the defect is in sync insertion.
   Still `'!!!!!!!!'` → sync is exonerated and the layout/memref commits are next.

2. **The compiled `.pto` uploaded as an artifact**, pass or fail. pypto emits the
   `.pto` and ptoas consumes it, so the same pypto emits byte-identical `.pto`
   under either pin and the entire v0.60/v0.61 difference is `ptoas(.pto)`. With
   those files the A/B reproduces on any box that has both assemblers — no card
   and no model weights needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2044,6 +2044,34 @@ jobs:
             --timeout 2400 --max-time 3600 \
             --run "source $GITHUB_WORKSPACE/serving-checkout/activate.sh && cd $GITHUB_WORKSPACE/pypto-serving && TASK_DEVICE=\$TASK_DEVICE PYTHONPATH=$GITHUB_WORKSPACE/pypto-serving PYPTO_DSV4_DSPARK_MODEL_DIR=$PYPTO_DSV4_DSPARK_MODEL_DIR PYPTO_RUNTIME_LOG=$PYPTO_RUNTIME_LOG python -m pytest tests/test_deepseek_dspark_accuracy.py -q -s"
 
+      - name: Upload the .pto serving compiled
+        # TEMP DIAGNOSTIC (throwaway branch, do not merge). pypto emits the .pto
+        # and ptoas turns it into device code, so two runs of the same pypto emit
+        # byte-identical .pto and the whole v0.60/v0.61 difference lives inside
+        # ptoas. Getting these files off the runner makes that A/B reproducible
+        # on any box with both assemblers -- no card, no model weights.
+        if: always()
+        working-directory: ${{ github.workspace }}
+        run: |
+          set -uo pipefail
+          out=pto-dump
+          mkdir -p "$out"
+          found=$(find pypto-serving/build_output -name '*.pto' 2>/dev/null | wc -l)
+          echo "collected $found .pto file(s)"
+          if [ "$found" -gt 0 ]; then
+            find pypto-serving/build_output -name '*.pto' \
+              -exec cp --parents {} "$out"/ \;
+          fi
+
+      - name: Publish the .pto dump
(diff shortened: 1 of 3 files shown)


luohuan19:
Closing — this arm has reported, and it is the one that changed the conclusion.

**What it established.** This branch is #2735 plus 28 lines of CI (a `.pto`
artifact upload) — same pypto code, same v0.61 pin — and it **passed** the Qwen3
accuracy guard that #2735 failed. One red and one green on identical content means
the corruption is **intermittent**, which no single run per arm could have shown.

**What its artifact enabled.** The upload captured 2911 `.pto` files — every kernel
serving compiles for Qwen3 and DeepSeek. Since pypto emits the `.pto` and ptoas
consumes it, the same pypto emits byte-identical `.pto` under either pin, so the
entire v0.60/v0.61 difference is `ptoas(.pto)` — reproducible on any box with both
assemblers, no card and no model weights. Assembling all of them both ways and
comparing op sequence, operand types, `set_flag`/`wait_flag`/`pipe_barrier`, and
dynamic views' constructor values found **2861 of 2861 structurally identical**.
v0.61's only difference is extra `GlobalTensor` declarations that no op consumes —
dead code from `667d43e4 simplify memref view materialization`.

That comparison was itself validated by mutation: five changes that would each
corrupt a kernel (sync event id, pad value, pipe barrier, tile shape, layout) were
all detected through the same pipeline, so the 2861/2861 is not an artifact of a
lossy comparison.

**One probe here was malformed** and proved nothing: `--enable-insert-sync` and
`--enable-inject-barrier-all-sync` are mutually exclusive in ptoas, and the first
push appended the second while leaving the first, so every kernel failed to
compile. It was dropped rather than fixed once the static comparison made it
unnecessary.

**Where it ended up:** the pin does not change what serving runs. #2686 is now
fully green, with DSpark dropped (it is red upstream too) and the context-parallel
pypto-lib prefill examples given the second card their `# ci: devices=2` markers
call for.

PR #2714, opened 2026-09-09
feat(distributed): register INT8 variant for builtin all_to_all_v

georgebisbas:
## Summary
- Register **INT8** alongside FP32 for managed `pld.tensor.all_to_all_v` on both HOST and L2 rails (variant mangling, lowering synthesis, codegen).
- Fix the shared kernel template tile UB assignment to `0x0` (dtype-agnostic; the prior `0x10000` was not INT8-safe and `static_cast<float>(0)` is invalid for `TASSIGN`).
- Document the FP32/INT8 contract and add UT coverage for INT8 kernel source generation and L2 lowering.

Prerequisite for RFC #2521 / A1 AllToAllV benchmark (canonical DSV4 dispatch payload is INT8 `4160 B/route`).

## Test plan
- [x] `pytest tests/ut/codegen/distributed/test_builtin_collective_kernel_source.py -q`
- [x] `pytest tests/ut/ir/transforms/test_lower_l2_tensor_collectives.py::test_int8_synthesizes_int8_variant -q`
- [x] pre-commit on all touched files (headers, docs parity, clang-format, cpplint, ruff, pyright)

PR #2738, opened 2026-09-10
perf(collectives): flat single-AIV TPUT for all_to_all_v (RFC #2521 K1)

georgebisbas:
## Summary

Implements **RFC #2521 work item K1**: rewrite the hand-written `builtin.tensor.all_to_all_v` kernel (HOST/L2 rail) to issue **one `TPUT` per destination** instead of a caller-managed per-chunk loop bracketed by two `pipe_barrier(PIPE_ALL)` calls per chunk.

Also extends **`InsertCommFence`** (pass 51) so L2/HOST pipelines that split **collective dispatch** and **InCore consume** across separate AIV tasks get automatic whole-GM cache coherency at the consumer entry, and makes HOST/L2 **`recv_counts` authoritative via a post-barrier local fill** (self from register, peers by pulling staged `send_counts`) — together fixing CI `recv_counts=0` flakes (self and peer slots) without hand-written test harness fences.

**8 files**: kernel template, IR pass, en/zh pass-51 docs (+ index), tensor_ops docstring, and InsertCommFence UTs.

**Rebase note (2026-09-14):** rebased onto current `main` after [#2714](https://github.com/hw-native-sys/pypto/pull/2714) (INT8 `all_to_all_v`) merged. Sole conflict in `kernel.cpp.in` resolved as **best of both**: keep K1's fixed `TileData send_tile(1, kTileCount)` width (flat single-TPUT-per-peer) **and** adopt #2714's dtype-agnostic `TASSIGN(send_tile, 0x0)` (replaces FP32-only `0x10000`). Tip: `ee47ddc94`.

---

## Connection to #2521 roadmap

[#2521](https://github.com/hw-native-sys/pypto/issues/2521) is the umbrella RFC for L2-orchestrated, runtime-selected multi-AIV collectives. The AllToAllV execution roadmap (Yunji, 2026-09-01) splits work into three tracks:

| Track | Item | Status |
|-------|------|--------|
| **O** (L2 sinking) | O1 — L2-managed `all_to_all_v` | ✅ merged (#2635) |
| **O** | O2 — dynamic `core_num` scalar end-to-end | not started |
| **K** (kernel rework) | **K1 — flat single-AIV TPUT + start sync** | **this PR** |
| **K** | K2 — multi-AIV peer×length lane split | blocked on K1 |
| **K** | K3/K4 — tile tuning, zero-length lanes, aligned tails | planned |
| **A** (benchmark/demo) | A1 harness, A2 DSV4 MoE demo, A3 EP tuning | A1 in progress |

**What this PR is:** the first K-track kernel change — removes redundant caller-side chunking/barriers so K2 can add parallel lanes on a clean single-TPUT-per-peer structure — plus the compiler coherency fix CI required for the L2/HOST ST harness.

**What this PR is not:** multi-AIV bandwidth scaling (K2), runtime launch-width selection (O2), or the formal perf benchmark (A1). Those remain follow-ups.

Prerequisites on `main` that this PR **preserves** (does not re-litigate):

- **C0** contract freeze (exact-traffic, signal protocol)
- **#2536**: `rows = clamp(send_counts[dest], 0, max_recv)` — only valid prefix crosses the wire
- **O1**: L2 lowering path for `all_to_all_v`

---

## Problem (why K1 exists)

The current kernel doesn't just chunk — it **serializes every chunk**:

```cpp
// OLD (removed): per 256-element slice
pipe_barrier(PIPE_ALL);
pto::comm::TPUT(...);
pipe_barrier(PIPE_ALL);
```

For the DSV4 dispatch anchor (~24.375 KiB/peer, `row_numel=4160` INT8), that is roughly **~98 TPUT calls + ~196 caller-side barriers per peer**. `TPUT_IMPL` already re-chunks internally via `TputProcessSlice`; the hand-written loop was redundant and doubled the barrier count at the caller.

---

### `kernel.cpp.in` — core K1 rework

1. **One TPUT per dest** over full `block_numel = rows * row_numel`. Internal chunking stays inside `TPUT_IMPL` (same 256-element granularity when `block_numel > kTileCount`).

2. **`ExchangeBarrier()` helper** — self-clearing N-way rendezvous on `signal_base`:
   - **Start-of-call**: prevents persistent-window races (`reset_persistent_windows=False`) where round N+1 TPUT starts before all ranks finish round N.
   - **End-of-call**: completion release (replaces inlined barrier loops).

3. **`PublishFence()`** — `pipe_barrier(PIPE_ALL)` + `dsb(DSB_DDR)` before completion barrier (RFC publish-fence contract).

4. **Fixed `kTileCount` staging tile** (`TASSIGN(..., 0x0)` after #2714) + **per-dest `ColMaskInternal` reset** before each TPUT:
   - **(a)** `GetValidCol()` sizes the hardware fast path — without capping to `block_numel`, short payloads (e.g. `rows==1`) DMA a full tile width off GM (sim-invisible; fails on real a2a3).
   - **(b)** `TputProcessSlice` mutates the caller tile's `ColMaskInternal` on ragged tail chunks and never restores it; `send_tile` is reused across the dest loop — stale mask shrinks later destinations (order-dependent perf drift).

5. **Authoritative post-barrier `recv_counts` fill (all slots)** — Phase 1 remote `TNOTIFY(Set)` of peer counts is kept only as best-effort / InCore wire parity. Remote `NotifyOp::Set` is a posted GM store; the completion barrier's `AtomicAdd` on `signal` does **not** prove that Set is visible (CI: `rank=0 src=1 recv_counts=0`). After the end `ExchangeBarrier`, every local slot is rewritten: self from register-captured `self_rows`; peers by `dcci` + load of that peer's staged `send_counts[my_rank]` with the same two-sided clamp; then local store + local `TNOTIFY(Set)` + trailing fence. Consume-side Phase B invalidate then sees locally written values.

6. **Comments** documenting: double in-call barrier reuse safety, early-return deadlock risk if guards diverge across ranks, ping-pong non-applicability (`Rows=1` fails `canSplitTile` gate), and why Phase 1 count Set is not authoritative on HOST/L2.

#### Motivation

CI skew / uniform STs failed with `recv_counts=0 != clamped(N)` on real NPUs. Two cooperating causes:

1. **Stale GM cache reads across separate AIV tasks** (consume after collective) — fixed by Phase B whole-GM invalidate at consumer entry.
2. **Peer count publish visibility** — remote `TNOTIFY(Set)` of `recv_counts` is not ordered with the barrier `AtomicAdd` on `signal`; Phase B alone cannot create a value that has not landed yet — fixed by the kernel's post-barrier local fill (item 5 above).

- **InCore intrinsic** tests pass because `consume_step` reads `recv_counts` in the **same InCore function** right after `pld.tensor.all_to_all_v(...)` — `InsertCommFence` phase A already covers that path.
- **L2/HOST** tests split work across orchestration dispatches:
  - L2: `chip_pipeline` → `all_to_all_v` kernel task → `consume_step` InCore task
  - HOST: global `all_to_all_v` → per-rank `consume_orch` → `consume_step`
- **`InsertCommFence` phase A is InCore-only** — it cannot insert markers into orchestration bodies (orchestration codegen rejects `system.cacheinvalid` / `system.fence`).

A manual region-scoped `cacheinvalid(recv_counts, ...)` in test `consume_step` helped locally but was **too weak under batch CI** (132 tests, shared window reuse, long post-fence retirement gaps).

#### Design

Convert `InsertCommFence` from a per-function pass to a **program pass** with two phases:

| Phase | Scope | Action |
|-------|-------|--------|
| **A** (existing) | InCore bodies | Insert markers after local/remote/opaque writes and after waits |
| **B** (new) | `Orchestration` / `Graph`, **and** HOST `Opaque` + `Role::Orchestrator` (real `host_orch`) | Scan for opaque publish → later InCore consume; prepend prologue to **consumer callee entry** |

Phase B **does not modify orchestration IR**. Markers land on the InCore consumer kernel where codegen can emit them — same pattern as the existing opaque cross-InCore-call rule (`test_opaque_cross_function_call_gets_whole_gm_marker`).

#### Opaque publish detection (orchestration dispatch)

A statement sets `seen_publish` when it dispatches:

- Any `builtin.tensor.*` call (HOST lowered collectives after `LowerHostTensorCollectives`)
- Any callee with `builtin_template_dir` (L2 synthesized AIV kernels after `LowerL2TensorCollectives`, e.g. `__builtin_all_to_all_v__fp32`)
- Any `Submit` (`pl.submit` / `pl.manual_scope`)

Regular InCore helpers (`stage_step`, `fill_counts_step`) are **not** flagged — only true opaque GM-publishing task launches. The scan recurses into `ScopeStmt` bodies so nested launches remain visible.

#### Consumer marking

When `seen_publish` is true and a later statement dispatches an InCore callee — plain `Call` **or** `Submit` — either directly or through a one-hop orchestration wrapper, the pass records that InCore function and prepends:

```python
pl.system.cacheinvalid()   # whole-GM
pl.system.fence()
```

at function entry. Idempotent via `HasConsumePrologue()` — skips if already present.

One-hop unwrap (`ResolveSingleInCoreDelegate`) peels `ScopeStmt` / singleton `SeqStmts` and accepts the post-`MaterializeRuntimeScopes` wrapper shape `scope { t = consume_step(...); return t }` (exactly one InCore callee). That is required because the default pipeline runs `MaterializeRuntimeScopes` **before** `InsertCommFence`, so a bare `return self.consume_step(...)` is no longer what the resolver sees.

#### Why whole-GM (not region-scoped)

Matches the existing InsertCommFence convention for **wait-then-read** and **opaque publish-then-read** inside InCore. Cross-task consume after a collective may read both `recv_counts` and peer-written `data` window slots; whole-GM is conservative and eliminates the batch flakes region-scoped invalidation left behind.

#### What this offers

- **Correctness by default** for any `collective → consume` orchestration pipeline without authors remembering NPU cache rules
- **Single source of truth** — coherency policy stays in pass 51, not duplicated in ST harness code
- **Extensible** — covers L2 `chip_pipeline`, HOST fan-out loops (including Opaque `host_orch`), Submit-based launches, and future collectives using the same pattern

#### Tradeoffs

- **Function-granular**: if `consume_step` is also called from a non-collective path, it still gets the prologue (correct, slightly conservative on perf)
- **`builtin.tensor.*` breadth**: all HOST collective builtins trigger marking (harmless extra flush; simpler than per-op whitelisting)

### `test_insert_comm_fence.py` — UT

Phase B coverage includes:

- Baseline orch collective → consume prologue
- HOST `builtin.tensor.*` publish → consume
- HOST Opaque + `Role.Orchestrator` (`host_orch`) → `consume_orch` → `consume_step`
- Submit-aware collective + consume (`pl.submit`)
- One-hop unwrap **after** `materialize_runtime_scopes` (Yunji P1 / real pipeline order)
- Stage-between / consume-before / if-branch / idempotence cases

Full file: **37** tests.

### Docs (`docs/en|zh/dev/passes/51-insert_comm_fence.md`)

Documents Phase B eligibility (incl. HOST Opaque orchestrator), Submit awareness, and RuntimeScope unwrap.

### `tensor_ops.py` — docstring only

Documents that HOST/L2 kernel exchanges the barrier signal **twice within one call** (start + end).

### ST tests (HOST + L2)

No manual fences in `consume_step` — coherency is compiler-inserted (+ kernel post-barrier authoritative `recv_counts` fill). Tests assert behavior only.

---

## Expected benefit (K1 kernel)

DSV4 dispatch anchor: ~24.375 KiB/peer → `ceil(24960/256)=98` chunks.

| Metric | Before (K0) | After (K1) |
|--------|-------------|------------|
| Caller-side `pipe_barrier` **in the per-peer transfer loop** | ~196 (`2×98` chunk brackets) | **0** |
| TPUT calls per peer | ~98 | **1** (+ internal chunks) |
| TPUT_IMPL internal `pipe_barrier` per peer | ~98 | ~98 (unchanged; moved entirely inside TPUT_IMPL) |
| Fixed overhead per kernel call (vs K0) | end `pipe_barrier`+`dsb` + trailing `pipe_barrier` | mainly **+1 start `ExchangeBarrier`**; `PublishFence` / end `ExchangeBarrier` / trailing barrier roughly replace K0's end-fence path; plus **O(P) scalar** post-barrier `recv_counts` fill (`dcci`+load peer `send_counts` / local store) — negligible vs transfer path |
| Intra-tile ping-pong | N/A (`Rows=1`) | unchanged |

Net transfer-path win: **eliminates ~196 duplicate caller-side chunk brackets** per peer; internal chunk barriers stay ~98. “After: 0” is the dest transfer loop only — the kernel still has fixed `pipe_barrier`/`ExchangeBarrier` outside that loop.

Further bandwidth gains require **K2** (multi-AIV lanes) or explicit two-tile PUT — out of scope here.

---

## Perf check (quick L2 swimlane, P=2)

Measured `__builtin_all_to_all_v__fp32` AIV exec (`chip_swimlane` `func_id=1`) on devices [6, 7], swapping only `kernel.cpp.in` between `origin/main` (`8ecc5d19f`) and this PR (`8da1fa989`). Methodology: 3 warmup (no swimlane) + 8 measure runs; p50 across measure means (both ranks).

| Case | block_numel | main p50 | PR p50 | Δ | speedup |
|------|-------------|----------|--------|---|---------|
| small_4k | 4096 | 700 µs | 399 µs | -43% | **1.76×** |
| mid_24k | 24576 | 3479 µs | 2267 µs | -35% | **1.53×** |
| anchor_25k (DSV4-ish) | 24960 | 2073 µs | 1146 µs | -45% | **1.81×** |

Static `pipe_barrier(PIPE_ALL)` occurrences in the template: main **7** → PR **6**.

**No regression on p50** — PR is ~1.5–1.8× faster across sizes. Run-to-run spread is still large (some samples ~0.15–4.5 ms), so treat as **directional**, not a formal A1 number. Post-barrier O(P) `recv_counts` fill does not show up as a regression vs K0's caller chunk brackets.

## Code diff vs `origin/main` (kernel shape)

Whole PR: **8 files**, +1269 / −67 (mostly InsertCommFence UTs/pass). Kernel alone: **+151 / −44** (224 → 331 lines).

| | `origin/main` (K0) | this PR (K1 + coherency) |
|--|--|--|
| Transfer loop | Per dest: **chunk loop** `base += tile_cols` with `pipe_barrier; TPUT; pipe_barrier` | Per dest: **one** `TPUT(block_numel)` (internal chunking in `TPUT_IMPL`) |
| Staging tile | `tile_cols = min(row_numel, 256)`; `TASSIGN(..., 0x0)` post-#2714 | Fixed `kTileCount=256` + `TASSIGN(..., 0x0)`; `ColMaskInternal = min(block_numel, 256)` each dest |
| Start sync | none | `ExchangeBarrier()` |
| End sync | inlined notify/wait/clear + `pipe_barrier` | `PublishFence()` + `ExchangeBarrier()` |
| `recv_counts` | remote `TNOTIFY(Set)` in dest loop only | same best-effort Set **+** post-barrier **local fill** (self=`self_rows`, peers=pull staged `send_counts`) |

**Removed (main transfer path):**
```cpp
for (base = 0; base < block_numel; base += tile_cols) {
  pipe_barrier(PIPE_ALL);
  TPUT(... chunk ...);
  pipe_barrier(PIPE_ALL);
}
```

**Added (PR transfer + count fill):**
```cpp
ExchangeBarrier(...);                 // start
for (dest...) {
  TNOTIFY(Set counts);                // best-effort
  if (block_numel > 0) TPUT(... full ...);  // no caller barriers
}
PublishFence();
ExchangeBarrier(...);                 // end
for (src...) { /* authoritative local recv_counts */ }
```

Also in this PR (not the TPUT path): InsertCommFence **Phase B** (+~295 pass / +~763 UT lines) prepends whole-GM `cacheinvalid`+`fence` on InCore consumers after orch/HOST collective publish.

### Kernel (K1)

1. **Start-of-call `ExchangeBarrier` on same signal window as end-of-call** — inline comment argues bounded ≤1-occurrence drift; NOTIFY from occurrence K cannot satisfy WAIT from K+1.
2. **`ColMaskInternal = min(block_numel, kTileCount)` per dest** — must be reset every iteration; bare `kTileCount` alone would re-break short-payload correctness.
3. **Post-barrier authoritative `recv_counts` fill** — self from `self_rows`; peers pull staged `send_counts[my_rank]` (not remote Phase-1 `TNOTIFY(Set)`). Phase 1 Set is best-effort only; consume still needs Phase B invalidate for cross-task cache.
4. **InCore rail parity** — InCore composite still has single end barrier (no start-of-call sync). HOST/L2 now has both. Persistent-mode soak may expose a rail gap; not changed in this PR.
5. **Multi-block launch** — comment notes missing intra-launch block sync if `block_num > 1`; current paths require `core_num == 1`.

### Compiler (InsertCommFence phase B)

1. **Sequential scan semantics** — `seen_publish` threads through `SeqStmts` so `stage_step → all_to_all_v → consume_step` marks only `consume_step`, not `stage_step`.
2. **If/else branch merge** — uses OR of branch flags (conservative for code after the if).
3. **HOST Opaque + one-hop unwrap after MaterializeRuntimeScopes** — real `host_orch` / `consume_orch` shape; deeper than one hop would still need extension.
4. **Submit-awareness** — both publish and consume via `Call` or `Submit`.

---

## Verification

- [x] `pre-commit` on changed files
- [x] Sim UT (`all_to_all_v`-related): pass (post-#2714 rebase includes fp32 + int8 kernel-source coverage)
- [x] InsertCommFence UT: **37/37** pass (Phase B + materialize/HOST/Submit coverage)
- [x] NPU P=2 STs (`--platform=a2a3`): local stress green after Phase B + post-barrier local `recv_counts` fill
  - Earlier: L2 `2-over_capacity` + HOST `2-one_and_full`: **0/30** (Phase B + self reinforce)
  - After peer-slot flake on uniform L2 `test_l2_all_to_all_v[2]`: pull-based fill — uniform + L2 `over_capacity` + HOST `one_and_full`: **0/20**
  - **2026-09-14 post-rebase** (192.168.150.12, local/unisolated, devices 4,5): L2+HOST **10/10** pass
- [x] NPU P=4 gate (2026-09-14, devices 4–7, local/unisolated): green; one L2 `zero_and_full[4]` SCHEDULER_TIMEOUT under contention passed on isolated retry
- [x] Quick L2 swimlane / directional perf (P=2) — see Perf check above; **2026-09-14** re-check on devices 4,5 (`ee47ddc94` vs main kernel): mid_24k **~1.76×**, anchor_25k **~1.80×** (directional)
- [x] Formal perf sweep vs plan 110 baseline (A1 harness) — still open

---

## Follow-ups (not in this PR)

- **K2**: peer×length lane partitioning on top of this flat kernel
- **O2**: runtime `core_num` scalar plumbing
- **A1/A2**: benchmark harness + DSV4 MoE integration demo
- NPU perf numbers quantifying the ~196 caller-side barrier elimination

Relates to #2521.


YunjiQin:
Follow-up on two correctness findings from review of `e78f9d6`, rechecked against the current head `30953c5`:

1. **[P1] Scope handling: scanner fixed, but the consumer-wrapper resolver still misses materialized scopes.**

   The new `As<ScopeStmt>` recursion in `OrchPostCollectiveScanner::ScanStmt` addresses the original scanner issue, including `RuntimeScopeStmt` and `CommDomainScopeStmt`. However, [`ResolveSingleInCoreDelegate`, lines 338–352](https://github.com/hw-native-sys/pypto/blob/30953c57ef499c2535516cfc2a42a3a64c0b58e6/src/ir/transforms/insert_comm_fence_pass.cpp#L338-L352) still accepts only a bare return/eval, optionally inside a singleton `SeqStmts`.

   In the default pipeline, `MaterializeRuntimeScopes` runs before `InsertCommFence` and wraps orchestration bodies. Thus the supported HOST pattern `collective -> consume_orch -> consume_step` reaches this resolver with `consume_orch.body` wrapped in `RuntimeScopeStmt`; the resolver returns null, so `consume_step` receives no entry invalidation. Scanning `consume_orch` independently cannot recover this, because its scan starts with `seen_publish=false`. The stale-cache-read risk remains on this cross-wrapper path.

   Please unwrap scope nodes in the delegate resolver as well, and add a regression that runs scope materialization before this pass (ideally a full lowering/compile test of the HOST wrapper pattern), rather than testing only the unwrapped input IR.

2. **Submit consumer recognition: addressed in the current head.**

   Previously, `MarkPostCollectiveInCoreConsumers` used `As<Call>` and skipped a consumer launched through `pl.submit(...)`, even when it depended on the preceding collective. The current `GetCallLikeOp` handles both `Call` and `Submit` and is used by consumer recognition, so this specific finding is resolved at source level. It should not be treated as an outstanding defect.

This follow-up is based on static inspection; I have not independently run the updated PR's build or NPU tests.

georgebisbas:
@YunjiQin Thanks — we hit the same gap independently while chasing the HOST/L2 `recv_counts=0` flakes and fixed it in parallel before this review landed.

[`3903dbfee`](https://github.com/hw-native-sys/pypto/commit/3903dbfee132158b460afecdbc9696132672032d) addresses **P1** specifically:

- `ResolveSingleInCoreDelegate` now peels `ScopeStmt` / singleton `SeqStmts`, and accepts the post-`MaterializeRuntimeScopes` wrapper shape `scope { t = consume_step(...); return t }` (exactly one InCore callee).
- Phase B also scans HOST `Opaque` + `Role::Orchestrator` bodies (real `host_orch`), not only `Orchestration`/`Graph`.
- Regression: `test_orch_consume_wrapper_unwraps_after_materialize_runtime_scopes` runs `derive_call_directions` → `materialize_runtime_scopes` → `insert_comm_fence`. Also added `test_host_opaque_orchestrator_collective_then_consume`.

Item 2 (Submit consumer recognition) agreed — already covered by `GetCallLikeOp` on the prior head.

Ref: https://github.com/hw-native-sys/pypto/pull/2738#issuecomment-5633282770

georgebisbas:
### Perf + kernel diff notes (main vs this PR)

Quick L2 swimlane check on devices [6, 7]: swap only `kernel.cpp.in` between `origin/main` (`8ecc5d19f`) and HEAD (`8da1fa989`). `3` warmup + `8` measure runs; p50 of per-dispatch builtin exec (`func_id=1`).

| Case | block_numel | main p50 | PR p50 | Δ | speedup |
|------|-------------|----------|--------|---|---------|
| small_4k | 4096 | 700 µs | 399 µs | -43% | **1.76×** |
| mid_24k | 24576 | 3479 µs | 2267 µs | -35% | **1.53×** |
| anchor_25k | 24960 | 2073 µs | 1146 µs | -45% | **1.81×** |

Static `pipe_barrier(PIPE_ALL)` in template: **7 → 6**. No p50 regression (~1.5–1.8× faster). Spread is still large (~0.15–4.5 ms) — directional, not formal A1.

#### Kernel shape vs `origin/main`

| | main (K0) | this PR |
|--|--|--|
| Transfer | per-dest chunk loop + 2×`pipe_barrier` per chunk | one `TPUT(block_numel)` / dest |
| Start sync | none | `ExchangeBarrier` |
| End sync | inlined barrier | `PublishFence` + `ExchangeBarrier` |
| `recv_counts` | remote `TNOTIFY(Set)` only | Set (best-effort) + post-barrier **local fill** (self register / peer pull `send_counts`) |

Also: InsertCommFence Phase B (consume-entry whole-GM invalidate) for L2/HOST cross-task coherency.

Mirrored in the PR description under **Perf check** / **Code diff**.

YunjiQin (commented):
## Review

Split into blocking / should-fix / nits. The two blocking items are design-level: one changes the op contract, the other is the Phase B mechanism itself.

### Blocking

#### B1. Phase 2's peer pull reads a non-window operand through a window pointer

`templates/kernel.cpp.in:310-325`

`send_counts` is **not** window-bound by contract. The deducer states the asymmetry explicitly:

```cpp
// src/ir/op/distributed/collective.cpp:716
CHECK(counts_type) << "send_counts must be a Tensor or DistributedTensor, got "
// :743
CHECK(recv_type)   << "recv_counts must be a DistributedTensor (window-bound), got "
```

and the L2 rail declares it a plain Tensor in the synthesized signature, on purpose:

```cpp
// lower_l2_tensor_collectives_pass.cpp:170-185
{"input", "target", "signal", "send_counts", "recv_counts"},
// `input` and `send_counts` are Tensor-like on the public op ... The three
// genuinely window-bound operands stay distributed ...
{false, true, true, false, true},
```

`CommRemotePtr` computes `windowsIn[pe] + (local_ptr - windowsIn[my_rank])`, which is meaningless for a pointer outside the window. Nothing on the L2 rail narrows `send_counts` to a window, so a legal program reads an arbitrary offset inside the peer's window; the two-sided clamp then turns that garbage into a plausible row count in `[0, max_recv]` — silent corruption, and harder to spot than the `recv_counts=0` it replaces.

The HOST rail happens to reject a plain Tensor today (`host_bound_args = {{0,"input"},{3,"send_counts"}}`, `lower_host_tensor_collectives_pass.cpp:545`), but for an unrelated, documented reason: *"the HOST builtin path requires both to be window-bound (`EmitBuiltinWindowCollectiveDispatch` has no dispatch arm for a plain Tensor)"* (:351-360). Closing that gap — the natural fix for that limitation — silently breaks the kernel.

The pull is also unsynchronized on its own terms: it runs **after** the end barrier, so a peer may already have returned and had its round N+1 `stage_step` overwrite that window before this rank reads it. Moving it between the two barriers fixes the race but not the operand-class problem.

Whatever replaces it has to publish visibility through the three cross-rank-addressable operands only — `target`, `signal`, `recv_counts`. One option: an extra `signal` column used as a per-peer "counts published" counter, so the count's visibility rides the same address family as the barrier itself, rather than relying on a remote `NotifyOp::Set` to a different address being ordered against the barrier's `AtomicAdd`.

Removing this block also removes three other issues that live only inside it:

- `dcci(..., SINGLE_CACHE_LINE)` (:318) — the identifier is only defined inside `struct cache_line_t` in pto-isa's `cpu_stub.hpp:127-131`; every other use in both trees writes `cache_line_t::SINGLE_CACHE_LINE`, and `reduce_scatter/templates/kernel.cpp.in:44-70` carries an explicit `__CPU_SIM` shim that this template lacks. Also, `pto::comm::detail::DcciSignal` (`TNotify.hpp:22-27`) already wraps this exact asm/dcci/asm triple.
- `__asm__ __volatile__("")` with no `"memory"` clobber around a non-volatile `*remote_send_slot` load (:317-320) — nothing stops the load being hoisted above the `dcci`.
- The clamp duplicated verbatim from Phase 1 (:320-323 vs :236-240), which the PR's own comments require to stay bit-for-bit identical with `LowerTensorAllToAllVRule`.

Please also re-test whether the register-captured **self**-slot fill is still needed once the consumer-side invalidate is in place — that flake plausibly had the same root cause as the peer one.

#### B2. Replace the Phase B orchestration scan with a local per-InCore rule

`insert_comm_fence_pass.cpp:332-500`

The scan approximates "who depends on whom" with a sticky `seen_publish` boolean over statement order. That approximation is simultaneously too wide and too narrow:

- `IsOpaquePublishingDispatchExpr` returns true for **any** `Submit` (:332). Every `pl.manual_scope` compute pipeline therefore gets a whole-GM flush prepended to every InCore callee after its first submit — `tests/st/runtime/scheduling/test_manual_scope_pipeline.py`, `test_phase_fence_dep_compression.py` (28 submits), `test_pl_at_deps_pipeline.py`, `examples/utils/phase_fence_dep_compression.py`. None of them involves a collective.
- `IsPhaseBScanTarget` (:344) excludes the `Group` / `Spmd` wrappers synthesized by `OutlineClusterScopes` (:321/376), so a `collective → pl.spmd consumer` pipeline gets no prologue at all — silently.
- `ResolveSingleInCoreDelegate` (:419) returns null on 2+ InCore delegates, and `CollectInCoreCallees` does not recurse into `ForStmt` / `IfStmt` / `WhileStmt`, so a `for r in pl.range(n): self.consume_step(...)` wrapper yields nothing.
- `MarkPostCollectiveInCoreConsumers` (:429) does not exclude `builtin_template_dir` AIV functions, so a body documented as *"carries no PyPTO-generated kernel body"* (`function.h:486`) gets rewritten.
- Loops are scanned once (:473), so a loop-carried publish→consume dependency is missed.
- Marking lands on the **function definition**, so a marked callee carries the prologue at every call site, including ones before any publish. (The header says *"the next InCore callee is recorded"*; `seen_publish` is sticky, so it is actually every subsequent one.)

A purely local rule does the same job with none of these questions:

> At an InCore function's entry, prepend a whole-GM `cacheinvalid` iff its body performs a **scalar read on a `DistributedTensor`**.

A window operand is by definition memory a peer may have written remotely, and a cached scalar load on it is the only read that can observe a stale line — which is precisely what this pass already asserts elsewhere:

```cpp
// True if evaluating `expr` can perform a memory read (a cached GM scalar load).
// ... (an InCore `tensor.read` lowers to a cached `pto.load_scalar`).
bool ExprMayRead(const ExprPtr& expr) {
  if (IsOp(call, "tensor.read")) return true;
```

It classifies both ST functions correctly with zero orchestration analysis:

| function | `DistributedTensor` params | scalar read target | marked |
| --- | --- | --- | --- |
| `stage_step` | `stage`, `counts` | `pl.read(counts_row, ...)` — a plain `Tensor` | no (it is the publisher) |
| `consume_step` | `data`, `recv_counts` | `pl.read(recv_counts, ...)` — a `DistributedTensor` | yes |

and it excludes the synthesized builtin function for free — its body is a lone `ReturnStmt`, so there is no read to match.

Match on the **type**, not on param identity, so rebound SSA vars are covered; `As<DistributedTensorType>` is the correct exact-kind query here per `.claude/rules/ir-kind-traits.md`. This also lets `InsertCommFence` revert to a **function pass** — the program-pass conversion existed only to enable the scan.

One note: a function that also contains a `pld.system.wait` would get two whole-GM invalidates (entry, plus Phase A's post-wait one). Neither ST `consume_step` has a wait — the barrier lives inside the collective kernel — so the common path does not double up. Deduplicating (skip the entry prologue when the scalar read is dominated by a wait) is an optimization, not a correctness matter.

### Should fix

#### S1. The `fence` in the consume prologue is redundant

`insert_comm_fence_pass.cpp:305`

This file's own header states the consume-side contract as invalidate-only:

```
 *   - after each **wait**: a no-arg (whole-GM) `system.cacheinvalid`.
 *   - **notify**: nothing.
 *   wait; read  ->  wait; cacheinvalid(); read
```

The fence belongs to the publish side (`store(win); cacheinvalid(win); fence; notify`), where it orders writes before the notify that releases them. At a consumer's entry there are no prior writes of this task to order. Drop it — and simplify `HasConsumePrologue` (:296) to test `stmts_[0]` only, or already-marked functions get the prologue a second time.

#### S2. `PublishFence()` is expanded verbatim at the kernel tail

`templates/kernel.cpp.in:329-330` is the same two lines as the helper introduced at :46-49. If the publish-fence recipe ever changes, this copy is missed.

### Nits

- **`python/pypto/ir/op/distributed/tensor_ops.py:492`** — the docstring documents the twice-per-call signal exchange but not that the call now **blocks at entry**: a rank that skips the dispatch (predicate, divergent guard) leaves every peer hung in `TWAIT`. DSL users see the docstring, not the kernel comments.
- **`tests/ut/ir/transforms/test_insert_comm_fence.py:1139`** — hand-rolls `backend.reset_for_testing()` + `set_backend_type` + `try/finally`; `tests/ut/conftest.py:93-109` provides the `ascend_backend` fixture explicitly to replace that pair. Using it also removes the two single-use imports that needed the ruff `PLC0415` fix.
- Once B2 lands, the 37 UTs need re-basing on the new criterion (with/without a `DistributedTensor` param, scalar read vs write-only, `tile.load` vs `tensor.read`, idempotence). The orchestration-shape tests — HOST Opaque, one-hop unwrap, Submit-awareness, post-`MaterializeRuntimeScopes` ordering — no longer describe anything the pass does.

### Not an issue

The start-of-call `ExchangeBarrier` relying on a zeroed `signal` window is a normal collective-API precondition, and the docstring already states the other half of it ("self-clearing ... safe to reuse inside a `for`/`while` loop"). The "window memory is not guaranteed zeroed" comment at :216-219 is about the `target` payload window's unwritten tail, which is unrelated.

### On the perf numbers

The 1.5–1.8x on the K1 transfer path looks solid and the barrier accounting in the table matches the diff. Worth re-measuring after B1 is resolved, since the replacement for the Phase 2 fill may sit on the critical path differently.

georgebisbas:
Hi Yunji — thank you for the detailed review. We've addressed both blocking items (**B1** illegal `send_counts` CommRemotePtr, **B2** InsertCommFence Phase B orch scan) plus the bundled **S1/S2** nits. Summary below; happy to iterate if you'd prefer a narrower escalation on B1.

Branch: `feat/all-to-all-v-flat-tput` (K1). Changes are local / pending push on this fork branch.

---

## Verification (NPU + UT)

| Gate | Result |
|------|--------|
| `tests/ut/ir/transforms/test_insert_comm_fence.py` | 34 passed |
| `tests/ut/codegen/distributed/test_builtin_collective_kernel_source.py` | passed (kernel source checks) |
| NPU **P=2** L2+HOST all_to_all_v (devices 6,7) | **10 passed / 10 skipped** (P=4 parametrizations skipped when only 2 devices) |
| NPU **P=4** L2+HOST all_to_all_v (devices 4,5,6,7) — **full clean rerun** | **20 passed** in 979s (`EXIT_P4=0`) |
| Directional L2 swimlane perf (P=2, devices 4,5, W=3/M=8) | PR faster on all three peer sizes (see Perf below) |

Skew cases covered on both L2 and HOST rails: `negative`, `one_and_full`, `over_capacity`, `zero_and_full`.

---

## B1 — Remove illegal `send_counts` CommRemotePtr; window-only count authority

### What you asked for

1. **Delete** all Phase-2 `CommRemotePtr(..., send_counts_base + …, …)` — `send_counts` may be a plain L2 `Tensor`, not window-bound.
2. **Phase 1** (unchanged wire): Set peer `recv_counts[my_rank]` via window `CommRemotePtr(recv_counts_base + my_rank, dest)`.
3. **Phase 2** (after publish fence + completion barrier): authoritative fill from **local window only** — self from `self_rows`, peers from **local** `recv_counts_base[src]` (value peer wrote in Phase 1), with cache visibility (`dcci` + no hoist).
4. **Escalation** if NPU still sees `recv_counts=0`: add signal **count-ready column** (`signal.shape[1] >= 2`), AtomicAdd when count is posted, Wait before Phase 2 local load — **without** reintroducing `send_counts` CommRemotePtr.
5. **S2**: deduplicate tail publish fence into `PublishFence()` helper.

### What we implemented

**Step 1 — Zero remote `send_counts` access** ✅

All `CommRemotePtr(..., send_counts_base + …)` blocks are removed. `send_counts_base` is used **locally only** (Phase 1 / Phase 2 clamp of row count per dest). Grep confirms no remote `send_counts` CommRemotePtr remains in `kernel.cpp.in`.

**Step 2 — Phase 1 window publish preserved** ✅

Per-dest loop still:
- clamps `send_counts_base[dest]` → `rows`
- captures `self_rows` for `dest == my_rank`
- `CommRemotePtr(recv_counts_base + my_rank, dest)` + `TNOTIFY(Set, rows)` (best-effort early publish)
- flat `TPUT` of payload block (with `ColMaskInternal` set per dest)

**Step 3 — Phase 2 publish + barrier** ✅

- `PublishFence()` after Phase 1 (covers all Sets + TPUTs)
- **Count lane publish** on signal column 1 (see escalation below)
- `PublishFence()` again
- `ExchangeBarrier(..., signal_cols)` on column 0 (unchanged self-clearing N-way rendezvous)
- Authoritative local `recv_counts_base[src]` rewrite + local `TNOTIFY(Set)` + final `PublishFence()`

**Step 4 — Self slot** ✅

Kept `self_rows` register path for `src == my_rank` in Phase 2 (per your “minimum: keep self slot write” guidance).

**S2 — `PublishFence()` helper** ✅

```cpp
AICORE inline void PublishFence() {
  pipe_barrier(PIPE_ALL);
  dsb(DSB_DDR);
}
```

Used at Phase 1→2 boundary, after count-lane publish, and at kernel tail (replacing duplicated barrier+dsb sequences).

### B1 escalation — signal column 1 (your Step 4)

We landed your escalation because the **literal** “Wait count-ready → load local `recv_counts_base[src]`” path still failed on **P=4 NPU** (`recv_counts=0` on skew cases) even after:

1. Moving count-ready **after** `PublishFence()` (so ready AtomicAdd cannot race ahead of Phase-1 recv_counts Set)
2. `dcci(SINGLE_CACHE_LINE)` + `asm volatile("" ::: "memory")` on the local recv_counts load

**Root cause (our read):** on a2a3, a count-ready `AtomicAdd` on the signal window and a separate `TNOTIFY(Set)` into `recv_counts` are independent GM posts. `TWAIT` on the ready lane proves the **ready post**, not that the **recv_counts Set** is cache-visible to the consuming AIV — hence P=4 still observed 0 in the recv_counts slot after Wait.

**Final protocol (still window-only; no `send_counts` CommRemotePtr):**

| Signal `[NR, S]`, `S >= 2` | Role |
|----------------------------|------|
| Column 0 | `ExchangeBarrier` (stride `signal_cols`) |
| Column 1 | **Count lane** |

After Phase-1 `PublishFence()`:
- For each peer `dest`: `AtomicAdd(signal[my_rank, 1], rows+1)` on dest's window  
  (`rows+1` bias so `rows==0` still satisfies `TWAIT >= 1`)

After second `PublishFence()` + column-0 barrier, Phase 2 peer fill:
- `TWAIT(signal[src, 1], >= 1)`
- `dcci` + load `*count_slot` → decode `rows = encoded - 1`, clamp
- `AtomicAdd(-encoded)` to clear lane for next call
- local store to `recv_counts_base[src]` + `TNOTIFY(Set)` (consume-side contract)

**Note on NotifyOp choice:** we tried `TNOTIFY(Set, rows+1)` on column 1; on a2a3 **`TWAIT` does not reliably wake on Set** (stream sync timeouts / poisoned lanes). Count lane **must** use `AtomicAdd` to pair with `TWAIT`, same address family as `ExchangeBarrier`.

**ABI:** enforced `signal.shape[1] >= 2` on public + builtin `all_to_all_v` (`collective.cpp`, host lower, composite lower); STs/UTs updated to `[nr, 2]` signal buffers.

### B1 alignment checklist

| Your requirement | Status |
|------------------|--------|
| No `CommRemotePtr(send_counts, …)` | ✅ |
| Phase 1 Set into peer recv_counts window | ✅ |
| Phase 2 window-only operands | ✅ (peer count from signal col 1, not remote send_counts) |
| Escalation via signal column if GM load flakes | ✅ (stronger: count **carried** on AtomicAdd lane, not just ready flag + GM load) |
| `PublishFence()` helper | ✅ |
| Do not require window-bound send_counts on L2 public op | ✅ (unchanged ST shape) |

**Question for you:** are you OK with carrying the clamped row count on signal column 1 (`AtomicAdd(rows+1)` / Wait / load / clear), rather than Wait-ready + local `recv_counts` GM load? It follows your escalation intent (signal column, same address family as barrier) but goes one step further because ready-only + GM load remained insufficient on P=4.

---

## B2 — InsertCommFence Phase B: local InCore rule, function pass

### What you asked for

1. **Drop** sticky `seen_publish` orchestration scan (too wide on any `Submit`, too narrow on Group/Spmd / loop-carried deps).
2. **Revert to function pass** (not program pass for Phase B).
3. **Rule:** at InCore entry, prepend whole-GM `system.cacheinvalid()` **iff** the body performs `tensor.read` on a **`DistributedTensor` parameter**.
4. **S1:** consume prologue is **invalidate-only** — no entry `system.fence`.

### What we implemented

- `InsertCommFence` is again a **`FunctionPass`** (`CreateFunctionPass`).
- Phase B helper `NeedsConsumePrologue(func)` walks the InCore body; matches `tensor.read` whose first arg is a **`DistributedTensor`-typed parameter** (covers rebound SSA aliases when the param var itself is read).
- When matched: prepend no-arg `system.cacheinvalid()` at function entry — **no** `system.fence`.
- Phase A (InCore publish-side region invalidate + fence after local/window writes, whole-GM invalidate after waits) unchanged.
- Removed all `seen_publish` / opaque Submit poisoning logic from Phase B.
- Updated `docs/en/dev/passes/51-insert_comm_fence.md` (+ zh) and rewrote Phase B tests in `test_insert_comm_fence.py` (prologue on `pl.read(recv_counts)`, none for plain Tensor read, idempotence, read under `if`, etc.).

### B2 alignment checklist

| Your requirement | Status |
|------------------|--------|
| No orch `seen_publish` scan | ✅ |
| Function pass | ✅ |
| Entry whole-GM invalidate iff DistTensor param + `tensor.read` | ✅ |
| Invalidate-only (no entry fence) | ✅ |
| Docs + tests | ✅ |

**Spot-check note:** we did not expand the criterion to `pl.load` on DistTensor (your review scoped to `tensor.read`; all_to_all_v consume uses `pl.read(recv_counts)`). If HOST allreduce/allgather load-only consumers regress, we can extend with NPU evidence.

---

## Perf re-measurement (post-B1 path change)

As you requested, we re-ran directional L2 swimlane after the B1 count path change.

**Methodology**
- Harness: `outputs/k1-perf-quick/compare_main_vs_pr.py`
- Swaps **only** `kernel.cpp.in` between `origin/main` and the PR working tree (same IR / runner / compile path otherwise)
- Measure: `__builtin_all_to_all_v` AIV exec time from `chip_swimlane_records` (`func_id=1`)
- Per (ref, case): fresh compile → warmup **without** swimlane → measure runs **with** swimlane; p50 across measure runs (mean of ranks per dispatch)
- Devices: **4,5** (P=2); warmup=**3**; measure=**8**
- Cases: full counts (`send_counts = max_recv` per dest); block sizes match roadmap peer sizes

**Results (p50 AIV µs; speedup = main/PR, >1 = PR faster)**

| Case | block_numel | main p50 | PR p50 | Δ | Speedup |
|------|-------------|----------|--------|---|---------|
| small_4k | 4096 | 765.1 µs | 493.8 µs | −271 µs (−35.5%) | **1.55×** |
| mid_24k | 24576 | 823.3 µs | 383.9 µs | −439 µs (−53.4%) | **2.14×** |
| anchor_25k | 24960 | 848.3 µs | 210.0 µs | −638 µs (−75.2%) | **4.04×** |

**Static barriers:** main `pipe_barrier(PIPE_ALL)` count = **7** → PR = **6** (PublishFence consolidation).

**Notes / caveats**
- An earlier quick run on devices 6,7 with W=1/M=3 showed noisy mid_24k scatter; W=3/M=8 on 4,5 is the number we trust.
- mid_24k PR measure set still has a long tail (min ~214 µs, max ~5289 µs across 8 samples); p50 is directional and favorable, but not a tight CI gate.
- Correctness smoke inside the harness: `recv_out == send_counts` after each warmup compile path.

---

## Files touched (high level)

- `python/pypto/runtime/builtins/collectives/all_to_all_v/templates/kernel.cpp.in` — B1 + S2
- `src/ir/transforms/insert_comm_fence_pass.cpp` — B2 + S1
- `src/ir/op/distributed/collective.cpp`, host/composite lowers — `signal S >= 2`
- `tests/ut/ir/transforms/test_insert_comm_fence.py`, ST signal buffer shapes, pass-51 docs (en/zh)

---

Thanks again — let us know if you'd like the count lane narrowed (e.g. ready `AtomicAdd(1)` + retry local `recv_counts` load after barrier) or if the current `AtomicAdd(rows+1)` carry is acceptable for K1.

YunjiQin (commented):
## Follow-up on the B1 / B2 changes

Thanks for the fast turnaround. Two items.

### 1. `recv_counts` must be the channel the counts are published on

My earlier review ended B1 with *"One option: an extra `signal` column used as a per-peer counts-published counter"*. **Please disregard that line** — I should not have put a suggested mechanism in a review, and it does not hold against the operand's semantics.

`recv_counts` is not a buffer the kernel happens to fill; it is *defined* as the window a peer publishes its count into. The deducer says so directly:

```cpp
// src/ir/op/distributed/collective.cpp
// recv_counts: window where each peer publishes how many rows it sent to me
// (MPI_Alltoallv recvcounts).
```

and the public docstring says the same, naming the publish mechanism:

> window-bound INT32 `recv_counts` [NR, 1] that **receives** per-source valid-row counts after the barrier (**published via `pld.system.notify`** as `clamp(send_counts[dest], 0, MAX_RECV)`)

Under the current change the count is published on a `signal` lane and `recv_counts` is filled locally afterwards from that lane. That inverts the operand's meaning: `recv_counts` becomes a local copy of a value carried elsewhere, and the "each peer publishes into it" contract no longer holds. `signal` is the barrier operand; it does not carry payload metadata.

So the count publish has to go back onto `recv_counts`, with `signal` back at `[NR, 1]`. Removing `CommRemotePtr(send_counts)` was correct and should stay — that part of B1 is settled; it is the replacement channel that has to change.

### 2. Phase B should match the read target by type, not by parameter identity

`insert_comm_fence_pass.cpp` — the header already records the gap:

> *"(matched by type, covering rebound SSA aliases only when the param var itself is read — see helper notes)"*

`ExprReadsDistTensorParam` resolves the read target and tests it against the parameter set:

```cpp
if (IsOp(call, "tensor.read") && !call->args_.empty()) {
  if (auto v = AsVarLike(call->args_[0]); v && dist_params.count(v.get()) > 0) {
    return true;
  }
}
```

A window param that is stored into before being scalar-read is rebound to a new `Var` (`w = pl.store(chunk, idx, w)`), which keeps `DistributedTensorType` but is no longer the parameter `Var` — so the read is missed and no prologue is inserted.

Match on the type of the read target instead:

```cpp
if (IsOp(call, "tensor.read") && !call->args_.empty() &&
    As<DistributedTensorType>(call->args_[0]->GetType())) {
  return true;
}
```

This covers rebound aliases by construction and lets `CollectDistTensorParams` and the `dist_params` plumbing go away entirely. Worth a UT for the store-then-scalar-read shape.

georgebisbas:
@YunjiQin Thanks for the follow-up — both items are addressed on the current head (rebased on latest `main`, single squashed commit).

### 1. Counts publish on `recv_counts` again; `signal` is barrier-only `[NR, 1]`

Agreed on the operand contract: authoritative counts belong on `recv_counts`, not a `signal` metadata lane. `CommRemotePtr(send_counts)` stays removed.

HOST/L2 visibility still needs the AtomicAdd/TWAIT family (plain Set + barrier was not enough on P=4; Set(rows+1)+TWAIT hung). So Phase 2 now publishes **on the `recv_counts` window**:

- `Set(0)` then `AtomicAdd(rows+1)` into peer `recv_counts[my_rank]` (clears prior-call residue; `rows==0` still wakes `Wait(GE 1)`)
- completion `ExchangeBarrier` on `signal` column 0 only
- peer: `TWAIT` → `dcci` → load → `AtomicAdd(-1)` leaving `rows` in the cell for consumers

InCore composite path unchanged (notify Set into `recv_counts`). Public/builtin deducers require `signal.shape[1] == 1` again.

### 2. Phase B matches by `DistributedTensorType`

`NeedsConsumePrologue` now treats any `tensor.read` whose first arg has exact-kind `DistributedTensorType` as a hit (no param-identity set). That covers SSA rebinds after `tile.store`. Added `test_incore_dist_scalar_read_after_store_rebind_gets_prologue`. Docs (en/zh pass-51 + index) updated.

Happy to iterate if you want a different clear/wake encoding on the `recv_counts` cell.

YunjiQin (commented):
### Counts: revert to the `main` shape; take the visibility question to pto-isa

Phase 2b should not exist, and nothing should run after the completion barrier. The local slot needs no special handling — publishing every slot with one `TNOTIFY(Set, rows)` in the push loop, self included, is what `main` already does, and it documents why that is enough:

> *a local self-notify (TNOTIFY on a local address is the same st_atomic the remote path uses)*

So the kernel's count path is:

```
for dest in 0..nranks:        # self included
    TNOTIFY(peer dest's recv_counts[my_rank], rows, Set)
    TPUT payload
PublishFence()
completion ExchangeBarrier()
# end of kernel
```

The `Set(0)` + `AtomicAdd(rows+1)` + `TWAIT(GE 1)` + `AtomicAdd(-1)` encoding, the `self_rows` register capture, and the post-barrier fill loop are all scaffolding around one unexplained observation: that a remote `NotifyOp::Set` on `recv_counts` is not visible to the peer after `PublishFence()` and the barrier.

That observation needs to go to pto-isa. If a remote `Set` released by a GM fence before a barrier notify is not visible to the peer after that barrier, either the marker contract is different from what this pass and this kernel assume, or there is a defect. Encoding around it in one collective's kernel leaves the same hazard in every other collective that publishes metadata this way, and leaves the contract unresolved.

#### First, please confirm the observation still holds

It may not survive the Phase B you just corrected, in which case there is nothing to escalate.

`consume_step` contains no `pld.system.wait`, so before Phase B existed, Phase A inserted **nothing** into it — its consume-side rule only fires after a wait. The count is then read as

```python
n_rows_i32 = pl.read(recv_counts, [src, 0])   # tensor.read -> cached pto.load_scalar
```

A stale cache line there reads back whatever the window held at init — `0`. That is exactly the reported signature, on self and peer slots alike, and it is what the entry invalidate now added by Phase B exists to prevent.

Lining the recorded verification up against the code it ran on:

| Verification (PR body) | Date | Code under test |
| --- | --- | --- |
| NPU P=2 STs green | — | "after Phase B **+ post-barrier local `recv_counts` fill**" |
| NPU P=4 gate green | 2026-09-14 | `ee47ddc9` — Phase B **v1 (orchestration scan)** + the `send_counts` pull |
| current head (`recv_counts` encoding + type-matched Phase B) | 2026-09-15 | no NPU run recorded |

The P=4 note in the kernel reads *"Ready-only AtomicAdd(1) + local recv_counts GM load still saw 0 on P=4"* — a ready-flag plus a local GM load, running on the old Phase B. So the combination that matters, **the corrected consumer invalidate together with `main`'s single `TNOTIFY(Set, rows)`**, has not been run at any point; each round added scaffolding on top of the previous one without isolating the cause.

Please run exactly that — count path reverted to the `main` shape above, current Phase B kept — and report P=2 and P=4:

- Green: the original symptom was the missing consumer invalidate. Phase 2a and Phase 2b both go away, and no pto-isa question arises.
- Still red: that run is the minimal reproducer to take to pto-isa — a single remote `Set`, released by a GM fence before the barrier notify, not visible to the peer after that barrier.

Either way, the count path reverts meanwhile.

<sub>(Reposted in English — the earlier Chinese body of this review was a mistake on my side.)</sub>

georgebisbas:
### Re: counts revert + isolation (review 5210528869)

Applied exactly the count path you asked for on tip [`aeb7044d`](https://github.com/georgebisbas/pypto/commit/aeb7044d6f4ed76419425d4a302350a98f295cdf) (pushed to `feat/all-to-all-v-flat-tput`):

```
for dest in 0..nranks:   # self included
    TNOTIFY(peer dest's recv_counts[my_rank], rows, Set)
    TPUT payload
PublishFence()
completion ExchangeBarrier()
# end of kernel
```

- Phase 2a / 2b encoding, `self_rows`, and any post-barrier fill are gone.
- Type-matched Phase B (consumer invalidate) kept as-is.
- No `AtomicAdd` count encoding restored.

**Isolation (main Set + current Phase B):** L2 + HOST `all_to_all_v` ST, P=2 and P=4, uniform + skew matrix — currently re-running on-device against this tip (prior onboard gate used an installed template that still matched `origin/main`; that is corrected now). First results + a 10× sweep will be posted as a follow-up; if green there is no pto-isa question, if red that run is the minimal reproducer.

Perf vs `origin/main` (P=2 / P=4, 10 measure rounds, swimlane AIV) is in flight on the same tip and will be reported separately.

georgebisbas:
### Isolation update (Yunji review 5210528869) — L2 red

On tip [`aeb7044d`](https://github.com/georgebisbas/pypto/commit/aeb7044d6f4ed76419425d4a302350a98f295cdf) (count path = main `TNOTIFY(Set)` + `PublishFence` + completion barrier; Phase 2a/2b gone; Phase B kept), onboard distributed ST reports:

**L2 `pld.tensor.all_to_all_v` — red**
- `test_l2_all_to_all_v[2]`: `recv_counts=0` (expected 2) at `rank=0 src=0` (self slot)
- `test_l2_all_to_all_v_skewed_counts[2-zero_and_full]`: `recv_counts=0` (expected 4) at `rank=1 src=1` (self slot)

**HOST `pld.tensor.all_to_all_v` — green** for the P=2 matrix in the same job (uniform + skew).

So this is Yunji’s “still red” isolation outcome for the **L2 rail**, not a regression to the deleted AtomicAdd encoding. Per that review: do **not** put the count encoding back; this run is the minimal L2 reproducer (remote/local `Set`, fence, barrier, consumer `tensor.read`) if we escalate visibility to pto-isa.

Caveat worth calling out before a blanket pto-isa escalation: HOST and L2 render the same builtin kernel source, but only L2 failed here. Digging whether Phase B’s entry `cacheinvalid` is present on the outlined L2 `consume_step` AIV in this build, and why HOST’s consume path does not show the same `0`, before treating it as a pure marker-contract defect.

Earlier commit message / first reply overstated “NPU green” — that was against an installed template that still matched `origin/main`. Retracted.

georgebisbas:
### Isolation update (rebased on `origin/main`) — Set + fence + barrier + Phase B still red

Rebased this branch onto latest [`origin/main`](https://github.com/hw-native-sys/pypto/commits/main) (`fb107da9`) and force-pushed tip [`f1900007`](https://github.com/georgebisbas/pypto/commit/f1900007) (runtime pin = main’s simpler `@22385d2b`, default packed window carve — no cache-line-align experiment, no Phase C, no AtomicAdd count encoding).

**What we ran (Yunji isolation shape + Phase B):**

```
# publish (builtin)
for dest: TNOTIFY(Set, rows) → peer recv_counts[my_rank]   # self included
PublishFence()   # pipe_barrier + dsb(DSB_DDR)
ExchangeBarrier()

# consume (InCore)
Phase B entry system.cacheinvalid()   # emitted as dcci(ENTIRE_DATA_CACHE)
pl.read(recv_counts, [src, 0])
```

Confirmed in emitted AIV `consume_step.cpp`: entry `dcci(ENTIRE_DATA_CACHE)` before the count load. Phase B UT: 35 passed after rebase.

**Onboard L2+HOST a2av ST** (`--device=4,5,6,7`): **13/20 pass, 7 fail** — still intermittent `recv_counts=0` (self and peer slots). Failures include L2 P=4 uniform and several L2/HOST skew cases.

So on this NPU, **Set + fence + barrier + consumer cacheinvalid + plain read is not sufficient**. Symptom matches the earlier isolation red; the only protocol we’ve previously seen green here was the AtomicAdd/TWAIT encoding on `recv_counts` (which this isolation intentionally dropped).

Happy to discuss next step: keep this kernel as the pto-isa repro, or restore the AtomicAdd/TWAIT count path as an a2a3 workaround with eyes open.

georgebisbas:
# PR #2738 - Yunji Spec Test Results

## Implementation

Per [review 5210528869](https://github.com/hw-native-sys/pypto/pull/2738#pullrequestreview-5210528869), implemented:

```cpp
// Push loop: for each dest, publish count via Set
TNOTIFY(remote_recv_slot, rows, NotifyOp::Set);
TPUT(...);  // payload

// After loop:
PublishFence();       // pipe_barrier(PIPE_ALL) + dsb(DSB_DDR)
ExchangeBarrier();    // AtomicAdd(+1) → TWAIT(GE 1) → AtomicAdd(-1) on signal
```

Consumer side (separate task) does `dcci` + read `recv_counts`.

## Test Results

| Rail | Test | Iterations | Passed | Failed | Failure Rate |
|------|------|------------|--------|--------|--------------|
| L2 | `test_l2_all_to_all_v[2]` | 20 | 19 | 1 | **5%** |
| HOST | `test_host_tensor_all_to_all_v[2]` | 5 | 4 | 1 | **20%** |

**Failure symptom**: `AssertionError: P=2 rank=0 src=0: recv_counts=0 != expected 2`

## Log

```
=== Testing Yunji spec: notify(Set) + fence(dsb) + barrier ===
Run 1-11: 1 passed
Run 12: FAILED - AssertionError: P=2 rank=0 src=0: recv_counts=0 != expected 2
Run 13-20: 1 passed
```

## Conclusion

The pattern `notify(Set) + fence(dsb) + barrier(notify+wait)` followed by `cache_invalid + read` does **NOT** reliably make the Set value visible. The failure is intermittent (~5-20%).

**Question for pto-isa**: Is `NotifyOp::Set` expected to be visible after this sequence? If not, what is the correct synchronization?

## Working Workaround

Using `AtomicAdd`/`TWAIT` on the **same** address (instead of Set + barrier on different address) works reliably:

```cpp
// Clear + encode on signal column
TNOTIFY(signal_slot, 0, NotifyOp::Set);
TNOTIFY(signal_slot, rows+1, NotifyOp::AtomicAdd);
PublishFence();
ExchangeBarrier();

// Consumer:
TWAIT(signal_slot, 1, WaitCmp::GE);  // ← This provides visibility
dcci + read → works
```

georgebisbas:
### ✅ hw-native-sys/pto-isa#323 helps to make this PR pass — no further pypto-side change needed

@YunjiQin — the L2/HOST redness on this branch comes from `pto::comm::TNOTIFY(..., NotifyOp::Set)` on a2a3, not from the K1 kernel rewrite or `InsertCommFence`. With **pto-isa #323** pinned, this PR goes green **as-is**.

**A/B — identical pypto head (`ea4a3f18`), identical command, only the pto-isa pin differs:**

| pto-isa pin | Result |
| --- | --- |
| `3b4faf67` (repo pin, store-based `Set`) | **8 failed, 26 passed** (582 s) |
| `bedec54c` (pto-isa #323) | **34 passed, 0 failed** (566–583 s) |

**What the ISA bug was.** a2a3's `NotifyOp::Set` was `dcci` + plain scalar store + `dcci`, i.e. it write-backs the **whole 64 B cache line**. A rank publishing its own `recv_counts[my_rank]` therefore clobbered the neighbouring ranks' counts packed into the same line (false sharing between packed int32 slots). #323 makes `Set` **word-published** — an atomic-sum delta of `(value - current)`, because `st_atomic` only accepts `ATOMIC_SUM` on this toolchain — so adjacent words are no longer disturbed.

Old-arm failure signature (7 of 8 failures were at P=4 — more writers per cache line):

```
AssertionError: P=2 case=one_and_full rank=0 src=0: recv_counts=0 != clamped(1) = 1
AssertionError: P=4 case=over_capacity rank=3 src=1: recv_counts=0 != clamped(7) = 4
```

The InCore/Intrinsic rail is green in **both** arms, which is what points at the shared-memory `Set` path rather than the TPUT rewrite.

**Repro** (8× Ascend 910B2, physical NPUs 4–7, CANN 9.0.0):

```bash
export ASCEND_RT_VISIBLE_DEVICES=4,5,6,7 PYPTO_CACHE=0
python -m pytest tests/st/distributed/collectives/test_l2_tensor_all_to_all_v.py \
                 tests/st/distributed/test_l3_host_tensor_all_to_all_v.py \
                 tests/st/distributed/collectives/test_l3_tensor_all_to_all_v_intrinsic.py \
                 --device 0,1,2,3 -v --forked
# pin bedec54c -> 34 passed;  pin 3b4faf67 -> 8 failed, 26 passed
```

Only ranks 2 and 4 were exercised (every case is parametrized `n_ranks=[2,4]`).

Re-verified after the review-fix commit on #323 (`10cde77e`, docs/ST-only — `git diff bedec54c 10cde77e -- include/` is 0 lines, so the ISA code is unchanged): still **34 passed, 0 failed**.

**Net:** #323 is sufficient — this PR needs no extra change on the pypto side, it just needs #323 to land (and the runtime ISA pin to move to a commit containing it).

georgebisbas:
@YunjiQin , Closing in favour of #2828 as discussed

PR #2740, opened 2026-09-11
feat(codegen): emit the declared GM cache policy as pto.tload's cache_policy

### LINKED ISSUE #2680: [Feature] Emit the declared GM cache policy as PTOAS's tload cache_policy attribute
## Background

In pypto-lib, the DeepSeek-V4 Flash MTP routed-expert kernel
(`models/deepseek_v4_flash_mtp/expert_routed.py`) wants its INT8 grouped-matmul
weight operands read **around L2**: the weight tile of each `(expert, row-tile)`
work item is streamed once and never revisited, so every L2 allocation it makes
is a guaranteed dead allocation that also evicts the activations which *do* get
reused. Measured on device by hand-editing the generated C++, the L2-bypass half
alone is worth **-7.9% on the routed-expert kernel / -2.8% of a whole decode
step**, and it is the larger of the two effects tracked in
hw-native-sys/pypto-lib#1039.

pypto already has the entire front end for this — `pl.set_cache_policy(t, pl.CachePolicy.BYPASS)`
and `pl.load(..., cache=...)`, shipped in #2540 — and the declaration reaches
codegen intact. `MakeTileLoadCodegenPTO` then **drops it**: it emits a warning
and an ordinary cached `pto.tload`. That was correct when it was written, and
the code says so:

```cpp
// src/backend/common/pto_ops_memory.cpp:113
// TEMPORARY (pypto #2534): PTOAS has no L2-bypass path yet
// (https://github.com/hw-native-sys/PTOAS/issues/1356), so a BYPASS request is
// carried through the IR but compiles as an ordinary cached access. When that
// issue closes, this warn is REPLACED in place by
// `GetOrCreateTensorView(tensor, policy)` against an addptr-rooted view — the
// declaration already reaches here, so nothing upstream changes.
```

**PTOAS 0.61 now has that path.** This issue asks for the in-place replacement
the comment already anticipates — with one simplification: 0.61 takes an
attribute on `pto.tload`, so no addptr-rooted view is needed.

Reproduce with:

    python repro_cache_policy.py

Reproduction environment:

| Component | Version |
|---|---|
| pypto-lib | `b445fe7e` (branch: `perf/optimize-dspark-moe-balanced-routing`) |
| pypto | `8ffd67fe` (branch: `main`) |
| simpler | `4e4d3a4a` (detached) |
| ptoas | `0.61` |
| pto-isa | `a8040450` (detached) |
| CANN | `9.0.0` |

Version mismatch (off pinned): ptoas expected `0.60` (`toolchain/versions.env`),
actual `0.61`. The divergence is deliberate — 0.61 is the first release that
implements the PTOAS side, so the pin needs to move too (see Proposed API /
Behavior).

Diagnosis: **pypto** — the assembler feature exists and the DSL declaration
exists; only the codegen hop between them is missing.

### Summary

`CachePolicy.BYPASS` is accepted by the DSL, survives every pass, and is then
discarded in codegen: `pto.tload` is emitted without a `cache_policy` attribute
and a `[CacheBypassUnsupported]` warning is logged instead. Generated code is
byte-identical with and without the declaration, so the feature is currently a
no-op end to end.

PTOAS 0.61 implements `cache_policy` on `pto.tload`:

```python
>>> from ptoas.mlir.dialects import pto
>>> list(pto.LoadCachePolicy.__members__)
['Default', 'L2Bypass']
>>> pto.LoadCachePolicyAttr.get(pto.LoadCachePolicy.L2Bypass, ctx)
#pto.load_cache_policy<l2_bypass>
```

Taking pypto's **own generated** `.pto` for the repro below, adding that
attribute to the weight `pto.tload` by hand, and re-running ptoas 0.61
(`--enable-insert-sync --pto-level=level3`) changes exactly one line of the
emitted CCE:

```diff
-  TLOAD(v45, v50);
+  TLOAD<pto::TLoadL2Hint::NotAllocKeep>(v45, v50);
```

That is a native pto-isa L2 hint — cleaner than the architecture-specific
`ptr + 0x80000000000ULL` address alias the original request assumed, which is
what pypto-lib had to hand-edit into the generated C++ to measure the numbers
above.

### Motivation / Use Case

Any GEMM operand that is streamed once and never revisited: the weight tile of a
grouped/batched matmul, any operand whose working set exceeds L2, one-shot
activation reloads. In the DeepSeek-V4 Flash decode MoE this is the dominant
term — measured per-tile-count on one a2a3 card, 100 rounds after 20 warmup,
`effective_us` mean, every point carrying its own baseline repeat (drift
0.03-0.57%):

| tiles | % of decode layers | L2 bypass's own share |
|---|---|---|
| 14 | 12.0% | -3.8 pp |
| 16 | 15.7% | -5.3 pp |
| 18 | 8.5% | -6.2 pp |
| 20 | 3.0% | -9.7 pp |
| 27 | 1.8% | -11.3 pp |

The effect **grows** with case size, because what it protects is the activation
working set the streamed weights would otherwise evict.

### Proposed API / Behavior

No DSL change — the front end is already there. In `MakeTileLoadCodegenPTO`,
replace the warn with an attribute on the emitted `pto.tload`:

```
pto.tload ins(%w_pview : ...) outs(%wt : ...) {cache_policy = #pto.load_cache_policy<l2_bypass>}
```

`CachePolicy.DEFAULT` keeps emitting no attribute, so nothing changes for code
that does not declare a policy. The MX `layout` attribute and `cache_policy` can
co-occur, so the emit site needs to join a list rather than print one attribute.

Requires PTOAS >= 0.61 (older releases reject the attribute), so it also needs
`PTOAS_VERSION` in `toolchain/versions.env` moved from `v0.60` to `v0.61` — or
the emit gated on the configured assembler version.

Reproduction case — `repro_cache_policy.py` (not committed; save anywhere and run):

```python
"""CachePolicy.BYPASS is dropped in codegen: no cache_policy on pto.tload.

    python repro_cache_policy.py
"""
import pypto.language as pl
from pypto.runtime import RunConfig

M = N = K = 256


@pl.jit
def mm_bypass(
    a: pl.Tensor[[M, K], pl.FP32],
    b: pl.Tensor[[K, N], pl.FP32],
    c: pl.Out[pl.Tensor[[M, N], pl.FP32]],
):
    with pl.at(level=pl.Level.CORE_GROUP, name_hint="mm"):
        pl.set_cache_policy(b, pl.CachePolicy.BYPASS)
        ta = a[:, :]
        tb = b[:, :]
        c[:, :] = pl.matmul(ta, tb)
    return c


if __name__ == "__main__":
    compiled = mm_bypass.compile(config=RunConfig(platform="a2a3"))
    print("compiled:", compiled)
```

Observed output (compile succeeds; the declaration is dropped):

```text
[warning] [CacheBypassUnsupported] tensor 'b' requests CachePolicy.BYPASS, but PTOAS
has no L2-bypass path yet (https://github.com/hw-native-sys/PTOAS/issues/1356);
compiling as an ordinary cached access at <repro-dir>/repro_cache_policy.py:20:9
```

Compiling the same kernel with and without the `pl.set_cache_policy` line and
diffing the two work dirs: `kernels/aic/mm.cpp` and `ptoas/mm.cpp` are byte-identical,
and `ptoas/mm.pto` differs only in `loc(...)` source line numbers.

### Additional Context

- Front end / docs: #2540, `docs/en/dev/language/05-cache-policy.md` (its
  "Current status" section is what this issue closes).
- PTOAS side: hw-native-sys/PTOAS#1356 — implemented in 0.61, though the issue
  is still open there.
- Related: hw-native-sys/pypto-lib#1039 (umbrella, carries the device measurements), #2534.

luohuan19:
Closes hw-native-sys/pypto#2680.

> **Stacked on hw-native-sys/pypto#2686 — merge that first.** This branch is cut
> from its head, because the feature needs the PTOAS v0.61 pin that PR moves. The
> base branch lives in the fork, and a PR here can only be based on a branch of
> this repository, so the commit list below also carries #2686's eight commits
> until it merges. **The only commit belonging to this PR is the last one**,
> `feat(codegen): emit the declared GM cache policy as pto.tload's cache_policy`.

`CachePolicy.BYPASS` was accepted by the DSL, survived every pass, and was then
dropped in codegen: `MakeTileLoadCodegenPTO` logged a `[CacheBypassUnsupported]`
warning and emitted an ordinary cached `pto.tload`. Generated code was
byte-identical with and without the declaration, so the feature was a no-op end
to end. That was correct while PTOAS had no bypass path, and the code said so —
naming, in a comment, the replacement this PR performs.

PTOAS v0.61 has that path, with one simplification over what the comment
anticipated: 0.61 takes an attribute on `pto.tload`, so no addptr-rooted view is
needed.

```mlir
pto.tload ins(%b__ssa_v0_pview : !pto.partition_tensor_view<256x256xf32>)
          outs(%b__ssa_v0_mat  : !pto.tile_buf<loc=mat, ...>)
          {cache_policy = #pto.load_cache_policy<l2_bypass>}
```

The assembler lowers that to pto-isa's own L2 hint, which is the whole difference
in the emitted CCE:

```diff
-  TLOAD(v45, v50);
+  TLOAD<pto::TLoadL2Hint::NotAllocKeep>(v45, v50);
```

## Three properties of the emit

Each is asserted by a test, and each is a way this could have gone wrong.

| Property | Why it matters |
| --- | --- |
| `CachePolicy.DEFAULT` emits **nothing** | A kernel that declares no policy keeps the PTO form it had before this existed. Stripping the attribute from the declared kernel's MLIR reproduces the undeclared kernel's, line for line — so the attribute is the *only* difference, not a codegen mode |
| The attribute is emitted **per load**, not per tensor | It is a property of the instruction. A hint on only the first of two loads would leave the second one allocating in L2. The superseded diagnostic was deliberately once-per-tensor — the opposite granularity — so its de-duplication state (`NoteCacheBypassWarned` / `cache_bypass_warned`) is **deleted**, not rewired |
| It joins the MX `layout` in **one** dict, after it | PTOAS takes all present attributes in a single dict. This follows the `attrs` vector `MakeTileStoreCodegenPTO` already uses in the same file, and keeping `layout` first leaves an MX load that declares no policy byte-identical |

## The per-access override is now observable

An explicit `cache=CachePolicy.DEFAULT` inside a bypassing scope emits no
attribute. The precedence was always documented, but while BYPASS was a no-op it
could only be checked on the IR
(`tests/ut/ir/transforms/test_convert_tensor_to_tile_ops.py`); there is now a
codegen-level test for it.

## No version gate

`toolchain/versions.env` pins PTOAS v0.61 (in #2686), so the emit is
unconditional. Nothing in the tree reads the pinned assembler version at compile
time, and adding that plumbing to make this land independently of #2686 would be
new machinery with no precedent — hence the stack.

## Testing

- `tests/ut/codegen` + `tests/ut/ir/transforms`: **4749 passed, 2 skipped**, on
  the rebased tree (rebuilt after rebasing onto #2686's current head).
- **`tests/ut/codegen/test_cache_policy_codegen.py` is rewritten, not adjusted.**
  Its previous assertions — "the generated MLIR is byte-identical with and
  without the declaration" and "the declaration warns once per tensor" — pinned
  exactly the behaviour this change removes. Five of its six tests are inverted
  or deleted on that basis; the surviving one now asserts the warning is *never*
  emitted again. Nine tests in total.
- **ptoas v0.61 run by hand on pypto's own emit.** Every UT runs with
  `skip_ptoas=True`, so no test in the tree would catch the assembler rejecting
  the attribute. Both paths were therefore assembled manually against the pinned
  release: the a2/a3 kernel above (accepted; only the declared operand's `TLOAD`
  carries the hint) and an a5 MX scale load, where `layout` and `cache_policy`
  share one dict (accepted; same hint). This is what the "one dict" property
  above rests on — the issue asserted co-occurrence without evidence.
- ruff check/format, pyright 1.1.407, clang-format 21.1.0, markdownlint, and the
  repo lint scripts (`check_docs_en_zh_parity`, `check_english_only`,
  `check_headers`, `check_emitter_check_classification`,
  `check_docs_symbol_coverage`): clean.

### Not covered locally

- **No hardware run.** The device measurements that motivate the request — the
  L2-bypass half is worth -7.9% on the DeepSeek-V4 Flash routed-expert kernel /
  -2.8% of a decode step — were taken in pypto-lib by hand-editing generated C++,
  and are tracked in hw-native-sys/pypto-lib#1039. This PR makes the declaration
  produce that code; it does not re-measure it.
- **No ST or example coverage of cache policy exists** (before or after this PR),
  which is why the assembler check above is manual. A future ST case would be the
  only thing that catches an assembler-side regression in CI.

## Docs

`docs/{en,zh}/dev/language/05-cache-policy.md` lose their "Current status"
no-op admonition for a **What codegen emits** section carrying the MLIR and CCE
above plus the three properties; `docs/{en,zh}/user/performance/05-memory.md`
state the v0.61 requirement instead of the warning. The `pl.set_cache_policy` and
`pl.load(cache=...)` docstrings follow.


luohuan19:
Superseded: reopening with only this change's own commit on top of main, so the diff is reviewable without #2686's eight commits in the list. CI here is expected to be re-run once #2686 lands the PTOAS v0.61 pin.

PR #2756, opened 2026-09-14
fix(ir): propagate rewritten tile layouts through aliases and loop carries

### LINKED ISSUE #2753: [Pass Bug] InferTileMemorySpace leaves stale TileView metadata on aliases and loop carries
### Pass Name

InferTileMemorySpace

### Description

On Ascend 950, inserting an Acc-to-Vec move changes a vector producer's result to a flat TileView, but an SSA alias retains its original boxed TileView. Loop-carried values can also retain stale layout metadata after their initializer is rewritten. This blocks native MXFP8 compensated accumulation used by Prefill SWA.

### Git Commit ID

8ae7db081c24591d8e6f613fd9393953b7e077a5

### Before IR (Input)

Select BackendType.Ascend950, parse this program, and run passes.infer_tile_memory_space().

```python
import pypto.language as pl

@pl.program
class Before:
    @pl.function(type=pl.FunctionType.AIV)
    def main(self, x: pl.Tile[[32, 128], pl.FP32, pl.MemorySpace.Acc]):
        value = pl.mul(x, 1.0)
        alias = value
```

The alias must inherit the rewritten producer's flat Vec view.

```python
@pl.program
class Expected:
    @pl.function(type=pl.FunctionType.AIV)
    def main(self, x: pl.Tile[[32, 128], pl.FP32, pl.MemorySpace.Acc]):
        x_Vec = pl.move(
            x, target_memory=pl.MemorySpace.Vec,
            blayout=pl.TileLayout.row_major,
            slayout=pl.TileLayout.none_box,
        )
        value = pl.mul(x_Vec, 1.0)
        alias = value
```

### Actual IR or Error

The reduced program produces a flat Vec `value`, but its alias has:

```python
alias: pl.Tile[[32, 128], pl.FP32, pl.Mem.Vec, pl.TileView(blayout=pl.TileLayout.col_major, slayout=pl.TileLayout.row_major)] = value
```

The MXFP8 accumulation reproducer subsequently fails device C++ compilation at TSTORE because the boxed vector tile is incompatible with the ND output. Replacing the alias with multiplication by 1.0 bypasses the reduced case, but a full-width K=5120 loop still fails with stale loop-carried views.

### NPU Kind (if relevant)

Other: Ascend 950 (A5)

Linux (aarch64)

### Additional Context

Related: #1262 (memory-space asymmetry; distinct from this stale-layout defect).

Related: https://github.com/hw-native-sys/pypto-lib/issues/1217

Scope: propagate rewritten TileView metadata through aliases and loop carries while preserving memory-space, allocation, and valid-shape contracts. No public API or toolchain pin changes.

Validation plan: focused alias and control-flow regression tests, relevant existing unit tests, and the reduced and full-width MXFP8 compensated-accumulation reproducers on Ascend 950. Compiler repair alone does not establish full Prefill SWA numerical acceptance.

lwDavid:
## Summary

Fixes #2753. `InferTileMemorySpace` can rebuild an Acc-to-Vec producer with a flat layout while its aliases and loop carries retain the old boxed layout. This mismatch can survive into device code generation and fail ND stores. The pass now propagates the rewritten layout through these value bindings, without changing public APIs or toolchain pins.

## Changes

- Synchronize plain Var/IterArg aliases with the rewritten RHS, extending the existing call-result synchronization.
- Propagate initializer layout through ForStmt/WhileStmt carries and loop results, including zero-trip loops, while preserving carry-specific shape, dtype, allocation, and valid-shape metadata.
- Add regression and print/parse round-trip coverage for alias chains, nested loops, multiple carries, passthrough IterArgs, while loops, zero/one/many iterations, and explicit boxed Vec layouts.
- Update the existing English and Chinese pass descriptions. No operator scaffolding, generated artifacts, or diagnostic scripts are included.

## Verification

- `pip install . --no-build-isolation --no-deps`: passed; built and installed from the working compiler checkout.
- `python -m pytest tests/ut/language/test_unified_ops.py tests/ut/ir/transforms/test_infer_tile_memory_space.py -n "$PYPTO_TEST_JOBS" -q`: 313 passed.
- `python -m pytest tests/ut/ -n "$PYPTO_TEST_JOBS" -q`: 13,028 passed, 4 skipped, 1 xfailed.
- `pre-commit run --files src/ir/transforms/infer_tile_memory_space_pass.cpp tests/ut/ir/transforms/test_infer_tile_memory_space.py docs/en/dev/passes/20-infer_tile_memory_space.md docs/zh/dev/passes/20-infer_tile_memory_space.md`: passed, including clang-format, cpplint, Ruff, Pyright, and repository policy checks.
- Ascend 950 device checks, through the device allocator: original plain-alias MXFP8 compensated accumulation passed at M=32/K=512/N=128 (BF16 output), and M=32/K=5120/N=1280 (FP32 output). The latter achieved relative L2 error 1.93896e-7 and maximum absolute error 1.19209e-6 against an independent FP64 reference.

This resolves the compiler blocker found during Prefill SWA development, not the separate end-to-end quantization-boundary precision issue. Full system tests and full Prefill SWA acceptance are not claimed here.


lwDavid:
## Implementation and validation report

### Root cause and fix

The original pass synchronized assignment types only when the RHS was a Call. A producer rewritten after Acc-to-Vec argument conversion therefore acquired a flat TileView while `alias = producer` retained a boxed view. IterArg rewriting and loop results also retained stale views when their recorded memory space did not change.

The fix extends assignment synchronization to Var/IterArg aliases, inherits the rewritten initializer's physical layout before visiting a loop body, and propagates that carry layout to loop results. Carry-specific data metadata is preserved; explicit non-default layouts are not replaced with defaults. This adds a bounded traversal of each loop's result slots and retains O(N log N) pass complexity. No public interface, runtime, PTOAS, or ISA pin changes are needed.

### Regression evidence

Before the fix, the alias-chain and nested-for cases failed on the diagnosed compiler revision; the explicit boxed-Vec preservation control passed. With the fix, all six new cases and all 101 existing tests in the pass file pass. Tests include structural equality and print/parse round trips.

The complete unit suite passes: **13,028 passed, 4 skipped, 1 xfailed**. An intermediate editable-install-only symlink-import failure was eliminated by validating the normal install from the same checkout; no existing test expectation was changed.

All applicable pre-commit hooks pass, including the repository policy checks, clang-format, cpplint, Markdown lint, Ruff, and Pyright.

### Ascend 950 numerical checks

Both original compensated-accumulation forms were executed on A5 hardware through `task-submit`, with the plain `acc = total` alias intact:

| Shape (M, K, N) | Output | Result |
| --- | --- | --- |
| (32, 512, 128) | BF16 | Device compilation, execution, and existing golden validation passed |
| (32, 5120, 1280) | FP32 | Device compilation, 20-partial accumulation, and numerical validation passed |

The full-width case includes the two previously identified sensitive input rows. Against independently dequantized operands multiplied in FP64 and rounded to FP32, relative L2 error is **1.9389567e-7**, and maximum absolute error is **1.1920929e-6**. The checks established before this run were `rtol=1e-4`, `atol=1e-5`, and relative L2 <= 1e-5. These are engineering checks for this compiler reproducer, not an official full-model acceptance threshold.

Environment: Ascend 950/A5, CANN 9.2, PTOAS 0.61, runtime `39ce891d`, PTO ISA `5a4f74cb`; compiler base `58732a93`. The reference uses the existing official-rule MXFP8 quantization fixture; the compiler fix does not change the attention or quantization algorithms.

### Limits

BF16 midpoint-sensitive differences remain: the full-width output has 11 BF16-rounded differences from the FP64 reference across 40,960 elements. The compiler blocker is resolved, but this does not prove that compensated accumulation resolves the separate Prefill SWA output-error outliers. Full-model/checkpoint validation, full distributed Prefill acceptance, full system tests, and performance improvement are not claimed.

lyfne123 (approved):

Issue #2759, opened 2026-09-15
[Feature] Add explicit constexpr parameters to JIT functions


lyfne123:
### Summary

Add explicit compile-time parameters to `@pl.jit`, using `pl.constexpr` (proposed spelling), while keeping ordinary `pl.Scalar[...]` parameters as runtime values.

### Motivation / Use Case

PR #2758 makes ordinary scalar parameters runtime values. Module-level and closure constants still support specialization, but selecting multiple tile sizes or algorithm configurations requires global-state changes or kernel factories. An explicit parameter makes these configuration choices visible in the signature and easier to enumerate for tuning.

The goal is a clearer specialization interface, not stronger optimization than equivalent captured constants.

### Proposed API / Behavior

Illustrative syntax for an already-defined kernel:

```python
# Signature: kernel(x, out, scale: pl.Scalar[pl.FP32], BLOCK: pl.constexpr)
kernel(x, out, scale=1.0, BLOCK=128)
kernel(x, out, scale=2.0, BLOCK=128)  # Reuse the specialization.
kernel(x, out, scale=2.0, BLOCK=256)  # Select a different specialization.
```

Acceptance criteria:

- Resolve constexpr arguments during specialization, supporting static tile shapes, unroll extents, and compile-time branch selection.
- Include their identity in in-memory and persistent cache keys; changing only runtime Scalars must not split the cache.
- Remove constexpr parameters from the generated runtime ABI. Preserve specialization through nested JIT calls, including calls to one dependency with different constant values.
- Define consistent binding/default rules for direct calls, `compile()`, `lower()`, and `warmup()`. Prepared artifacts receive only runtime arguments.
- Reject missing or non-constant constexpr arguments with clear diagnostics. Document supported value types and their cache identity rules.
- Test cache reuse/separation, generated IR and ABI, nested calls, persistent restoration, and value-dependent execution. Update English and Chinese documentation with migration examples.

### Alternatives Considered

- Module-level constants: sufficient for fixed configuration, but switching configurations through mutable globals obscures the call interface.
- Closure factories: allow multiple configurations without global mutation, but require users and tuning code to manage separate JIT function instances.
- Restore implicit specialization of ordinary Scalars: would reintroduce value-dependent recompilation for runtime data.

Keep Python constants and closures supported; constexpr would complement them.

### Additional Context

Related: #2751 — establishes runtime Scalar defaults and asks to retain explicit specialization.

[...]

PR #2772, opened 2026-09-15
feat(jit): compile a dep once per constexpr value it is called with

lyfne123:
## Summary

Completes #2759. #2764 added `pl.constexpr` but emitted one generated function per dependency, so every call site of a dep had to agree on every compile-time value — the last acceptance criterion, *"including calls to one dependency with different constant values"*, was deliberately deferred there and is what this PR delivers.

Calling one helper at two tile sizes is the reason to reach for a compile-time parameter in the first place. It was refused with an error naming both values, leaving hand-duplicating the helper as the only way through.

## What changes

A generated function is identified by `(function, binding)` rather than by the function alone. Call sites that agree still collapse onto one, so a single-binding program emits exactly what it emitted before — down to the generated names. A dep reached with two different constants is emitted twice, and each call site is rewritten to its own:

```python
@pl.jit.incore
def helper(a, out, N: pl.constexpr):
    pl.store(pl.load(a, [0, 0], [N, N]), [0, 0], out)
    return out

@pl.jit
def entry(a, out):
    helper(a, out, 16)          # -> self.helper(a, out)        folds [16, 16]
    return helper(a, out, 32)   # -> self.helper__2(a, out)     folds [32, 32]
```

Splitting propagates. A dep that forwards its value onward splits everything downstream, and each copy calls its own: `entry -> mid(N) -> leaf(N)` at 8 and 24 emits `mid`/`mid__2` and `leaf`/`leaf__2`, with `mid__2` calling `leaf__2`.

## Three things had to stop keying on the function alone

| | Before | Now |
| --- | --- | --- |
| **Name allocation** | one name per function | one per `(function, binding)` — the `__2` suffixing that already separates two deps sharing a `__name__` does both jobs, so this stays one mechanism rather than two |
| **Call rewriting** | resolved by call name | resolved per call site; the JIT layer answers by `(call name, ordinal)` and the specializer resolves that onto its own nodes |
| **Call-site metadata** | read from the first call bearing the name | read from the call that produced each specialization |

Functions keep their former relative order in name allocation, and within a function variants are named in call-site source order, so the first site keeps the unsuffixed name.

The metadata one is a correctness requirement, not a nicety. Reading the first call site was sound while one binding meant one function; with two it folds the second binding's constant over the *first* site's tensors — a `[32, 32]` extent loaded from a `[16, 16]` parameter, with no diagnostic. There is a regression test for exactly that shape.

The specializer cannot be handed the JIT layer's call nodes, because it parses its own copy of the source. Both sides enumerate `ast.walk` matches sorted by position over the same dedented string, so the Nth call of a name means the same call in both.

## Cache identity

One record group per binding rather than per function, so a program calling a dep at 16 and 32 no longer collides with one calling it twice at 16 — different programs, and the second would otherwise be served the first's artifact.

## Not changed

Unbound and non-constant constexpr arguments are refused exactly as before, including the shadowed-global rule. Only the divergence error is gone, replaced by the specialization it was standing in for. Two tests that asserted that error now assert the specialization instead.

## Testing

- `tests/ut/jit/` — 764 passed, 2 skipped
- `tests/ut/` — 13338 passed, 10 skipped, 1 xfailed; the single failure is `test_symlinked_import_path_still_names_the_caller`, a pre-existing editable-install environment issue unrelated to this change
- `ruff check` / `ruff format --check` clean on the changed files; `pyright` 0 errors

New coverage: two call sites at different constants each getting a function; agreeing sites still sharing one; a diamond whose two callers disagree; a forwarded constant splitting a whole chain with each copy calling its own; each specialization reading its own call site's metadata; both bindings reaching the cache key; and variant-aware name allocation.

## Docs

`docs/{en,zh}/user/00-getting_started.md` — the "one generated function per dependency, so every call site must agree" rule is replaced by the per-binding rule and the propagation note.


lyfne123:
Fixed in `25d27c00b`: the call-site → specialization wiring is now part of the cache identity.

**Confirmed the reported case.** With three sites at `16`, `32` and `cfg.N`, flipping `cfg.N` between the two literals left the binding *set* at `{16, 32}` and the identity records byte-identical, while the generated program moved the third call from `self.helper` to `self.helper__2`:

```
cfg.N = 16   ->  ..., return self.helper(a, out_v2)
cfg.N = 32   ->  ..., return self.helper__2(a, out_v2)
generated code differs : True
records differ         : False
CACHE KEY differs      : False     <-- the bug
```

`cfg.N` is an attribute, so `_get_source_hash` could not see it either — it renders free names only. Both signals were blind at once, which is why this reached a silently wrong artifact rather than a miss.

`_constexpr_identity_records` now also emits, per variant, one record per call site: caller position and binding, call name and ordinal, callee position and binding. Same repro after the fix reports `CACHE KEY differs : True`.

**The wiring is emitted only when something is actually split.** With one binding per function each call name has exactly one possible target, already pinned by the call graph in the source hash, so emitting it there would invalidate every existing key and its artifacts to say nothing new. `test_an_unsplit_program_records_no_call_wiring` locks that in.

Regression tests, both in `tests/ut/jit/test_cache.py`:

- `test_retargeting_a_call_site_moves_the_key` — the reported config switch, asserting both that only the third call site moves and that the key moves with it.
- `test_an_unsplit_program_records_no_call_wiring` — the no-churn guarantee above.

`tests/ut/` 13341 passed; the one failure is the pre-existing editable-install `test_symlinked_import_path_still_names_the_caller`. ruff/pyright clean.

Hzfengsy (approved):

PR #2820, opened 2026-09-18
fix(codegen): cite the live ptoas bug behind the co-live slot rejection

Hzfengsy:
## Summary

Under `memory_planner=PTOAS`, codegen refuses a `pl.MemRef(slots=N)` whose loop body selects two of its slots. The error, comments and docs gave [PTOAS#1118](https://github.com/hw-native-sys/PTOAS/issues/1118) as the reason ("ptoas guards only the first `multi_tile_get`", ptoas 0.54). That bug was fixed in ptoas 0.56. The rejection is still needed, but for a different bug: [PTOAS#1519](https://github.com/hw-native-sys/PTOAS/issues/1519), which the pinned ptoas 0.61 still has. This PR corrects the stated reason everywhere. **The rejection and the pinned ptoas version are unchanged; only the error text differs.**

| Form of a two-slot loop body | ptoas behaviour | Upstream |
| --- | --- | --- |
| Two slots filled and read in the same iteration | ≤ 0.55: second load unguarded (measured wrong on device with 0.54). 0.56+: one static event guards the whole region; correct, but no per-slot overlap | PTOAS#1118, fixed in 0.56 (PR #1132) |
| Prefetch: slot 0 filled before the loop, then each iteration fills `(i+1) % 2` while reading `i % 2` | ≤ 0.62 primes and drains both slots' events as it would for a one-slot rotation, which is off by one here: wrong data for an even trip count, a device hang for an odd one | PTOAS#1519, fixed in 0.63 |

`CoLiveSlotCollector` only counts slot selections per loop body, so it cannot tell the two forms apart. Lifting the rejection needs the ptoas pin moved to 0.63 or later and a device run of both forms. That is a separate change.

Separately, the default PyPTO planner's "no region at level3" gate cited [PTOAS#1106](https://github.com/hw-native-sys/PTOAS/issues/1106) (the address fan-out is not folded). That issue is closed: given a constant `addr`, ptoas derives the same per-slot sync at level3 as at level2. The gate's actual reason is on PyPTO's side: codegen emits no `addr` for a region. Comments and docs now say so. Behavior is unchanged.

## Changes

- `src/codegen/pto/pto_codegen.cpp`:
  - The `CoLiveSlotCollector` comment now describes both forms.
  - The `PlanMultiBufferRegions` level3 comment is corrected.
  - The co-live error now reads "…ptoas before 0.63 mis-synchronizes one form of that — a slot filled an iteration ahead of its read — into wrong data or a device hang (hw-native-sys/PTOAS#1519)". The prefix the unit test matches is unchanged.
- `include/pypto/codegen/pto/pto_codegen.h`: the `PlanMultiBufferRegions` doc comment is updated for both corrections.
- `docs/{en,zh}/dev/codegen/00-pto_codegen.md`: "One slot per iteration" now has the table above, and the level3 note is corrected.
- `docs/{en,zh}/dev/language/00-python_syntax.md`, `docs/{en,zh}/dev/passes/30-lower_pipeline_to_slots.md`, `docs/{en,zh}/user/performance/04-incore.md`, `docs/{en,zh}/user/appendix/01-faq.md`: updated to match; the pass doc now links to the codegen section instead of restating the reason.
- `tests/ut/codegen/test_multi_buffer_codegen.py`, `tests/st/runtime/ops/test_memory_planner_ptoas.py`: docstrings and comments updated. No assertion changed.

## Verification

- `pytest tests/ut/codegen/test_multi_buffer_codegen.py tests/ut/ir/transforms/test_lower_pipeline_to_slots.py`: 49 passed. This was run on the rebased commit after an incremental rebuild.
- `pytest tests/st/runtime/ops/test_memory_planner_ptoas.py -k colive_slots_rejected`: 1 passed (the rejection still fires through `ir.compile`).
- `pre-commit run --from-ref upstream/main --to-ref HEAD`: all hooks passed (clang-format, cpplint, markdownlint, ruff, pyright, en/zh doc parity).
- ptoas 0.61, `--enable-insert-sync --pto-level=level2 --pto-arch=a3`, on PTOAS's `multi_tile_same_region_two_gets_sync.pto` (the same `lo`/`hi` + `tadd` shape as `CoLiveSlotsInLoop`): one static `PIPE_V→PIPE_MTE2` wait before the first `TLOAD` covers both loads.
- ptoas 0.61, a one-slot `i % 2` rotation compiled at level2 (no `addr`) and at level3 (`alloc_multi_tile addr = <const>`): the two sync sequences are identical, with per-slot dynamic events.
- A DSL kernel with a prologue load into slot 0 and a loop that loads slot `(i+1) % 2` while reading slot `i % 2` (bound with `pl.tile.create`) reaches codegen. It is refused only by `CoLiveSlotCollector`, now with the #1519 message. Writing the same prefetch with a loop carry fails earlier, in MemoryReuse's undecidable-slot carry check.
- No device run. The change touches only text, not codegen output.


Hzfengsy:
@codex review

lyfne123 (approved):

Issue #2831, opened 2026-09-20
[jit] Persistent cache always bypasses with wheel-installed ptoas: pip 26 console-script grammar rejected


ndleslx:
### Summary

`pypto.jit._toolchain._console_interpreter()` rejects the console script that current pip generates for the `ptoas` entry point. On hosts where ptoas is installed from the release wheel with a recent pip, `capture_toolchain` reports an unusable identity, so the persistent JIT cache silently bypasses on every request and never turns on.

### Environment

- pypto @ `04f1f510` (main), installed with `pip install -e ".[dev]"`
- ptoas 0.61 from the GitHub release wheel (`ptoas-0.61-cp312-cp312-manylinux_2_34_aarch64.whl`), installed with pip 26.1.2 into a conda env

### Observed

`$PREFIX/bin/ptoas` as generated by pip 26:

```python
#!/.../bin/python3.12
import sys
from ptoas._cli import main
if __name__ == '__main__':
    sys.argv[0] = sys.argv[0].removesuffix('.exe')
    sys.exit(main())
```

`_CONSOLE_BODIES` (`pypto/jit/_toolchain.py:296`) only accepts the two older templates (the `re.sub(r'(-script\.pyw|\.exe)?$', ...)` form and the `-script.pyw`/`.exe` slicing form), so the AST-equality check in `_console_interpreter` raises `Unsupported PTOAS console-script grammar` and every component of `capture_toolchain` fails with it:

```
pypto: Unsupported PTOAS console-script grammar: .../bin/ptoas;
runtime: Unsupported PTOAS console-script grammar: ...;
pto_isa: Unsupported PTOAS console-script grammar: ...;
ptoas: Unsupported PTOAS console-script grammar: ...;
device_toolchain: Unsupported PTOAS console-script grammar: ...
```

With `PYPTO_CACHE=1`, `pypto.cache_stats().bypasses` increments on every JIT call and `last_bypass_reason` shows the grammar rejection, so the persistent cache never engages. Rewriting the launcher to the `re.sub` template (functionally identical) makes `capture_toolchain` pass, which confirms the template whitelist is the only blocker.

### Impact

Any deployment installing ptoas from the release wheel with a current pip silently loses persistent JIT caching — compilation still works, the cache just never enables. It also blocks the pypto-serving integration that routes `--use-compile-cache` through this cache: its READY-reuse acceptance asserts zero bypasses and fails on such hosts.

### Suggestion

[...]

PR #2836, opened 2026-09-20
fix(jit): inventory an editable install's second tree instead of refusing it

### LINKED ISSUE #2834: [jit] Persistent cache always bypasses on editable installs: scikit-build-core extension split rejected as import redirect
### Summary

`pypto.jit._toolchain._package("pypto")` requires the whole `pypto` package tree (including `pypto_core`) to load from one directory. A scikit-build-core editable install deliberately splits it — Python sources load from the checkout while the built `pypto_core*.so` is a real file under `site-packages/pypto/` — so the toolchain inventory rejects it as an "external import redirect" and the persistent JIT cache bypasses on every call.

### Environment

- pypto @ `04f1f510` (main), installed with the documented dev workflow `pip install -e ".[dev]"` (README "Installation" / AGENTS.md) into a conda env
- scikit-build-core default editable mode (no `editable.mode` override; generates `_editable_skbc_pypto.pth` + finder)

### Observed

After an editable install, imports resolve from two different trees:

```python
>>> import pypto, pypto.pypto_core
>>> pypto.__file__
'/path/to/checkout/python/pypto/__init__.py'
>>> pypto.pypto_core.__file__
'/path/to/env/lib/python3.12/site-packages/pypto/pypto_core.cpython-312-aarch64-linux-gnu.so'
```

`_package()` (`pypto/jit/_toolchain.py:151`) takes the package root from `pypto/__init__.py` and then flags the submodule, so every component of `capture_toolchain` fails with:

```
pypto: Compiler package uses an external import redirect: pypto.pypto_core
runtime: Compiler package uses an external import redirect: pypto.pypto_core
pto_isa: Compiler package uses an external import redirect: pypto.pypto_core
ptoas: Compiler package uses an external import redirect: pypto.pypto_core
device_toolchain: Compiler package uses an external import redirect: pypto.pypto_core
```

With `PYPTO_CACHE=1`, `pypto.cache_stats().bypasses` increments on every JIT call and `last_bypass_reason` shows the redirect — the persistent cache never engages. This is the split scikit-build-core's default editable mode intentionally creates (sources mapped to the checkout, built extensions installed as real files), so any environment installed per the README hits it.

Reinstalling non-editable (`pip install .`) puts everything in one `site-packages/pypto` tree and `capture_toolchain` becomes usable again, which confirms the split is the only blocker.

### Impact

The persistent JIT cache (#2723) is unusable in every editable dev environment, i.e. for the workflow the README and AGENTS.md document. It also blocks downstream repos' cache acceptance tests when their dev stacks build pypto editable — pypto-serving's READY-reuse guard (asserting zero bypasses) cannot pass against an editable pypto.

Found while validating the pypto-serving JIT-cache integration, alongside the wheel-ptoas console-script rejection in #2831.

### Suggestion

Any of these would work:

1. Teach `_package` (or the pypto component) to recognize the scikit-build-core editable split — e.g. accept extension files under `site-packages/pypto` when the editable finder maps the package to a checkout, hashing both trees into one installation identity.
2. Support/advertise `editable.mode = "inplace"` (builds extensions into the source tree, one root) for developers who want the persistent cache.
3. At minimum, document that persistent caching requires a non-editable install, next to the `pip install -e` instructions.

lyfne123:
## Summary

`_package` required every already-imported submodule to load from the one
directory holding the package's `__init__.py`. A scikit-build-core editable
install deliberately splits that — Python sources map to the checkout, the
built extension stays a real file under `site-packages/<name>`:

```
pypto.__file__      = /path/to/checkout/python/pypto/__init__.py
pypto_core.__file__ = /env/lib/python3.10/site-packages/pypto/pypto_core...so
```

That is the layout the documented `pip install -e ".[dev]"` workflow produces,
so `_package` raised, the error propagated out of `_discover`, and **every**
component of `capture_toolchain` carried it — leaving the persistent cache
bypassing on every request, in every editable dev environment.

## What changed

Refusing the split never made the identity safer; it made the toolchain
unavailable, which switches the cache off silently. A submodule loading from
outside the first tree now adds **its** tree, and both are inventoried and
hashed — so the identity covers the extension that is actually loaded, rather
than declining to have an identity at all.

Two properties are kept deliberately:

- **A redirect is only accepted when it lands inside a directory named for the
  package.** That is the build system placing the package's own files, and
  every later import from it is covered because the whole directory is
  inventoried. A redirect anywhere else is still refused — nothing bounds what
  it would drag in.
- **Shared-library closures are resolved per tree.** In an editable install the
  checkout holds no extension at all; the `.so` whose `ldd` closure matters
  lives only in the second tree, so sweeping just the first would miss it.

## Residual gap

Only submodules already imported when `capture_toolchain` runs are visible in
the second tree. Inside the first tree that does not matter — the whole root is
inventoried recursively, so a later import is covered. Adding the *directory*
rather than the file narrows the gap to the same shape for the second tree, but
a tree that appears only after capture is still invisible. Worth knowing if a
component ever redirects lazily.

## Verification

- Observed on this host, which is installed exactly as the issue describes
  (`ScikitBuildRedirectingFinder` on `sys.meta_path`, sources in the checkout,
  a 407 MB `pypto_core...so` alone under `site-packages/pypto/`). The editable
  install here targets an older checkout, so the end-to-end failure is
  reproduced with a synthetic package of the same shape rather than with
  `pypto` itself.
- Before: `_package` → `ValueError: Compiler package uses an external import
  redirect`. After: both trees returned, and modifying a file in the second
  tree changes the component digest — so it is covered, not merely listed.
- A redirect to a file outside any directory named for the package is still
  refused.
- `pytest tests/ut/jit tests/ut/runtime tests/ut/backend -n 16`: 1812 passed,
  2 skipped.
- Four new tests. Mutation checks: reverting to the unconditional refusal,
  accepting any redirect, and sweeping only the first tree's `.so` files each
  fail a different one.
- Full `pre-commit` locally: exit 0.
- Not run: system or device tests. Nothing here reaches device code.

## Related

Third instance of one failure shape: a discovery step raises, all five
components come back unavailable, and the cache turns itself off with no
signal but `cache_stats().last_bypass_reason`. The others are #2831 (fixed by
#2833) and an unfiled one — `RunConfig()` defaults to `a2a3sim`, whose host
compiler probe fails when `g++` on PATH is ccache. A shared diagnostic for
"identity unavailable" seems worth having.

Closes #2834


lyfne123:
Folded into #2833 as its second commit, together with two further defects in the same class (argv[0] wrappers, and probes reading translated output) that were found while verifying it. Closing here so the four are reviewed as one change.

PR #2837, opened 2026-09-20
fix(toolchain): Update runtime and adapt codegen to PTOAS v0.64

Little-oil:
Update runtime to `6e383fc57c12a8bd4dce045289b8b4f5d55a15f1` and PTOAS to `v0.64`, including the matching CPython 3.10 wheel hashes and Python minimum version. The runtime also advances its PTO-ISA pin to `c0d7148e`.

PTOAS 0.64 removes `pto.load_scalar` and `pto.store_scalar`. Emit the replacement `pto.load` and `pto.store` operations for tensor scalar accesses and distributed CommContext reads, preserving the existing pointer, offset, and type syntax. Update codegen assertions and documentation, and add scalar load/store compilation coverage for FP32, INT32, UINT32, and INT64.

Validation:
- Pre-commit checks passed before commit; wheel hashes match GitHub release digests.
- Local build, targeted clang-tidy, and runtime-independent compilation tests are in progress. Pushed before local validation completes so CI can run in parallel.
- Previous CI checkout failures were TLS/early-EOF download errors and are not addressed here.


lyfne123:
@codex review

lyfne123 (approved):

PR #2841, opened 2026-09-20
fix(pass): keep an inline callee's rebinding off a pass-by-value arg

Hzfengsy:
## Summary

`InlineFunctions` applied its param-substitution seed at def-sites as well as use-sites for **every** parameter, which made every parameter pass-by-reference. A callee doing `n = n + 1`, called as `m = self.bump(k)`, spliced `k = k + 1` into the caller, so every later read of `k` saw the bumped value:

```python
@pl.function(type=pl.FunctionType.Inline)
def bump(self, n: pl.Scalar[pl.INDEX]) -> pl.Scalar[pl.INDEX]:
    n = n + 1
    return n

@pl.function
def main(self, k: pl.Scalar[pl.INDEX]) -> pl.Scalar[pl.INDEX]:
    m: pl.Scalar[pl.INDEX] = self.bump(k)
    return k + m                    # wants k + (k + 1) == 2k + 1
```

Before, the pass emitted `k = k + 1; m = k; s = k + m` — `2k + 2`. Silent wrong results, no diagnostic, and `ConvertToSSA` cements the dataflow (`s = k__ssa_v1 + m`), so nothing downstream can recover the original value.

The fix consults the callee's declared `param_directions_`, which the pass never read. A rebinding lands on the caller's `Var` only where aliasing is the contract:

| Parameter | Rebinding lands on |
| --------- | ------------------ |
| `pl.Out[...]` / `pl.InOut[...]` | the caller's `Var` — the author opted in explicitly |
| Any tensor / tile param | the caller's `Var` — see below |
| Everything else (scalar, `Array`, `Ptr`) | a fresh `<param>_inline<n>` `Var` |

**Direction alone cannot be the test.** `@pl.jit.inline` deliberately strips `pl.Out` / `pl.InOut` from shaped params (`specializer.py:2361`) because an inline callee's tensor params *are* aliases of the caller's handles — that is what lets `c[...] = v` (parsed as `c = pl.tensor.assemble(c, ...)`) write through. Keying purely on direction would have silently broken that pattern, which `examples/models/qwen3_jit` and others rely on.

Pass-by-value params now take the call-site-temporary path the pass already had for a slice / `IterArg` argument:

```python
n_inline0: pl.Scalar[pl.INDEX] = k
n_inline0 = n_inline0 + 1       # stays callee-local
m: pl.Scalar[pl.INDEX] = n_inline0
s: pl.Scalar[pl.INDEX] = k + m  # the ORIGINAL k
```

### `@pl.jit.inline` is affected too

The specializer's alpha-rename does **not** cover this. `_visit_simple_assign` renames a rebinding only at the scope depth where the name was first bound, so a rebind inside `pl.range` / `pl.spmd` reaches the pass as a genuine def-site of the param. That variant previously clobbered the caller's scalar all the way into the final orchestration IR — `CtrlFlowTransform` turned the clobber into a loop carry over the caller's `k`, and `s` read the loop result twice.

## Changes

- `src/ir/transforms/inline_functions_pass.cpp` — `CloneInlineBody` gains a `rebind_aliases_caller` test over `param_directions_` and the param's type; the existing temporary-binding branch handles the rest. Seed-map comment updated.
- `tests/ut/ir/transforms/test_inline_functions.py` — three regressions in `TestInlineFunctionsParamRebinding`: the scalar case, a tensor-param counterpart pinning that aliasing is **kept**, and an end-to-end `@pl.jit.inline` loop-carry case asserting the caller's param is never an assignment LHS.
- `docs/{en,zh}/dev/passes/01-inline_functions.md` — new "Rebinding a parameter" section with the table above, plus an edge-case row and an algorithm-step correction.

## Reviewer notes

A rebound **non-`pl.Out` tensor** param still retargets the caller's `Var`: `t = pl.tensor.add(t, t)` binds `t` to a *new* buffer, yet the caller's name follows it, so an inline helper reading and rebinding a plain `pl.Tensor` param changes what the caller sees. That is name aliasing applied to a non-in-place op, and it is a real defect — but closing it needs a policy decision (either stop stripping `pl.Out` on `@pl.jit.inline` shaped params, making direction authoritative at the cost of a user-visible break, or distinguish in-place ops from value-producing ops at the def-site). Deliberately out of scope here.

## Verification

- `cmake --build build --parallel "$PYPTO_BUILD_JOBS"` (20 jobs): passed.
- `python -m pytest tests/ut/ -n "$PYPTO_TEST_JOBS" -q` (8 workers): **13,919 passed**, 5 skipped, 1 xfailed.
- `python -m pytest tests/ut/ir/transforms/test_inline_functions.py tests/ut/jit -n 8 -q`: **851 passed**.
- `python tests/lint/clang_tidy.py --strict-version --jobs=20 --diff-base=HEAD`: clean.
- `pre-commit run --files <changed>`: all hooks pass, including cpplint, markdownlint and Pyright.

Each fix was confirmed against a built repro before and after: the `@pl.program` scalar case, the deep-scope `@pl.jit.inline` case lowered end-to-end through `lower(platform="a2a3")`, and the `pl.Out` aliasing counterpart. No device or simulator execution was performed; system suites are left to CI.


Hzfengsy:
@codex review

Hzfengsy:
@codex review

Hzfengsy:
@codex review

lyfne123 (approved):

